PR #8, opened 2023-10-12
[Workflow] Update Packages Deploy

stevensJourney:
This PR updates the package deploy workflow process. 

# Versioning 
Package versioning tools like Lerna https://github.com/lerna/lerna/issues/2342  and Changesets https://github.com/changesets/changesets/issues/818 have open issues when working with Git submodules. 

Submodules in this repo help with development and testing when using included testing Applications. Development packages are supported when making changes that include submodule packages, but production packages require the submodule packages to be versioned, tagged and published from their own repo before updating local packages here.

Versioning Instructions are added to the readme.

# Fixes

This PR contains a fix to allow watched queries to reload when `local_only` tables are updated.


packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -35,15 +35,17 @@ export interface WatchOnChangeEvent {
   changedTables: string[];
 }
 
+export interface PowerSyncDBListener extends StreamingSyncImplementationListener {}
+
+const POWERSYNC_TABLE_MATCH = /(^ps_data__|^ps_local__)/;
```
rkistner, line 40: This should be `/(^ps_data__|^ps_data_local__)/`

rkistner (approved):

packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
stevensJourney: 🙏  Good catch. Thanks for pointing that out.

PR #19, opened 2023-10-31
Update connectivity status

stevensJourney:
This PR adds more frequent updates of sync status connectivity.


packages/powersync-sdk-common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -68,6 +71,10 @@ export abstract class AbstractStreamingSyncImplementation extends BaseObserver<S
     return this.options.logger!;
   }
 
+  get isConnected() {
+    return this._isConnected;
+  }
```
Manrich121, line 76: Why use a `getter` instead of reporting the state directly?

Manrich121 (approved):

packages/powersync-sdk-common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
stevensJourney: This way it's read only

PR #20, opened 2023-10-31
[Feature] `@journeyapps/powersync-attachments` library

Manrich121:
PowerSync library to manage attachments in TypeScript and React Native apps


packages/powersync-attachments/README.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+2. Implement `attachmentIds`, an `AsyncIterator` method to return an array of `string` values of IDs that relate to attachments in your app. We recommend using `PowerSync`'s `watch` query to return the all IDs of attachments in your app. 
+In this example, we query all photos that have been captured as part of an inspection and map them to an array of `string` values.
+
+```typescript
+import { AbstractAttachmentQueue } from '@journeyapps/powersync-attachments';
+
+export class AttachmentQueue extends AbstractAttachmentQueue {
+  async *attachmentIds(): AsyncIterable<string[]> {
```
stevensJourney, line 67: If I recall previous feedback correctly, these examples should target JavaScript in order to be copy pasteable in Javascript apps

stevensJourney (dismissed):

packages/powersync-attachments/src/StorageAdapter.ts
```diff
@@ -0,0 +1,39 @@
...
+    options?: { mediaType?: string }
+  ): Promise<void>;
+
+  downloadFile(filePath: string): Promise<Blob>;
+
+  writeFile(
+    fileURI: string,
+    base64Data: string,
+    options?: { encoding?: EncodingType }
+  ): Promise<void>;
```
rkistner, line 19: For a future version, it may be better to just take an ArrayBuffer (or Uint8Array) as parameter, and do the base64 encoding in the adapter implementation.

Also, not directly relevant here, but the [demo implementation](https://github.com/journeyapps/powersync-supabase-react-native-todolist-demo/pull/12/files#diff-21e1ce17f668cd0137419c202dcca197864d18b90e44e366027a713b4d368f8eR50) defaults to UTF8 decoding, despite the parameter being called `base64Data`.

packages/powersync-attachments/src/StorageAdapter.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  readFile(
+    fileURI: string,
+    options?: { encoding?: EncodingType; mediaType?: string }
+  ): Promise<ArrayBuffer>;
```
rkistner, line 24: Is encoding relevant when an ArrayBuffer is returned either way?

packages/powersync-attachments/src/StorageAdapter.ts
```diff
@@ -0,0 +1,39 @@
+export enum EncodingType {
+  UTF8 = "utf8",
+  Base64 = "base64",
+}
+
+export interface StorageAdapter {
```
rkistner, line 6: It would help to have documentation for the methods and parameters here - especially the difference between file URI and path.

rkistner (approved):
LGTM - just noted some future improvements for the adapter APIs.

PR #24, opened 2023-11-02
[Fix] Common SDK fixes for Web and React Native

stevensJourney:
This PR contains some fixes identified when implementing the Web SDK:

- Fixed: `get`, `getAll` and `getOptional` helper methods should execute inside a readLock for better concurrency
- Removed `user-id` header from backend connector and remote headers
   The user-id field is not used and is not accepted for CORS
- Added `waitForReady` method on PowerSyncDatabase
- Fixed: Added correct typings for React hooks. Previously hooks would return `any`.
- Bumps `@journeyapps/react-native-quick-sqlite` which has bug fixes from https://github.com/powersync-ja/react-native-quick-sqlite/pull/9
- Updates SDK readme docs URL and new Github org repo location

TODO:
- [x] Bump package for react-native-quick-sqlite (currently cannot publish due to Github org permissions error)


packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -71,13 +71,28 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
...
+    // This ensures the `initialized` promise is pending even before `init` is called
+    this.initialized = new Promise((resolve, reject) => {
+      this.init = async () => {
+        try {
+          resolve(await this.initFn());
+        } catch (ex) {
+          reject(ex);
+        }
+      };
+    });
```
Manrich121, line 95: It is not very clear to me why this type of Promise juggling is necessary. If done right, an initialization lifecycle should be very deterministic and consistent. Lets talk offline

packages/powersync-sdk-react-native/src/db/adapters/react-native-quick-sqlite/RNQSDBAdapter.ts
```diff
@@ -55,6 +59,16 @@ export class RNQSDBAdapter extends BaseObserver<DBAdapterListener> implements DB
     return this.baseDB.execute(query, params);
   }
 
+  /**
+   * This provides a top-level read only execute method which is executed inside a read-lock.
+   * This is necessary since the high level `execute` method uses a write-lock under
+   * the hood. Helper methods such as `get`, `getAll` and `getOptional` are read only,
+   * and should use this method.
```
Manrich121, line 66: What is a `top-level` read-only execute method? 
According to the comment, `get`, `getAll` and `getOptional` should use this, but it does not seem to be the case

stevensJourney: The DB adapter and the powersync client expose SQL execution methods such as `execute`, `getAll`, `get` and `getOptional` on the class (top-level) e.g. `db.execute()`, `db.get()` and they also expose these methods in lock and transaction contexts e.g. `tx.get()`. 

The top-level `get` functions were previously using the top-level `execute` function which would actually schedule a writeLock and then use the writeLock's context `execute` method. These methods should actually be using a `readLock` instead since those can execute concurrently with writeLocks. This method provides a read only `execute` method on the DB adapter class as shorthand for starting a readLock and using it's `execute` method. 

This method is used to create the top-level `get` helpers here https://github.com/journeyapps/powersync-react-native-sdk/blob/fc2eb5d9bf74e71ee3cd7fe896ecba5ca4f27081/packages/powersync-sdk-react-native/src/db/adapters/react-native-quick-sqlite/RNQSDBAdapter.ts#L30

packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
stevensJourney: Updated this to use a listener pattern. The necessity of the promise is to ensure that user call's to methods such as `execute` do not occur before the client has been initialised. This is mainly due to that the DB schema and tables are not necessarily created until `init` has been called.

Manrich121 (approved):
100%

PR #33, opened 2023-11-28
[Improvement] Better locks and Sync

stevensJourney:
# Description

This PR introduces some improvements for the sync locking and status flows. This PR also adds changesets to release out of Beta to v1.0.0

## Sync Status
The SyncStatus now includes the state of if the connector is uploading or downloading data.

## Sync Flow
Crud uploads are now debounced.
Crud uploads now are also triggered on `execute` methods.
Database name is now added to the `DBAdapter` interface for better identification in locks (for Web SDK)
Failed crud uploads now correctly throw errors, to be caught upstream, and delayed for retry.

### DIFF
--- a/packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
+++ b/packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
@@ -339,7 +339,9 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
    */
   async execute(sql: string, parameters?: any[]) {
     await this.waitForReady();
-    return this.database.execute(sql, parameters);
+    const result = await this.database.execute(sql, parameters);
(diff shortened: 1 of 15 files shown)


packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -339,7 +339,9 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
    */
   async execute(sql: string, parameters?: any[]) {
     await this.waitForReady();
-    return this.database.execute(sql, parameters);
+    const result = await this.database.execute(sql, parameters);
+    _.defer(() => this.syncStreamImplementation?.triggerCrudUpload());
```
rkistner, line 343: For future reference, it would be good if we could do these using update and/or commit hooks based on changes to the `ps_crud` table instead.

rkistner (dismissed):

dylanvorster (approved):

PR #37, opened 2023-12-01
[Feat] add docs to github pages

DominicGBauer:
## Description
Create a pipeline that automates doc creation using MD files which can be displayed using Docusaurus. Using Docusaurus allows us access to a vast amount of functionality and an ever-improving ecosystem and is maintained by Meta.

## Work Done 
* Used typedoc to autogenerate docs usable by Docusaurus
* All Docusaurus output can be found in a new docs subdirectory from the root (not there was an issue with how yarn workspaces managed packages and it created conflicts with the autogeneration of the docs hence I left docs out of the yarn workspaces)

## How to test
* Run `yarn build:packages`
* Run `cd docs && yarn build && yarn start`
* Run `cd docs && yarn build`
* In docs folder open index.html
* Review 

## Screenshots
image


docs/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Build
+
+```
+$ yarn build
+```
+
+This command generates static content into the `build` directory and can be served using any static contents hosting service.
+
+### Deployment
```
stevensJourney, line 27: Seems like this should be removed if we are using a Github action to deploy? Or is it intended to be deployed from a local machine?

docs/docusaurus.config.ts
```diff
@@ -0,0 +1,181 @@
...
+                  <img src="/${PROJECT_NAME}/img/web.svg" loading="lazy" alt="" height="30">
+                </a>
+              </div>
+            `}
+          ],
+        },
+      ],
+      copyright: `Copyright © ${new Date().getFullYear()} Journey Mobile, Inc.`,
+    },
+    // algolia: {
```
stevensJourney, line 167: Why is this commented out?

stevensJourney (commented):
Looks good so far, left some comments and questions

docs/README.md
DominicGBauer: You can deploy from your local machine as well, which can be helpful when not necessarily always wanting to push up to the branch and fire an action

docs/docs/index.md
```diff
@@ -0,0 +1,94 @@
+---
+id: "index"
+title: "PowerSync React Native SDK Docs"
+sidebar_label: "Readme"
+sidebar_position: 0
+custom_edit_url: null
+---
+
+# PowerSync React Native SDK
```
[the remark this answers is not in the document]
DominicGBauer: This is the homepage when going to https://powersync-ja.github.io/powersync-react-native-sdk/. If we don't have the file nothing is displayed and there is a page not found instead. Ideally, this page will be changed in future to be more like a homepage.

docs/docusaurus.config.ts
DominicGBauer: Sorry I should have explained in a comment above. Algolia search isn't working at the moment and I am trying to get some support but because we're on the free tier it's taking some time. Will update with a comment explaining the issue

docs/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "docs",
+  "private": true,
+  "version": "0.0.0",
+  "scripts": {
+    "docusaurus": "docusaurus",
+    "start": "docusaurus start",
+    "build": "docusaurus build",
+    "swizzle": "docusaurus swizzle",
```
Manrich121, line 9: Are all these scripts needed for local dev?

docs/sidebars.ts
```diff
@@ -0,0 +1,63 @@
+import type { SidebarsConfig } from '@docusaurus/plugin-content-docs';
+
+/**
+ * Creating a sidebar enables you to:
```
Manrich121, line 4: We probably don't need the template comments here. Remove it or make it specific to our use

docs/sidebars.ts
```diff
@@ -0,0 +1,63 @@
...
+    label: 'Libraries',
+    collapsible: false,
+    items: [
+      {
+        type: 'category',
+        label: 'React Native SDK',
+        items: [
+          {
+            type: 'autogenerated',
+            dirName: 'react-native-sdk',
```
Manrich121, line 25: Am I correct that this `dirName` needs to match with the `plugins` registered in the main config? If so, it would be good to have a way to set it once and automatically get it from there. Something like `Plugin.generateSidebarItem()`

package.json
```diff
@@ -7,7 +7,9 @@
     "clean:packages": "lerna run clean",
     "build:packages": "lerna run build",
     "watch:packages": "yarn workspace  @journeyapps/powersync-sdk-react-native watch",
-    "release": "yarn build:packages && yarn changeset publish"
+    "release": "yarn build:packages && yarn changeset publish",
+    "build:docs": "cd docs && yarn && yarn build",
+    "start:docs": "cd docs && yarn && yarn build && yarn start"
```
Manrich121, line 12: Is the `cd docs` really needed, or could we use `lerna` to run this from the root?

docs/docusaurus.config.ts
```diff
@@ -0,0 +1,184 @@
...
+      ],
+    },
+    footer: {
+      style: 'light',
+      links: [
+        {
+          title: 'Community',
+          items: [
+            {
+              html: `
```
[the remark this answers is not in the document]
DominicGBauer: After exploring this option, I don't think the effort and needing to install extra packaging so the bundler knows how to handle HTML files is worth it in this case.

docs/docs/index.md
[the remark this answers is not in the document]
DominicGBauer: I think this might be something for the future since we probably don't really want the README to be the homepage. I've fixed the links though so they now redirect to the correct readme

rkistner (approved):

PR #47, opened 2024-02-07
chore: move to js monorepo

DominicGBauer:
## Description
Move toward a JS monorepo with Pnpm

## Work Done
* Moved to pnpm
* Included react native project
* Changed github actions to use pnpm
* Packages now use workspace deps for local packages that they are dependent on (not sure if this is correct or you want to keep them set?)


package.json
```diff
@@ -1,32 +1,31 @@
 {
-  "private": true,
```
stevensJourney, line 2: This should probably stay private

package.json
```diff
@@ -1,32 +1,31 @@
...
-    "nohoist": [
-      "**/@rneui/**",
-      "**/react-native",
-      "**/react-native/**",
-      "**/expo",
-      "**/expo/**"
-    ]
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
stevensJourney, line 20: Whats the reason for changing this to ISC?

stevensJourney (commented):
Left some comments. The `workspace` package version dependencies should be fine to use. 

Some notes from testing the demo app:
It seems like some of the Android config which enabled streamed queries is no longer applied. When testing on a device the connection did not resolve.

iOS seems to work fine 😄

demos/supabase-todo-list-react-native/.envrc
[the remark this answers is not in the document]
DominicGBauer: Does having it in the root apply to all subdirectories?

.github/workflows/deploy-docs.yaml
```diff
@@ -27,23 +27,29 @@ jobs:
...
+            echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
+      - uses: actions/cache@v3
+        name: Setup pnpm cache
+        with:
+          path: ${{ env.STORE_PATH }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-store-
+      - name: Install dependencies
+        run: pnpm install
```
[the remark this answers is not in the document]
DominicGBauer: It's enabled by default in CI env https://pnpm.io/cli/install#--frozen-lockfile

demos/supabase-todo-list-react-native/.envrc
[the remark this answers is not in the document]
stevensJourney: It should: https://direnv.net/ (if using Direnv)

> Before each prompt, direnv checks for the existence of a .envrc file (and [optionally](https://direnv.net/man/direnv.toml.1.html#codeloaddotenvcode) a .env file) in the current and parent directories.

But even if it didn't: it would be nice to have it in the monorepo root as well.

package.json
DominicGBauer: Mistake sorry

demos/supabase-todo-list-react-native/android/app/src/main/java/com/powersync/example/MainApplication.kt
```diff
@@ -0,0 +1,65 @@
...
+    super.onCreate()
+    SoLoader.init(this, false)
+    if (!BuildConfig.REACT_NATIVE_UNSTABLE_USE_RUNTIME_SCHEDULER_ALWAYS) {
+      ReactFeatureFlags.unstable_useRuntimeSchedulerAlways = false
+    }
+    if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
+      // If you opted-in for the New Architecture, we load the native entry point for this app.
+      load()
+    }
+    // if (BuildConfig.DEBUG) {
```
stevensJourney, line 55: We should probably make a note to update the Docs about how to perform this step in Kotlin based native apps

DominicGBauer: Yeah agreed do you know where the best place to note this would be?

stevensJourney (approved):
Excellent work!

demos/supabase-todo-list-react-native/android/app/src/main/java/com/powersync/example/MainApplication.kt
benitav: if you don't have anything on the engineering side, we have this catch-all doc: https://paper.dropbox.com/doc/2023-09-PowerSync-Docs-TODOs--CJApoZRQ9bWecJtak7hDuuiZAg-Z4p6MStzftuLhHNrYx3hn (but might get lost there)

DominicGBauer: @benitav we actually just updated the docs in gitbook will add you as reviewer for final sign off :)

PR #50, opened 2024-02-08
chore: add web-sdk

DominicGBauer:
## Description
Move Web-SDK into monorepo

## Work Done
* Moved over web-sdk package and demos
* Updated package.json to use local repos where I could


stevensJourney (approved):
LGTM

PR #66, opened 2024-02-12
Improve Readmes and code commenting

benitav:
* Leaner Readme's with details moved to our docs SDK references. 
* More consistent Readme's across RN and Web (Flutter is TODO separately).
* Added JSDoc comments for common methods and properties

- [x] This PR is dependent on merging this Gitbook change request: https://app.gitbook.com/o/YjXqDRfjWRhJMPvQn7Cm/s/V9geAO9ITPi8WOYK5o0r/~/diff/~/changes/391/client-sdk-references/react-native-and-expo


packages/powersync-sdk-common/src/client/sync/bucket/OpType.ts
```diff
@@ -7,6 +7,9 @@ export enum OpTypeEnum {
 
 export type OpTypeJSON = string;
 
+/**
```
DominicGBauer, line 10: if this is internal should we not ignore it?

packages/powersync-sdk-common/src/db/DBAdapter.ts
```diff
@@ -42,12 +42,14 @@ export interface DBGetUtils {
 }
 
 export interface LockContext extends DBGetUtils {
-  /** Execute a statement and optionally return results. */
+  /** Execute a single write statement. */
```
DominicGBauer, line 45: are you sure this is limited to `writes` ?

packages/powersync-sdk-common/src/utils/BaseObserver.ts
```diff
@@ -15,6 +15,9 @@ export class BaseObserver<T extends BaseListener = BaseListener> implements Base
     this.listeners = {};
   }
 
+  /**
+   * Register a listener for updates on the PowerSync client.
```
DominicGBauer, line 19: would it not be updates to* as oposed to on?

packages/powersync-sdk-common/src/db/DBAdapter.ts
benitav: mm, according to https://pub.dev/documentation/powersync/latest/powersync/PowerSyncDatabase/execute.html and https://docs.powersync.com/client-sdk-references/react-native-and-expo#using-powersync I thought so yes, but perhaps I mapped it incorrectly here?

packages/powersync-sdk-common/src/client/sync/bucket/OpType.ts
benitav: Possibly; I'm keeping ignores for a future PR.

packages/powersync-sdk-common/src/db/DBAdapter.ts
DominicGBauer: Yeah I think you are correct. Thanks

PR #71, opened 2024-02-19
[Fix] Shared sync

stevensJourney:
# Description
It was recently discovered in the YJS NextJS demo app that watched uploads would take long to trigger if multiple Powersync instances were created for the same DB. This was due to the sync status not being correctly shared between the (multiple) instances. The second instance would not trigger uploads since it did not know a connection to the PowerSync instance was established by the first instance. 

This same issue would manifest itself in regular apps (with a single instance per tab). Sync statuses between multiple tabs would not be shared. Subsequently opened tabs would remain in `connected: false` mode. 

The root cause of this issue was that the `SharedSyncImplementation.worker.ts` shared worker failed to run and share syncing information between tabs and instances. The worker thew an exception when trying to access the `window` object in the global worker scope. The `window` object is not directly used in this worker codebase, but it is tested against in the `object-hash` library. The issue seems to be cause from bumping the NextJS version to `14.1.0`. [This](https://github.com/vercel/next.js/issues/60644) issue mentioned that versions after `14.0.0` can incorrectly compile checks on `window` objects. This issue seems to be resolved here https://github.com/vercel/next.js/commit/cfedc529c713e66fefd51df9c8e226b2ddfd95ad. For better compatibility the `object-hash` library is removed since it's not actually needed. This fixes the shared worker from trying to access `window`. 

# Testing

Functionality was confirmed for the YJS demo app and React Native Todolist apps.

https://github.com/powersync-ja/powersync-js/assets/51082125/534d2237-fb43-42a6-a9ad-764a733a17e5

### DIFF
--- a/packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
+++ b/packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
@@ -299,7 +299,7 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
     });
   }
 
-  /** 
+  /**
    * Close the database, releasing resources.
    *
    * Also disconnects any active connection.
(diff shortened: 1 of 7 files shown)


packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -352,26 +352,28 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
    * and a single transaction may be split over multiple batches.
    */
   async getCrudBatch(limit: number): Promise<CrudBatch | null> {
-    const result = await this.database.execute(
-      `SELECT id, tx_id, data FROM ${PSInternalTable.CRUD} ORDER BY id ASC LIMIT ?`,
-      [limit + 1]
-    );
+    return this.readLock(async (db) => {
+      const result = await db.getAll<CrudEntryJSON>(
```
rkistner, line 356: You shouldn't need a ready lock when there aren't any other database operations in the same lock.

rkistner (approved):

packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
stevensJourney: Correct 🙏 . The readLock call has been removed.

rkistner (approved):

PR #72, opened 2024-02-20
[Improvement] Shared Sync Worker

stevensJourney:
# Description

Previously a shared worker would only co-ordinate syncing between multiple clients in multiple tabs. One tab would be responsible for connecting to the PowerSync backend instance and applying updates.

This fully moves the sync stream connection and sync bucket manager operations into the shared sync worker. Multiple tabs will receive status updates as usual, but will also directly control the shared sync manager's connection state with calls to `connect` and `disconnect`. If any tab performs a `disconnect`, then the sync client will be disconnected for all tabs until a `connect` is called from a tab. 

The upload connector cannot be easily transferred to the shared sync worker. The sync worker will fetch credentials and perform CRUD uploads on the last connected client/tab.

The shared sync worker cannot instantiate it's own connection to the shared DB worker. The `DBAdapter` required for the SQLite bucket storage adapter is obtained by the host `PowerSyncDatabase` client passing it's DB worker `MessagePort` to the shared sync worker. The Shared sync worker can then use the passed port to instantiate DB connections from the shared DB worker. 

## TODO

- [x] Potentially share logs between shared worker and individual tabs for easier debugging
- [x] Handle closed tabs better. Remove them from shared sync worker when closed.
    - [x] Handle calls to `close` better in multiple tab mode.


packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -307,12 +292,20 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
    *
    * Once close is called, this connection cannot be used again - a new one
    * must be constructed.
+   *
+   * @param disconnect Disconnect the sync stream client. This is usually true,
+   * but can be false for Web when using multiple tabs and a shared sync provider.
    */
-  async close() {
+  async close(disconnect = true) {
```
rkistner, line 299: Rather use something like `options?: {disconnect?: boolean}` - it makes client code more clear when it's `close({disconnect: false})` rather than `close(false)`.

rkistner:
Issue found when testing Yjs demo:
1. Open tab A and incognito tab I to monitor changes. Edits from A propagate correctly to I.
2. Open tab B. Edits from both A and B propagate correctly to I.
3. Close tab B. Now edits from tab A stop propagating to tab I. (If you close tab A instead, it keeps on working)

packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
stevensJourney: 👍 This is better.

stevensJourney:
> Issue found when testing Yjs demo:
> 
> 1. Open tab A and incognito tab I to monitor changes. Edits from A propagate correctly to I.
> 2. Open tab B. Edits from both A and B propagate correctly to I.
> 3. Close tab B. Now edits from tab A stop propagating to tab I. (If you close tab A instead, it keeps on working)

I was able to reproduce and implement a fix for this issue. I believe this was due to closing the tab while the shared worker was awaiting a response from the proxied `uploadData` function. The YJS demo doesn't include an `expiredAt` field in the returned credentials, which causes credentials to be fetched for every request. This Supabase endpoint is very slow, which causes the `uploadData` function to be slow. 

I cleaned up the `unload` event handler for the window and added abort controllers for aborting requests to clients if their port is removed.

packages/powersync-sdk-web/src/db/PowerSyncDatabase.ts
```diff
@@ -24,19 +26,59 @@ export interface WebPowerSyncFlags {
    * Open in SSR placeholder mode. DB operations and Sync operations will be a No-op
    */
   ssrMode?: boolean;
+  /**
+   * Externally unload open PowerSync database instances when the window closes.
+   * Setting this to `true` requires calling `close` on all open PowerSyncDatabase
+   * instances before the window unloads
+   */
+  externallyUnload?: boolean;
```
rkistner, line 34: What does this mean in practice? (What's "externally" here?)
What happens when you set this to true and don't call close on the open instances?

stevensJourney: Currently each PowerSync database instance will register a listener for the `unload` event and dispose itself by default. 

This was added with the thought process that a user might have their own custom `unload` event handler logic which they might want to still use the Powersync client for. It might be a bit of an over-engineered case though. I would be happy to remove if necessary.

It is expected that all the instances should be closed. At this point failing to do so will most likely result in a "locked" sync worker. We might be able to improve this in future by adding some sort of `ping` `pong` method to MessagePorts.

rkistner (approved):

PR #76, opened 2024-02-22
Web tests with Vitest

stevensJourney:
This PR adds automated tests for the Web SDK using Vitest and WebdriverIO. 

Tests are performed in a browser environment due to the requirements for `SharedWorker` `Navigator` locks and WASM loading. 

This PR uses an interesting TypeScript and Vite config to allow resolving web worker URIs to be compatible with Webpack and Vite. In the web SDK package: Vite will only resolve web worker URIs correctly if the file path ends with `.ts` while the URI should use `.js` in published packages (not using the source files directly).

The method used here is to use the `@journeyapps/powersync-sdk-web` package path alias as an import source for the `**/*.test.ts` files. This alias is then resolved in the Vite bundling step to the Javascript `dist` version of the `src`.


.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
+# Ensures packages build correctly
+name: Build Packages
```
stevensJourney, line 2:
```suggestion
# Ensures packages test correctly
name: Test Packages
```

packages/powersync-sdk-common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -443,7 +444,7 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
    */
   async execute(sql: string, parameters?: any[]) {
     await this.waitForReady();
-    return this.database.execute(sql, parameters);
+    return this.writeTransaction((tx) => tx.execute(sql, parameters));
```
rkistner, line 447: Some statements such as some PRAGMA statements do not work in transactions - execute shouldn't use transactions by default

rkistner (approved):

PR #83, opened 2024-02-28
Switch from NextJS -> Vite for demos

rkistner:
NextJS's server-side rendering mostly just conflicts with our local-first approach, with multiple workarounds required to make it functional. This attempts to simplify the supabase-todolist demo by not using NextJS.

The app is mostly unchanged. The main changes are:
1. Use Vite as a bundler.
2. Use react-router instead of NextJS navigation.
3. Shuffle the providers around a bit to better work with react-router.
4. Move lists out as a separate TodoListsWidget.

In the future, it will also allow:
1. Using react-router's loaders to load data before navigating to a new page - see #84 

This similarly migrates the Yjs demo, adds a new "example-nextjs" to still have the basic boilerplate for NextJS available.

This also fixes some null-check issues, giving better feedback when credentials are not provided.

TODO:
 * [x] Check if more config or dependencies can be cleaned up.
 * [x] Rename demo to `react-supabase-todolist`.
 * [x] Do the same for the Yjs demo.
 * [x] Add a new basic NextJS demo, to still have a working example (similar to the current Vite demo).
 * [x] Fully offline-capable PWA.


demos/react-supabase-todolist/README.md
```diff
@@ -0,0 +1,59 @@
...
+```bash
+cp .env.local.template .env.local
+```
+
+And then edit `.env.local` to insert your credentials for Supabase.
+
+Run the development server:
+
+```bash
+pnpm watch
```
stevensJourney, line 32: This script doesn't exist in the `package.json`

demos/react-supabase-todolist/README.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+Open [http://localhost:3000](http://localhost:3000) with your browser to see the result.
+
+## Progressive Web App (PWA)
```
stevensJourney, line 37: Is this demo still PWA compatible? I couldn't find any service worker config, maybe it's automatically generated?

demos/react-supabase-todolist/src/app/page.tsx
```diff
@@ -1,27 +1,26 @@
...
- * This page shows a loading spinner until the _layout.tsx
- * file detects a session and redirects either to the app or
- * auth flow.
+ * This page shows a loading spinner we detect a session
+ * and redirect either to the app or auth flow.
  */
 export default function EntryPage() {
-  const router = useRouter();
   const connector = useSupabase();
+  let navigate = useNavigate();
```
stevensJourney, line 19:
```suggestion
  const navigate = useNavigate();
```

demos/react-supabase-todolist/README.md
rkistner: Updated the readme.

demos/react-supabase-todolist/README.md
rkistner: I added back PWA support using `vite-plugin-pwa`. The Next.js version had basic caching support (you had to visit each page before going offline), but this now works fully offline.

demos/example-nextjs/.env.local.template
```diff
@@ -0,0 +1 @@
+NEXT_PUBLIC_POWERSYNC_URL=
```
benitav, line 1: We probably need to mention setting this field (and renaming this file) in the Readme?

rkistner: Actually getting sync to work would also require setting up a connector. The other two examples (example-vite and example-webpack) also don't have anything for that.

Do you think it's worth expanding all those examples, perhaps with instructions using temporary development tokens?

benitav:
IIRC we published those to showcase bundling with these frameworks; I think
showcasing the sync setup is also going to be helpful, but could also be a
future addition. If we implement this later I'd just say we add a note here
explaining how this field would be used, or remove it.

demos/react-supabase-todolist/README.md
stevensJourney: Awesome!

stevensJourney (dismissed):

rkistner:
> IIRC we published those to showcase bundling with these frameworks; I think showcasing the sync setup is also going to be helpful, but could also be a future addition. If we implement this later I'd just say we add a note here explaining how this field would be used, or remove it.

@benitav I added minimal notes on configuring the example with a development token.

demos/example-nextjs/.env.local.template
guillempuche: As newcomer, fake auth connection is good for learning and testing purposes.

stevensJourney (approved):

stevensJourney (approved):

PR #90, opened 2024-03-05
 Remove lodash (mostly)

rkistner:
Remove lodash imports to decrease bundle size. Changes the bundle size of example-vite from 127kB -> 56kB. (This excludes the web workers and SQLite WASM build, but those can be loaded asynchronously).

The only remaining import in the core SDKs is `lodash/throttle`.

The updated code is not quite equivalent to the previous lodash version, so it needs some testing.

### DIFF
--- a/packages/powersync-sdk-web/src/worker/sync/SharedSyncImplementation.ts
+++ b/packages/powersync-sdk-web/src/worker/sync/SharedSyncImplementation.ts
@@ -1,4 +1,3 @@
-import _ from 'lodash';
 import { BaseListener, BaseObserver, SyncStatusOptions } from '@journeyapps/powersync-sdk-common';
 
 export enum SharedSyncMessageType {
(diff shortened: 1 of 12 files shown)


packages/powersync-sdk-web/src/worker/sync/SharedSyncImplementation.ts
```diff
@@ -38,7 +37,10 @@ export class SharedSyncImplementation extends BaseObserver<SharedSyncImplementat
   }
 
   updateState(status: SharedSyncStatus) {
-    this.status = _.merge(this.status, status);
+    this.status = {
```
stevensJourney, line 40: This might break the state when updating between the nested `dataFlowStatus`  fields. as here https://github.com/journeyapps/powersync-react-native-sdk/blob/f81714007c29e3c0de32459f5fe94e66355727dc/packages/powersync-sdk-common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts#L166. 

`_.merge` was doing a recursive merge before, while this would erase the state of `dataFlowStatus.uploading` if `dataFlowStatus.downloading` was updated.

rkistner: Good catch - I updated to merge `dataFlow` as well.

stevensJourney (approved):

stevensJourney:
Should we add a changeset for these changes? I think it might be worth while mentioning the decreased bundle size 🤔

stevensJourney (approved):
LGTM. Tested Both todo list apps - they seem to work as before.

PR #98, opened 2024-03-13
[DEMOS] chore: updated react-native-supabase-todolist demo to use expo environment variables

Chriztiaan:
# Description
Updated `react-native-supabase-todolist` to support EAS.

Instead of configuring the variables in the `AppConfig.ts` file, configuration is done via a `.env` or `.env.local` file.
An example file is provided named `.env.template`, I added the example file to the `react-native-supabase-group-chat` demo too and removed the default `.env` that was there.


demos/react-native-supabase-todolist/library/supabase/AppConfig.ts
```diff
@@ -0,0 +1,6 @@
+export const AppConfig = {
```
stevensJourney, line 1: It should be possible to define the TypeScript interface for `process.env`. This will remove the requirement for `as string` and also give some type safety for which variables are expected to be present.

PR #112, opened 2024-04-04
chore: prettier formatting

Chriztiaan:
Diff was generated from `format` command.


stevensJourney (approved):
We should probably add a test in future to ensure formatting for all PRs.

PR #120, opened 2024-04-10
[Fix] - Possible race condition/handling uncaught exception in watch()

Chriztiaan:
There was a small window between getting the first result and setting up the onchange handler. The window is smaller now, considered setting up the onchange handler before retrieving first result. Will look into it as part of investigating backpressure. Also took a fix found from the `vue` branch and added it here to handle an uncaught exception.


rkistner (approved):

Chriztiaan:
Two unit test were failing because of this change. It seems that previously a `watch` with an immediate `insert` would end up with two onResult yields. That case now results in a single onResult yield. For these specific tests which check that only watched table table changes trigger yields - I put a sleep between setting up the watch and the insert.

@stevensJourney confirm that this is expected please.

PR #123, opened 2024-04-16
Update lodash reference in vite optimizeDeps demo configs

aguynamedloren:
Hey folks!

I ran into a bit of trouble during setup with a new codebase, using powersync-sdk-web, bundled with vite.  I modeled my vite config after the [demo config here](https://github.com/powersync-ja/powersync-js/tree/main/demos/example-vite). Upon page load, the following error was displayed in the browser console:

```Uncaught SyntaxError: The requested module '.../lodash/throttle.js' does not provide an export named 'default' (at AbstractStreamingSyncImplementation.js)```

I was able to resolve the issue by changing the vite config `lodash` depdency line to `lodash/throttle`.  I suspect this may've been a regression with the recent [lodash optimization](https://github.com/powersync-ja/powersync-js/pull/90).  I'm new to vite, not sure if this is the best solution, but it worked for me, and I hope it'll help others hitting the same wall I ran into.

Appreciate the great work you're doing here!

### DIFF
--- a/demos/example-vite/vite.config.ts
+++ b/demos/example-vite/vite.config.ts
@@ -18,7 +18,7 @@ export default defineConfig({
     // Don't optimize these packages as they contain web workers and WASM files.
     // https://github.com/vitejs/vite/issues/11672#issuecomment-1415820673
     exclude: ['@journeyapps/wa-sqlite', '@journeyapps/powersync-sdk-web'],
-    include: ['object-hash', 'uuid', 'event-iterator', 'js-logger', 'lodash', 'can-ndjson-stream']
+    include: ['object-hash', 'uuid', 'event-iterator', 'js-logger', 'lodash/throttle', 'can-ndjson-stream']
   },
   plugins: [wasm(), topLevelAwait()],
   worker: {
--- a/demos/react-supabase-todolist/vite.config.mts
+++ b/demos/react-supabase-todolist/vite.config.mts
@@ -25,7 +25,7 @@ export default defineConfig({
     // Don't optimize these packages as they contain web workers and WASM files.
     // https://github.com/vitejs/vite/issues/11672#issuecomment-1415820673
     exclude: ['@journeyapps/wa-sqlite', '@journeyapps/powersync-sdk-web'],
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by Chriztiaan** [comment]
Thanks for this. There has been another reported issue with a project setup where someone had to `includes` the [nested dependency](url). When I tested this on my side in a standalone repo I was only able to get this working with the nested definition.

Could I ask that you could test the following in your project? Just want to make sure it doesn't break anything: 
```
  include: [
      "event-iterator",
      "uuid",
      "js-logger",
      "lodash/throttle",
      "can-ndjson-stream",
      "@journeyapps/powersync-sdk-web > event-iterator",
      "@journeyapps/powersync-sdk-web > uuid",
      "@journeyapps/powersync-sdk-web > js-logger",
      "@journeyapps/powersync-sdk-web > lodash/throttle",
      "@journeyapps/powersync-sdk-web > can-ndjson-stream",
    ],
    ```

aguynamedloren:
@Chriztiaan thanks for the quick reply!  Yeah, that config worked for me -- didn't break anything.

I wasn't familiar with the nested dependency syntax until you shared it; my intuition is that it's probably the better approach.

Out of curiosity, I tried with nested dependency declarations only.  It almost worked, just had to change the `uuid` path:

```
    include: [
      '@journeyapps/powersync-sdk-common > uuid',
      '@journeyapps/powersync-sdk-web > event-iterator',
      '@journeyapps/powersync-sdk-web > js-logger',
      '@journeyapps/powersync-sdk-web > lodash/throttle',
      '@journeyapps/powersync-sdk-web > can-ndjson-stream'
    ]
```

Chriztiaan:
Thanks for checking! I'll fiddle a bit and test it on my side.

Chriztiaan:
I've updated the diff to use the nested dependency syntax. Thanks for the help.

stevensJourney (dismissed):

Chriztiaan:
Updated this after our package renaming `@journeyapps/powersync-xxx` -> `@powersync/xxx`.
Verified that this works without needing a `pnpm overrides` entry for `uuid` after bumping all the underlying UUID versions in https://github.com/powersync-ja/powersync-js/pull/130

stevensJourney (approved):
Thanks!

PR #124, opened 2024-04-16
chore: add django repo to monorepo

DominicGBauer:
## Description
Migrate django react native example and update it to expo 50

## Work Done
* Moved django react native example
* Updated to Expo 50

## How to test
Follow readme on how to get app up and running

## Screenshot
image


demos/django-react-native-todolist/README.md
```diff
@@ -0,0 +1,61 @@
...
+This package is currently in an alpha release. Functionality could change dramatically in future releases. Certain functions may be partially implemented or buggy.
+
+## Overview
+
+Demo app demonstrating use of the [PowerSync SDK for React Native](https://www.npmjs.com/package/@journeyapps/powersync-sdk-react-native) together with a Django backend.
+The sample backend that you can run alongside this demo can be found here: [PowerSync Django Backend: Todo List Demo](https://github.com/powersync-ja/powersync-django-backend-todolist-demo)
+
+## Running the App
+
+Install the React Native SDK, then:
```
benitav, line 14: Should this say clone this repo and/or cd into this folder or similar instead?

DominicGBauer: We don't do that in any of the other demos and it does seem implied since the readme is in the demo folder. But I'm happy to add it to the others as well if you think it will be useful?

benitav: We should remove "Install the React Native SDK, then" at least, since that doesn't make sense. A better way to put this would be to indicate that `pnpm install` installs all necessary dependencies, so say something like "Install the React Native SDK and dependencies:" or remove that line altogether.

demos/django-react-native-todolist/library/django/AppConfig.template.ts
```diff
@@ -0,0 +1,4 @@
+export const AppConfig = {
+    djangoUrl: '', // This is base url to the Django project
+    powersyncUrl: '' // This is the PowerSync url provided in the PowerSync dashboard
```
benitav, line 3:
```suggestion
    powersyncUrl: '' // This is the PowerSync instance url provided in the PowerSync dashboard
```

demos/django-react-native-todolist/README.md
benitav, line 14:
```suggestion
Install the dependencies, including the React Native SDK:
```

benitav (approved):

Issue #93, opened 2024-03-09
Vite example shows this error `Could not POST streaming to /sync/stream` in `WebRemote.ts` and `404 Not Found http://localhost:3020/sync/stream`


guillempuche:
When running `example-vite` `pnpm dev` with `vite.config.ts`:
```ts
server: {
   port: 3020,
},
```

The web seems working. But my browser shows this error:

<img width="469" alt="imatge" src="https://github.com/powersync-ja/powersync-js/assets/3519924/f7b325bf-f500-4826-a649-5fe622a793f9">

```sh
Could not POST streaming to /sync/stream - 404 - Not Found: [WebRemote.ts:58:14](http://localhost:3020/@fs/Users/guillem/programacio/codi/cites/databases/powersync-js/packages/powersync-sdk-web/src/db/sync/WebRemote.ts)
```

And also `XHR POST http://localhost:3020/sync/stream [HTTP/1.1 404 Not Found 1ms]`

Do you know why?

guillempuche:
I write the question here to be searchable in Google for others.

stevensJourney:
Hi @guillempuche this is currently expected behaviour. 

The log above the error mentions that there aren't any actual backend credentials configured. This will cause any network requests to the PowerSync instance to fail. 

The connecting step is only included in the Vite example to verify that all web worker config has been bundled correctly.

guillempuche:
Great @stevensJourney, I imagined this. To stop confusion, I'd add this hint to README, like linking it to this issue.

I'm working on integrating auth with Stytch https://stytch.com/ (support React and React Native). Let's see if I finish it and open source the example.

PR #126, opened 2024-04-16
Update description

benitav:
* Per a recent question in Discord, update the description of the attachments helper packages to indicate where it can be used.
* Per [this feedback](https://github.com/powersync-ja/powersync-js/issues/93), improve the minimal example project Readmes.

### DIFF
--- a/demos/example-webpack/src/index.js
+++ b/demos/example-webpack/src/index.js
@@ -31,11 +31,7 @@ let PowerSync;
 const openDatabase = async () => {
   PowerSync = new WASQLitePowerSyncDatabaseOpenFactory({
     schema: AppSchema,
-    dbFilename: 'test.sqlite',
-    flags: {
-      // This is disabled once CSR+SSR functionality is verified to be working correctly
-      disableSSRWarning: true
-    }
+    dbFilename: 'test.sqlite'
   }).getInstance();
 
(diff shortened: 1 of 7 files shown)


demos/example-webpack/src/index.js
```diff
@@ -15,7 +31,11 @@ let PowerSync;
 const openDatabase = async () => {
   PowerSync = new WASQLitePowerSyncDatabaseOpenFactory({
     schema: AppSchema,
-    dbFilename: 'test.sqlite'
+    dbFilename: 'test.sqlite',
+    flags: {
+      // This is disabled once CSR+SSR functionality is verified to be working correctly
+      disableSSRWarning: true
+    }
```
benitav, line 38: This is from the vite example, not sure if applicable here too? https://github.com/powersync-ja/powersync-js/blob/main/demos/example-vite/src/index.js#L35

rkistner: We can probably remove it from both - I don't think the flag is applicable unless using NextJS? @stevensJourney

stevensJourney: This is correct. It's only relevant when server side rendering is in use. Which is not the vase for the webpack or vite examples.

DominicGBauer (approved):

PR #134, opened 2024-04-19
fix default onError callback values for watch() and onChange()

Chriztiaan:
Fumbled `this` usage. Issue occurred when an onError handler wasn't specified for a watched query.


rkistner:
For others reading, it fixes this issue, when using `usePowerSyncWatchedQuery()` on a query with a syntax error:

image
```
logger.js:100 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'invoke')
    at error (logger.js:100:9)
    at AbstractPowerSyncDatabase.ts:729:18
```

PR #142, opened 2024-04-24
feat(react): add useQuery and useStatus hooks

DominicGBauer:
## Description
Implement new hooks

## Work Done
* Added `useQuery` hook 
* Added `useStatus` hook
* Deprecated `usePowersyncWatchedQuery`, `usePowerSyncQuery` and `usePowerSyncStatus` hooks
* Updated demos to use latest hooks

## How to test
* Load any of the demos and confirm that all items still load as expected

**Note:** The suspense and kysely hook changes will follow after this, and I will make a separate change for the Vue hooks so that they conform to this new change

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -47,10 +47,10 @@ export const TodoListDisplay = () => {
 The provided PowerSync client status is available with the `usePowerSyncStatus` hook.
 
 ```JSX
-import { usePowerSyncStatus } from "@powersync/react";
+import { useStatus } from "@powersync/react";
 
 const Component = () => {
-  const status = usePowerSyncStatus();
+  const status = useStatus();
 
   return <div>
     status.connected ? 'wifi' : 'wifi-off'
--- a/packages/react/src/hooks/useQuery.ts
+++ b/packages/react/src/hooks/useQuery.ts
@@ -1,0 +1,111 @@
+import { SQLWatchOptions } from '@powersync/common';
+import React from 'react';
+import { usePowerSync } from './PowerSyncContext';
+
+interface AdditionalOptions extends Omit<SQLWatchOptions, 'signal'> {
+  runQueryOnce?: boolean;
+}
+
+export type WatchedQueryResult<T> = {
+  data: T[];
+  /**
+   * Indicates the initial loading state (hard loading). Loading becomes false once the first set of results from the watched query is available or an error occurs.
+   */
+  isLoading: boolean;
+  error: Error;
+  /**
+   * Function used to run the query again.
+   */
+  refresh?: () => Promise<void>;
+};
+
+/**
+ * A hook to access the results of a watched query.
+ * @example
+ * ```tsx
+ * export const Component = () => {
+ * const lists = useQuery('SELECT * from lists');
+ *
+ * return <View>
+ *   {lists.map((l) => (
+ *     <Text key={l.id}>{JSON.stringify(l)}</Text>
+ *   ))}
+ * </View>
+ * }
+ * ```
+ */
+export const useQuery = <T = any>(
+  sqlStatement: string,
+  parameters: any[] = [],
+  options: AdditionalOptions = {}
+): WatchedQueryResult<T> => {
+  const powerSync = usePowerSync();
+  if (!powerSync) {
+    return { isLoading: false, data: [], error: new Error('PowerSync not configured.') };
+  }
+
+  const [data, setData] = React.useState<T[]>([]);
+  const [error, setError] = React.useState<Error>(undefined);
+  const [isLoading, setIsLoading] = React.useState(true);
+
+  const memoizedParams = React.useMemo(() => parameters, [...parameters]);
+  const memoizedOptions = React.useMemo(() => options, [JSON.stringify(options)]);
+  const abortController = React.useRef(new AbortController());
+
+  const handleResult = (result: T[]) => {
+    setIsLoading(false);
+    setData(result);
+    setError(undefined);
+  };
+
+  const handleError = (e: Error) => {
+    setIsLoading(false);
+    setData([]);
+    const wrappedError = new Error('PowerSync failed to fetch data: ' + e.message);
+    wrappedError.cause = e;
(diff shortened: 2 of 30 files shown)


packages/react/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,111 @@
...
+   */
+  refresh?: () => Promise<void>;
+};
+
+/**
+ * A hook to access the results of a watched query.
+ * @example
+ * ```tsx
+ * export const Component = () => {
+ * const lists = useQuery('SELECT * from lists');
```
stevensJourney, line 27: I think this should destructure the returned value and use the `data` member?

packages/react/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,109 @@
...
+ * const { data: lists }  = useQuery('SELECT * from lists');
+ *
+ * return <View>
+ *   {lists.map((l) => (
+ *     <Text key={l.id}>{JSON.stringify(l)}</Text>
+ *   ))}
+ * </View>
+ * }
+ */
+export const useQuery = <T = any>(
```
Chriztiaan, line 35: Nice! 
Are you planning to add the isFetching variable in a follow up PR?

DominicGBauer: Thanks! I'm not sure how it would work with the watch, how do I know when watch is running to change it to isFetching? From what I can see watch only allows onResult or onError so my assumption is that it's too late to set it then?

packages/react/README.md
```diff
@@ -47,10 +47,10 @@ export const TodoListDisplay = () => {
 The provided PowerSync client status is available with the `usePowerSyncStatus` hook.
```
Chriztiaan, line 47: I guess we need to update the readme to have an entry for useQuery(with the destructuring) and mention that watched query is deprecated?

Chriztiaan (dismissed):

stevensJourney (commented):
It would be nice to see other react hook results like `isLoading` or `hasSynced`  being used in the demos and mentioned in the readme.

packages/react/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,114 @@
...
+    } catch (e) {
+      handleError(e);
+    }
+  };
+
+  React.useEffect(() => {
+    // Abort any previous watches
+    abortController.current?.abort();
+    abortController.current = new AbortController();
+
```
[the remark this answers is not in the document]
rkistner: That's an interesting case I didn't think about, and I'm wondering whether loading should stay false in that case.

My general view of the "isLoading" use case is so that the developer can display something else when there is no data yet, which is not the case when re-fetching due to a change in the sql/parameters.

Do we have some example scenarios where the sql/parameters change?
What do other libraries do in that case?

I'd be hesitant to add a third state to distinguish first load from reload due to query change, but not sure what the other options are. There is a workaround for the developer to distinguish between those by keeping additional state in the component, but that won't work when later using suspense.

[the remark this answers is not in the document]
DominicGBauer: @rkistner my only experience with using hook queries was with RTK Query. In that package the `isLoading` was set to `true` the first time the query ran but `false` thereafter even if a subsequent async request was run. They included an `isFetching` method that is set to `true` whenever the hook is running an async request.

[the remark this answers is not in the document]
stevensJourney: An good example of where the SQL would change is the "SQL Console" view of the Todolist demo app. 

My view/understanding is that if the SQL changed it essentially changes the query to a new one which loads from scratch. 

Taking the SQL console view as an example: I would ideally like same `isLoading` to be `true` when the query changes in order to convey that the query is loading in the UI. I would also expect the previous `data` state to be reset (don't want to show old rows in the table).

This is just what I would like if I was using the API. Not sure if this is a common pattern or not.

packages/react/README.md
```diff
@@ -47,27 +47,27 @@ export const TodoListDisplay = () => {
...
```
benitav, line 47:
```suggestion
The provided PowerSync client status is available with the `useStatus` hook.
```

packages/react/src/hooks/useQuery.ts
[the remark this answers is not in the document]
rkistner: I looked at this again, and think we should go with the RTK Query behavior here (only set `isLoading: true` on the very first load).

There's a straight-forward workaround if you do want `isLoading: true` for cases where the query changes such as the SQL console: Move the query into a child component, and set `key={query}` on that component. That causes the entire hook to be re-initialized, giving `isLoading: true` again. It can't be used in all cases, but should cover most of them.

Simplified:

```js
export default function SQLConsolePage() {
  const inputRef = React.useRef<HTMLInputElement>();
  const [query, setQuery] = React.useState(DEFAULT_QUERY);

  return (<div>
            <QueryInput value={query} onChange={setQuery} />
            <!-- Setting the key here is core to controlling the isLoading behavior -->
            <QueryResults key={query} query={query} />
          </div>);
}

function QueryResults(props: { query: string }) {
  const { data, isLoading } = useQuery(props.query);

  if (isLoading) {
    return <CircularProgress />;
  }

  return <DataGrid data={data} />
}
```

Going the other direction is much more tricky, especially when using suspense.

Another option is to make this behavior configurable, but I'm not convinced we really need that.

packages/react/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,130 @@
...
+  };
+
+  React.useEffect(() => {
+    // Abort any previous watches
+    abortController.current?.abort();
+    abortController.current = new AbortController();
+    let tables = [];
+    fetchTables().then((t) => {
+      tables = t;
+    });
```
rkistner, line 102: `tables` is only set after the callback is registered, resulting in not watching any tables.

Additionally, the results should always be fetched once even if the callback is not fired.

stevensJourney (approved):
LGTM from my side

PR #149, opened 2024-04-30
fix: react hook not updating data

DominicGBauer:
## Description
Fix react hook not updating due to table not updating


packages/react/src/hooks/useQuery.ts
```diff
@@ -96,7 +96,7 @@ export const useQuery = <T = any>(
       await fetchTables();
       await fetchData();
     })();
-  }, []);
+  }, [powerSync, sqlStatement]);
```
rkistner, line 99: I didn't test, but `memoizedParams` is likely also needed here to make sure `fetchData()` is called again when it changes.

PR #150, opened 2024-04-30
feat(reat): allow compilable query as hook argument

DominicGBauer:
## Description
Allow compilable queries as hook arguments so that ORM's e.g. kysely queries can be used as arguments in place of string SQL statements

## Work Done
* Allowed useQuery to accept compilable query as an argument
* Added testing

## How to Test
* Try use a compilable query in a `useQuery` hook

### DIFF
--- a/packages/react/src/hooks/useQuery.ts
+++ b/packages/react/src/hooks/useQuery.ts
@@ -1,4 +1,4 @@
-import { type SQLWatchOptions, parseQuery, type CompilableQuery } from '@powersync/common';
+import { type SQLWatchOptions, parseQuery, type CompilableQuery, type ParsedQuery } from '@powersync/common';
 import React from 'react';
 import { usePowerSync } from './PowerSyncContext';
 
@@ -46,9 +46,12 @@ export const useQuery = <T = any>(
     return { isLoading: false, isFetching: false, data: [], error: new Error('PowerSync not configured.') };
   }
 
-  const parsedQuery = parseQuery<T>(query, parameters);
-  if (parsedQuery instanceof Error) {
-    return { isLoading: false, isFetching: false, data: [], error: parsedQuery };
+  let parsedQuery: ParsedQuery;
+  try {
+    parsedQuery = parseQuery(query, parameters);
+  } catch (error) {
+    console.error('Failed to parse query:', error);
+    return { isLoading: false, isFetching: false, data: [], error };
   }
 
   const { sqlStatement, parameters: queryParameters } = parsedQuery;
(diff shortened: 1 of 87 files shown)


packages/react/src/hooks/useQuery.ts
```diff
@@ -46,13 +46,15 @@ export const useQuery = <T = any>(
...
+  const { sqlStatement, parameters: queryParameters } = parseQuery<T>(query);
+
   const [data, setData] = React.useState<T[]>([]);
   const [error, setError] = React.useState<Error | undefined>(undefined);
   const [isLoading, setIsLoading] = React.useState(true);
   const [isFetching, setIsFetching] = React.useState(true);
   const [tables, setTables] = React.useState([]);
 
-  const memoizedParams = React.useMemo(() => parameters, [...parameters]);
+  const memoizedParams = React.useMemo(() => parameters, [...parameters, ...queryParameters]);
```
rkistner, line 57: I think in the case of a compiled query, we should error if any additional parameters are passed in, instead of just concatenating. At least in the case of Kysely, any additional parameters will cause the query to fail.

packages/react/src/hooks/useQuery.ts
```diff
@@ -37,7 +37,7 @@ export type QueryResult<T> = {
  * }
  */
 export const useQuery = <T = any>(
-  sqlStatement: string,
+  query: string | CompilableQuery<T>,
```
Chriztiaan, line 40: Worth a mention/example in the Readme?

packages/react/src/hooks/useQuery.ts
DominicGBauer: I've made a change to address this

packages/react/src/hooks/useQuery.ts
DominicGBauer: Agreed, added a mention.

packages/react/tests/useQuery.test.tsx
```diff
@@ -142,5 +143,47 @@ describe('useQuery', () => {
...
+      () => useQuery({ execute: () => [] as any, compile: () => ({ sql: 'SELECT * from lists', parameters: [] }) }),
+      { wrapper }
+    );
+    const currentResult = result.current;
+    expect(currentResult.isLoading).toEqual(true);
+  });
+
+  // The test returns unhandled errors when run with all the others.
+  // TODO: Fix the test so that there are no unhandled errors (this may be a vitest or @testing-library/react issue)
+  it.only('should show an error if parsing the query results in an error', async () => {
```
Chriztiaan: skip?

PR #152, opened 2024-05-02
fix: use correct parameters variable in memoized params

DominicGBauer:


rkistner (commented):
Similar issue on line 88:

```ts
await powerSync.getAll<T>(sqlStatement, parameters);
```

rkistner (approved):

PR #153, opened 2024-05-03
feat(vue): update vue composables and compilable queries

Chriztiaan:
Updated composable package to use same signatures as the new hooks from `@powersync/react` https://github.com/powersync-ja/powersync-js/pull/142. Added the same compilable query support added to React hooks in https://github.com/powersync-ja/powersync-js/pull/150.

Introduces `useQuery` and `useStatus` composables that replace the existing `usePowerSyncWatchedQuery`, `usePowersyncQuery`, and `usePowerSyncStatus`.

Verified that the Kysely driver works in the Vue package with `useQuery`, and that watches trigger when dependent tables change.

The following snippet was used to confirm:
```
import { usePowerSync, useQuery } from '@powersync/vue';
import { wrapPowerSyncWithKysely } from '@powersync/kysely-driver';
import { Database } from '@/library/powersync/AppSchema';

const powerSync = usePowerSync();
const db = wrapPowerSyncWithKysely<Database>(powerSync.value);

const {
  data: querySQLResult,
  isLoading,
  isFetching,
  error
} = useQuery(db.selectFrom('lists').selectAll().where('name', 'like', '%Shopping%'));
```


packages/vue/src/composables/useQuery.ts
```diff
@@ -0,0 +1,156 @@
...
+    try {
+      const result = await powerSync.value.getAll<T>(sql, parameters);
+      handleResult(result);
+    } catch (e) {
+      console.error('Failed to fetch data:', e);
+      handleError(e);
+    }
+  };
+
+  let abortController = new AbortController();
```
DominicGBauer, line 101: why can't you define this in the watchEffect?

Chriztiaan: I guess you can. I was simply using the react package as a reference.
I also kinda prefer the current version to something like:
```
 watchEffect(async (onCleanup) => {
    onCleanup(() => abortController.abort());
    let abortController = new AbortController();
    abortController = new AbortController();
```

Maybe this leaves the door open to allowing users to cancel/pause watch themselves later?

DominicGBauer: It doesn't really make sense right now, and since it isn't really doing anything it should probably be removed. If we include cancel/pausing at a later stage then we can reinclude it?

DominicGBauer: Looking at the react one the `abortController.current?.abort();`abort is running when the useEffect is run to avoid having multiple instances.

Chriztiaan: It's definitely needed to ensure the onChanged is cleaned up when the sql/parameters change.

Chriztiaan: Sorted

DominicGBauer (approved):

stevensJourney (approved):

PR #151, opened 2024-04-30
Diagnostics app

benitav:
Initial release of tools/diagnostics-app

PR #159, opened 2024-05-08
[diagnostics-app] Fix sync status

rkistner:
The diagnostics app uses a custom instance of `WebStreamingSyncImplementation`, so the sync status is not reported on the database, and `useStatus()` doesn't work correctly.

Reverts https://github.com/powersync-ja/powersync-js/pull/151/commits/c19246c5a3cece2519dad46a8ce51a4ff51cbea4.


tools/diagnostics-app/src/app/views/layout.tsx
```diff
@@ -86,6 +86,16 @@ export default function ViewsLayout({ children }: { children: React.ReactNode })
...
 
+  // Cannot use `useStatus()`, since we're not using the default sync implementation.
+  React.useEffect(() => {
+    const l = sync.registerListener({
+      statusChanged: (status) => {
+        setSyncStatus(status);
+      }
+    });
+    return () => l();
+  }, [powerSync]);
```
stevensJourney, line 97: Should the dependency be `sync`?

rkistner: `sync` is a constant, so fixed it to just not have any dependencies.

stevensJourney (approved):

PR #160, opened 2024-05-08
Remove uuid and react-native-get-random-values dependencies

rkistner:
This helps to reduce build issues, especially for react-native which needs `react-native-get-random-values` for that.

Most of the places we use it is in-memory only can be replaced by a simple incrementing integer id.

`deleteBucket` is slightly more complex, but we can let SQLite generate a uuid for us in that case.

Some demos still use `uuid` in other places, but none of the core packages should need it anymore.

Also see https://github.com/powersync-ja/react-native-quick-sqlite/pull/31.

### DIFF
--- a/demos/react-native-supabase-group-chat/package.json
+++ b/demos/react-native-supabase-group-chat/package.json
@@ -49,7 +49,6 @@
     "react-native": "0.73.4",
     "react-native-fetch-api": "^3.0.0",
     "react-native-gesture-handler": "~2.14.1",
-    "react-native-get-random-values": "~1.8.0",
     "react-native-pager-view": "6.2.3",
     "react-native-polyfill-globals": "^3.1.0",
     "react-native-reanimated": "~3.6.2",
--- a/packages/web/src/worker/db/open-db.ts
+++ b/packages/web/src/worker/db/open-db.ts
@@ -1,7 +1,6 @@
 import * as SQLite from '@journeyapps/wa-sqlite';
 import '@journeyapps/wa-sqlite';
 import * as Comlink from 'comlink';
-import { v4 as uuid } from 'uuid';
 import { QueryResult } from '@powersync/common';
 
 export type WASQLExecuteResult = Omit<QueryResult, 'rows'> & {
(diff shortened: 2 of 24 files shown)


demos/react-native-supabase-group-chat/package.json
```diff
@@ -49,7 +49,6 @@
     "react-native": "0.73.4",
     "react-native-fetch-api": "^3.0.0",
     "react-native-gesture-handler": "~2.14.1",
-    "react-native-get-random-values": "~1.8.0",
```
Chriztiaan, line 52: Can we do the same for react-native-supabase-todo?

rkistner: That would basically be replacing it with `expo-crypto` - I guess that's slightly better (less issues)?

rkistner: Did that now.

I do see that `react-native-polyfill-globals` still depends on `react-native-get-random-values`, but I hope there is less potential for weird version conflicts now.

Chriztiaan:
`react-native-supabase-todolist` and `react-supabase-todolist` work on my side 👍

stevensJourney:
We'll also have to remove the uuid requirement from our React Native Quick SQLite package. I'll do that quickly.

Chriztiaan (approved):

packages/web/src/worker/db/open-db.ts
```diff
@@ -40,14 +41,14 @@ export async function _openDB(dbFileName: string): Promise<DBWorkerInterface> {
...
-  const listeners = new Map<string, OnTableChangeCallback>();
+  const listeners = new Map<number, OnTableChangeCallback>();
 
   sqlite3.register_table_onchange_hook(db, (opType: number, tableName: string, rowId: number) => {
     Array.from(listeners.values()).forEach((l) => l(opType, tableName, rowId));
   });
 
   const registerOnTableChange = (callback: OnTableChangeCallback) => {
-    const id = uuid();
+    const id = nextId++;
```
stevensJourney, line 51: It's probably unlikely, but it seems like JavaScript does not overflow to `0`, like other low-level lanuages, if an integer increases over `Number.MAX_SAFE_INTEGER`. Should we be manually rolling over to `0`?

rkistner: `Number.MAX_SAFE_INTEGER` is 2^53. Even at a rate of a million per second, it will take over 280 years to reach that limit.

stevensJourney (approved):
We should add changesets to track this change

stevensJourney (approved):

PR #161, opened 2024-05-09
chore: Exclude sourceMaps from prod builds and introduced base tsconfig

Chriztiaan:
This fixes a reported issue with vite not being able to resolve sourcemaps in projects outside of the `ps-js` monorepo.

Sourcemaps are excluded for production builds of the packages only and are still available for local development in `ps-js`.

Introduced a tsconfig.base in the root of the repo which gets extended by all the packages. I tried to not make any improvements, simply extracted the common config. I attempted to add `strictNullChecks` and `declarations` to the base config but that caused issues with some of the packages/demos(which use the packages). To simplify this diff I opted to leave that work for a future PR.


demos/react-native-supabase-todolist/package.json
```diff
@@ -5,8 +5,7 @@
   "scripts": {
     "android": "expo run:android",
     "ios": "expo run:ios",
-    "clean": "watchman watch-del-all && rm -rf .expo",
-    "preinstall": "cd ../.. && pnpm build:packages"
```
Chriztiaan, line 9: This causes a build:package to occur after running pnpm install in the root. Bad side effect that I believe I introduced a while ago.

stevensJourney (dismissed):

stevensJourney (approved):

PR #169, opened 2024-05-14
chore: updated vite confs to work outside of monorepo

Chriztiaan:
The current demo vite.conf files caused errors when used in demos outside of this monorepo.


stevensJourney (approved):
Thank you sir!

PR #171, opened 2024-05-16
chore: add common as peer dependency to packages

DominicGBauer:
## Description
Add common to peerDependency of packages so that users are warned if common is mismatched under breaking change.

There is no need to have common a dependency anymore as peerDeps are auto installed in npm and pnpm and a warning is given in yarn.

**Example warning in NPM and Yarn (using an example from the internet)**:
image

image

**Notes:**
* There is a possibility to use changesets to update peerDeps but that appears to have issues as outlined here https://github.com/changesets/changesets/issues/1011
* NPM, PNPM and Yarn handle these differently. There seems to be an issue with PNPM not adhering to the peerDeps as mentioned here [pnpm does not fail when missing peer deps with strict-peer-dependencies](https://github.com/pnpm/pnpm/issues/6893)

### DIFF
--- a/.changeset/tasty-teachers-switch.md
+++ b/.changeset/tasty-teachers-switch.md
@@ -1,0 +1,10 @@
+---
+"@powersync/kysely-driver": patch
+"@powersync/react-native": patch
+"@powersync/attachments": patch
+"@powersync/react": patch
+"@powersync/vue": patch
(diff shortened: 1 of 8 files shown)


stevensJourney (approved):
Looks good to me

PR #172, opened 2024-05-16
[Fix] Improve bson usage

stevensJourney:
# Description

Some issues were raised with the recent addition of the `bson` package for WebSocket streaming.

## Vite warnings
A warning for `crypto.randomBytes` being used would be shown in Vite applications even if the WebSockets mode was not used.

The warning is now removed due to using a dynamic import of `bson` in the Web SDK. The Vite config `resolve` setting is now no longer required.
 
## Webpack example

The example Webpack and NextJS demos was broken due to bundling issues. BSON requires certain dependencies to be externalised. These have been added to the example config. These are not required for NextJS configs.

## React Native

In some React Native environments the wrong import might be used, causing top level await issues. The React Native import has been fixed according to [these](https://github.com/mongodb/js-bson?tab=readme-ov-file#react-native) docs.

# Testing

This was tested on all the demos with HTTP streaming and Web socket modes enabled.

NextJS demo working for demonstration:
image


demos/example-electron/vite.renderer.config.ts
```diff
@@ -30,7 +30,6 @@ export default defineConfig((env) => {
         '@powersync/web > js-logger',
         '@powersync/web > lodash/throttle',
         '@powersync/web > can-ndjson-stream',
-        '@powersync/web > bson',
```
stevensJourney, line 33: This is no longer needed in our Vite configs

demos/react-supabase-todolist/package.json
```diff
@@ -11,12 +11,15 @@
   "dependencies": {
     "@powersync/react": "workspace:*",
     "@powersync/web": "workspace:*",
+    "@emotion/react": "11.11.4",
```
stevensJourney, line 14: Added these as it was necessary when copying this project our to a standalone demo folder.

packages/common/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -171,13 +173,20 @@ export abstract class AbstractRemote {
...
+  abstract getBSON(): Promise<BSONImplementation>;
+
   /**
    * Connects to the sync/stream websocket endpoint
    */
   async socketStream(options: SyncStreamOptions): Promise<DataStream<StreamingSyncLine>> {
     const { path } = options;
     const request = await this.buildRequest(path);
 
+    const BSON = await this.getBSON();
```
DominicGBauer, line 188: can this be lower case so that it doesn't get confused with BSON type?

stevensJourney: Can do that. It's now lowercase.

DominicGBauer (approved):

Chriztiaan (approved):

PR #175, opened 2024-05-17
fix: add changeset config and update package.json to avoid versioning issues

DominicGBauer:
## Description
There is an issue where packages a major version bumped due to peer dependency changes.

This PR fixes this by setting a changeset config flag to only major version bump when the peerDependency becomes out of range.

The range has also changed from `1.x` to `^1` so that when the changeset happens it allows all minor versions instead of hard setting the package version. 

Example of the update when using `npx changeset version`:

image


.changeset/config.json
```diff
@@ -7,5 +7,8 @@
   "access": "restricted",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": []
+  "ignore": [],
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
```
stevensJourney, line 11: Should we lock the version of `@changesets/cli`?

stevensJourney: `WILL_CHANGE_IN_PATCH` makes me think it will change in patch versions.

DominicGBauer: Yeah not worth the risk. Makes sense to rather pin it then and upgrade separately if we need any new features of changeset. Will add change.

stevensJourney (approved):

PR #176, opened 2024-05-20
chore: update dependency range and changeset config

DominicGBauer:
## Description
Add `"updateInternalDependents": "out-of-range"` option to changeset config so that underlying dependencies (e.g. `@powersync/react` are only updated when `@powersync/common` is major version updated.

I have further changed the dependency range to be more specific by using `>= 1.7.0 < 2`.

This has been tested locally to verify that it works as expected. 
**Example**:  if you have 
```  
  "name": "@powersync/react",
  "version": "1.3.3",
"peerDependencies": {
    "@powersync/common": "workspace:>= 1.7.0 < 2"
  },
```
And you do a minor or patch version update to `@powersync/common` the package that has the above peerDependencies will not require a version bump. However if you do a major version bump to `@powersync/common` then it will be changed to 
```  
  "name": "@powersync/react",
  "version": "2.0.0",
"peerDependencies": {
    "@powersync/common": "workspace:>=2.0.0"
  },
```
and require a major version bump of the dependent package.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -9,6 +9,7 @@
   "updateInternalDependencies": "patch",
   "ignore": [],
   "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
-    "onlyUpdatePeerDependentsWhenOutOfRange": true
+    "onlyUpdatePeerDependentsWhenOutOfRange": true,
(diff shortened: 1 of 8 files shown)


stevensJourney (approved):
LGTM

PR #178, opened 2024-05-21
feat: allow for web sdk to be used without web worker

DominicGBauer:
## Description
Allow for the option to choose if you want the web SDK to use web workers. This will resolve the issue found while creating the demo of the capacitor app.

## Work Done
* Created a useWebWorker flag that is set to true by default
* Moved usage of web worker related functionality to be hidden behind the flag
* Adapted Web SDK to run without being in a wen worker
* Cleaned up some of the imports
* Updated the tests (might need more here?)

## Video
React app where useWebWorker is set to false

https://github.com/powersync-ja/powersync-js/assets/46312751/1c376595-761d-447f-a8cd-aa9599fae4cb


packages/web/tests/main.test.ts
```diff
@@ -8,18 +8,81 @@ type User = {
...
+    db = factory.getInstance();
+  });
+
+  afterEach(async () => {
+    await db.disconnectAndClear();
+    await db.close();
+  });
+
+  describe('executeQuery', () => {
+    it('should execute a select query using getAll', async () => {
```
stevensJourney, line 41: Do we need to duplicate these tests just to test with a different flag enabled on the DB? Is it not possible to declare the test logic once and use that on different DBs?

packages/web/src/db/adapters/AbstractWebPowerSyncDatabaseOpenFactory.ts
```diff
@@ -20,6 +20,7 @@ export const DEFAULT_POWERSYNC_FLAGS: WebPowerSyncOpenFlags = {
    * Multiple tabs are by default not supported on Android, iOS and Safari.
    * Other platforms will have multiple tabs enabled by default.
    */
+  useWebWorker: true,
```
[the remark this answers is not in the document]
stevensJourney: This still seems to be a comment about multiple tabs?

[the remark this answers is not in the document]
DominicGBauer: It's outdated, I moved `useWebWorker` above the comment if you look at `packages/web/src/db/adapters/AbstractWebPowerSyncDatabaseOpenFactory.ts` in Files you will see this.

packages/web/tests/main.test.ts
DominicGBauer: I think I'm confused by what you are asking. Do you want there to be 1 test that does the same thing for different configurations of the DB so that we would have a test that tests two different DB's in a single test?

stevensJourney: That is roughly in the direction of what I was suggesting. Currently it's the same test logic being used on two different DB configurations in two different (manually specified) test sets. There should not be a need to duplicate the test logic.

DominicGBauer: Done

packages/web/tests/main.test.ts
```diff
@@ -1,67 +1,62 @@
...
+      const result = await db.getAll('SELECT * FROM customers');
       expect(result.length).toEqual(0);
     });
 
-    it('should allow inserts', async () => {
+    itWithDBs('should allow inserts', async () => {
       const testName = 'Steven';
-      await db.execute('INSERT INTO users (id, name) VALUES(?, ?)', [uuid(), testName]);
-      const result = await db.get<User>('SELECT * FROM users');
+      await dbWithWebWorker.execute('INSERT INTO customers (id, name) VALUES(?, ?)', [uuid(), testName]);
```
stevensJourney, line 39: these tests aren't using the correct `db` provided as a parameter in the callback

DominicGBauer: Sorry about that, PR updated

PR #179, opened 2024-05-21
chore: reinclude common package as a dependency

DominicGBauer:
## Description
Reinclude `@powersync/common` package as a dependency to downstream dependencies so that we do not rely on the auto installation of peer dependencies.

### DIFF
--- a/packages/attachments/package.json
+++ b/packages/attachments/package.json
@@ -28,5 +28,8 @@
   },
   "peerDependencies": {
     "@powersync/common": "workspace:>=1.9.0"
+  },
+  "dependencies": {
+    "@powersync/common": "workspace:*"
   }
 }
--- a/packages/react-native/package.json
+++ b/packages/react-native/package.json
@@ -35,6 +35,7 @@
   },
   "dependencies": {
     "@powersync/react": "workspace:*",
+    "@powersync/common": "workspace:*",
     "async-lock": "^1.4.0",
     "bson": "^6.6.0"
   },
--- a/.changeset/sour-coats-lay.md
+++ b/.changeset/sour-coats-lay.md
@@ -1,0 +1,10 @@
+---
+'@powersync/kysely-driver': patch
+'@powersync/react-native': patch
+'@powersync/attachments': patch
+'@powersync/react': patch
+'@powersync/vue': patch
+'@powersync/web': patch
+---
+
+Reinclude common package as a dependency
--- a/packages/kysely-driver/package.json
+++ b/packages/kysely-driver/package.json
@@ -28,7 +28,8 @@
     "@powersync/common": "workspace:>=1.9.0"
   },
   "dependencies": {
-    "kysely": "^0.27.2"
+    "kysely": "^0.27.2",
+    "@powersync/common": "workspace:*"
   },
   "devDependencies": {
     "@powersync/web": "workspace:*",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -31,6 +31,9 @@
     "react": "*",
     "@powersync/common": "workspace:>=1.9.0"
   },
+  "dependencies": {
+    "@powersync/common": "workspace:*"
+  },
   "devDependencies": {
     "@testing-library/react": "^15.0.2",
     "@types/react": "^18.2.34",
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -31,6 +31,9 @@
     "vue": "*",
     "@powersync/common": "workspace:>=1.9.0"
   },
+  "dependencies": {
+    "@powersync/common": "workspace:*"
+  },
   "devDependencies": {
     "flush-promises": "^1.0.2",
     "jsdom": "^24.0.0",
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -37,6 +37,7 @@
     "@powersync/common": "workspace:>=1.9.0"
   },
   "dependencies": {
+    "@powersync/common": "workspace:*",
     "async-mutex": "^0.4.0",
     "buffer": "^6.0.3",
     "bson": "^6.6.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1176,7 +1176,7 @@ importers:
   packages/attachments:
     dependencies:
       '@powersync/common':
-        specifier: workspace:>=1.9.0
+        specifier: workspace:*
         version: link:../common
 
   packages/common:
@@ -1231,7 +1231,7 @@ importers:
   packages/kysely-driver:
     dependencies:
       '@powersync/common':
-        specifier: workspace:>=1.9.0
+        specifier: workspace:*
         version: link:../common
       kysely:
         specifier: ^0.27.2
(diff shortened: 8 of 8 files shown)


packages/attachments/package.json
```diff
@@ -28,5 +28,8 @@
   },
   "peerDependencies": {
     "@powersync/common": "workspace:>=1.9.0"
+  },
+  "dependencies": {
+    "@powersync/common": "workspace:*"
```
stevensJourney, line 33: Why is this a peer dependency and a dependency? Why does the peer dependency have a version range, while the dependency resolves to the fixed version?

DominicGBauer: The peer Dependency tells the user what valid ranges of the dependency are that they need to have installed for the package to work while the dependency is using whatever the latest version is that we have in the repo. This way if the package manager is using an older version of the package because of some other install the user will be notified about it. 

I'm not sure exactly how all the package managers resolve different versions of installed dependencies so having the peer dependency range is a fail safe to avoid an issue where a package might use the wrong version of @powersync/common without at least warning the user

stevensJourney: I might be wrong, but this approach seems to be more likely to cause multiple versions of `@powersync/common` being installed in a project due to the `workspace:*` range. I get the impression that `@powersync/common` should only be a peer dependency for `kysely-driver`, `react`, `vue` etc... while it should only be a dependency of `@powersync/web` and `@powersync/react-native`.

DominicGBauer: Why would it be different for web and react native?

rkistner: I'm with @stevensJourney on this.

`@powersync/web` actively

rkistner: I'm with @stevensJourney on this.

`@powersync/web` basically "warps" `@powersync/common`. You expect to be able to just install `@powersync/web` and use it, without adding any additional dependencies. (Same for `@powersync/react-native`). So for this, a normal dependency makes sense.

For `@powersync/react` etc - you can't use it as a standalone package - you use it together with `@powersync/web` or `@powersync/react-native`. A `peerDependency` here says "you need these packages additionally", and that still holds even if it's not installed automatically.

So overall, you could have say `@powersync/web` and `@powersync/react` as direct dependencies. `@powersync/common` would be automatically installed due to it being a dependency of `@powersync/web`, while still satisfying the peerDependendency of `@powersync/react`.

One thing I'm not 100% sure of is whether the above actually works with `@powersync/common` being a transitive dependency and not a direct dependency of the app. We should test this properly with different package managers and options, e.g. pnpm's `strict-peer-dependencies` option.

stevensJourney: The current use case for developers is to install the SDK package as the first step. The other packages are optional helpers which would be installed after the SDK. If something has to actually provide the `common` package (without having to install it separately) I think it would be the SDK package(s).

packages/attachments/package.json
```diff
@@ -28,5 +28,8 @@
   },
   "peerDependencies": {
     "@powersync/common": "workspace:>=1.9.0"
```
rkistner, line 30: While we're looking at these dependencies, should this rather be `workspace:^1.9.0`? (That's equivalent to `>=1.9.0 <2`)

DominicGBauer: @rkistner there seems to be an issue with changesets where it automatically changes this. You can see it in this one https://github.com/powersync-ja/powersync-js/pull/173/files if you look at for example `packages/attachments/package.json`. I've raised an issue with changesets to hopefully get more info

packages/attachments/package.json
DominicGBauer: Okay cool I'll adapt based on the above. Thanks guys!

packages/attachments/package.json
rkistner: Do changesets perhaps handle the `^1.9.0` syntax better?

DominicGBauer: I tested it locally and that works. Will also add that in.

packages/attachments/package.json
DominicGBauer: I tested installing web and react only in an empty repo for pnpm yarn and react using the changes in this PR in a dev publication of the packages. common is installed in all 3 cases.

packages/react-native/package.json
```diff
@@ -31,10 +31,11 @@
     "react": "*",
     "react-native": "*",
     "react-native-polyfill-globals": "^3.1.0",
-    "@powersync/common": "workspace:>=1.9.0"
+    "@powersync/common": "workspace:^1.9.0"
   },
   "dependencies": {
     "@powersync/react": "workspace:*",
+    "@powersync/common": "workspace:*",
```
stevensJourney, line 38: Should the version range here also be `workspace:^1.9.0`?

packages/attachments/package.json
DominicGBauer: It worked when I added
```
auto-install-peers = false
```
and
```
omit = peer
```
to `.npmrc` for npm and pnpm. Yarn didn't auto install before but it is installed now that it is a dep of web

packages/attachments/package.json
DominicGBauer: It worked when I added
```
auto-install-peers = false
```
and
```
omit = peer
```
to `.npmrc` for npm and pnpm. Yarn didn't auto install before but it is installed now that it is a dep of web

DominicGBauer: I have now also made a new repo housing the react-demo app and tested the following:
```
pnpm i && pnpm build && pnpm start
yarn && yarn build && yarn start
npm i && npm run build && npm run start
```

In all 3 cases there were no errors or type issues.

Are there any other tests you would like me to run?

packages/react-native/package.json
rkistner: I think the react-native (and web) packages should probably depend on the specific version (which `workspace:*` does), unless we're very careful about stability of the interface between common and the sdks.

stevensJourney: That makes sense. I like that.

packages/attachments/package.json
stevensJourney: Nothing from my side that I can think of.

stevensJourney (approved):

packages/react-native/package.json
DominicGBauer: So I tested this by changing @powersync/common to v2.0.0 and v1.8.0. An error will be thrown that looks like this 
```
 ERR_PNPM_NO_MATCHING_VERSION_INSIDE_WORKSPACE  In ../attachments: No matching version found for @powersync/common@workspace:^1.9.0 inside the workspace
```
v1.9.1 would install (as expected)

is this the behaviour we want to happen?

stevensJourney: I think that happens since the `attachments` package is using the `workspace:^1.9.0` version specifier for `@powersync/common`. If you change the version of `@powersync/common` out of that range in the workspace then it can't find the matching version in the workspace (needs to satisfy both the `workspace` and version range). 
This should be normal.

PR #190, opened 2024-05-30
update: added a mechanism for throttling callback executions

Chriztiaan:
# Description
In scenarios where a watch()'s onChange callback is triggered frequently or in a short time span, invoking the same query can cause multiple redundant query executions. 

This PR introduces a throttle mechanism that queues the onChange invocations and effectively discards any redundant queries between the one currently executing and latest queued query. For example if 5 changes are picked up (A, B, C, D, E) at almost the same time, A will start executing, B will be queued, overwritten by C, which is overwritten by D, which is finally overwritten by E. So just A and E would execute (verified by manually testing in the React and Vue todo apps and the unit test included in this PR.

This can be a nice improvement for bigger datasets.

# Testing of PR 
Insert data and confirm that everything works, maybe insert 1k, 5k and 10k?

# Minor benchmarks
Some numbers against the React todo app where I inserted X amount of todos in Supabase and observed the data syncing to the application:

For insert ops there were about 10-13 queries discarded.
5k ops about 45 queries discarded
10k ops threw out about 120-130

Test inverse, with throttling disabled how many query execs do we normally get?:
1000: 11-14 execs
5k: 50-55
10k: 133-140

So it does seem like there's a good ratio of redundant queries thrown out.

### DIFF
--- a/.changeset/forty-avocados-yell.md
+++ b/.changeset/forty-avocados-yell.md
@@ -1,0 +1,5 @@
+---
+'@powersync/web': major
+---
+
+Bump version for Stable release.
--- a/.changeset/itchy-gifts-dance.md
+++ b/.changeset/itchy-gifts-dance.md
@@ -1,0 +1,5 @@
+---
+'@powersync/common': minor
+---
+
+Added a mechanism for throttling watch callback executions.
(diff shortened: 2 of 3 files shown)


Chriztiaan:
FYI @rkistner

.changeset/forty-avocados-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/web': major
+---
+
+Bump version for Production release.
```
benitav, line 5:
```suggestion
Bump version for Stable release.
```

stevensJourney (approved):
Looks good to me!

benitav (approved):

PR #204, opened 2024-06-06
update: added user parameters to PowerSyncCredentials which are available in sync rules.

Chriztiaan:
This PR introduces the ability to pass parameters from a client SDK to the sync rules. The idea is to specify parameters to the PowerSync `connect` method's options . Backend changes to support this are available [here](https://github.com/powersync-ja/powersync-service/pull/7). 

```
export const SystemProvider = ()=> {
  const [connector] = React.useState(new DemoConnector());
  const [powerSync] = React.useState(db);

  React.useEffect(() => {
  
    powerSync.init();
    powerSync.connect(connector, {
      params: { device_id: "276fca51-041e-4061-b7e3-3e2e699b40b4" }  // <- Specifying user parameters
    });
  }, [powerSync, connector]);
  ```

Parameters are available in the sync rules under `user_parameters` (alongside the already supported `token_parameters`).

For example:
```
bucket_definitions:
  user_lists:
    parameters: select id as list_id from lists where owner_id = user_parameters.device_id
    data:
      - SELECT * FROM lists where id = bucket.list_id
````

# Testing
Verified values pull through via self-host project.

### CODE REVIEW

File: packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts [hunk 1]
```diff
@@ -431,6 +431,7 @@ export abstract class AbstractStreamingSyncImplementation
         let appliedCheckpoint: Checkpoint | null = null;
 
         let bucketSet = new Set<string>(initialBuckets.keys());
+        const { params = undefined } = (await this.options.remote.getCredentials()) ?? {};
```
Chriztiaan, line 434: I couldn't find a handle on the params upstream in the stack trace. If `getCredentials` is already called higher up in the stack, we could pass the value down instead.

rkistner (commented):
Could we instead pass the parameters when calling `connect()`?

The credentials are fetched at non-deterministic times, which can make it difficult to update the parameters.

Chriztiaan:
> Could we instead pass the parameters when calling `connect()`?
> 
> The credentials are fetched at non-deterministic times, which can make it difficult to update the parameters.

Sounds good

Chriztiaan:
@rkistner, I've updated the PR to use `connect` instead, updated the description's example.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -73,6 +73,11 @@ export interface PowerSyncConnectionOptions {
    * Defaults to a HTTP streaming connection.
    */
   connectionMethod?: SyncStreamConnectionMethod;
+
+  /**
+   * These parameters are passed to the sync rules, and will be available under the`user_parameters` object.
+   */
+  params?: Record<string, any>;
```
stevensJourney, line 80: Could we make the `any` a bit more restricted? I'm not sure if we allow nested objects here or not?

Reason for mentioning is in the case of web these params need to be cloneable in order to pass to the shared web worker over the message port. There are also probably some restrictions in general for making the actual network request.

Chriztiaan: After testing locally, I can confirm that arrays and nested objects aren't supported in the sync rule at the moment.
I have restricted the param types to `string | number | boolean | null | undefined`.

Should this be applied to the service's [parameter type](https://github.com/powersync-ja/powersync-service/pull/7/files#diff-e9ececf52e503dd0bd1b31b471ba5e113f040c75cfa8641a391190e4eb5e9f51R97) as well?

stevensJourney: `string | number | boolean | null | undefined` is probably okay in the Client for now - until when or if we do support arrays and objects on the server side.

I think it would be good to show a type error if trying to use types which aren't yet supported. I'm not sure how much into detail we want to go for the JavaScript use case - i.e. if a user is not using TS we could parse and check this at runtime or the server could return a validation error.

rkistner: Sync rules does indirectly support nested objects and arrays, in the same way that SQLite does: It is represented as a JSON string. `normalizeTokenParameters` already takes care of this on the service. So arbitrary nested values should be fine here - anything that can be serialized as JSON. Then in sync rules you can access it as for example:

```sql
-- params: { some_object: { nested_property: 'test' }}
select user_parameters.some_object ->> 'nested_property' as id
-- params: { project_ids: [1,2,3] } 
select project.id from projects where project.id in params.project_ids
```

Chriztiaan:
Deployed the service changes to staging and confirmed that the `powersync-js` react todo app works when providing the parameters during connections.

image
image

rkistner (approved):

PR #209, opened 2024-06-10
[diagnostics-app] Include empty buckets in sync diagnostics

rkistner:
This PR will let the service include empty buckets in the sync stream: https://github.com/powersync-ja/powersync-service/pull/12

This change here will let the diagnostics app display those empty buckets like this:
image

This will help to diagnose sync issues, by having a distinction between issues in parameter queries (which could cause the bucket to not be synced) versus issues in data queries (which could cause no data within the bucket).


stevensJourney (dismissed):
Can we add a changeset to the `diagnostics-app` folder to result in this change being in the changelog?

stevensJourney (approved):

PR #217, opened 2024-06-28
Django RN demo app: fix some issues picked up during Django backend cleanup

kobiebotha:
Builds on work in https://github.com/powersync-ja/powersync-django-backend-todolist-demo/pull/3

- [ ] useState in HeaderWidget is currently broken
- [ ] use latest React hooks and context providers

### DIFF
--- a/demos/django-react-native-todolist/library/powersync/AppSchema.ts
+++ b/demos/django-react-native-todolist/library/powersync/AppSchema.ts
@@ -2,9 +2,10 @@ import { Column, ColumnType, Index, IndexedColumn, Schema, Table } from '@powers
 
 export const AppSchema = new Schema([
   new Table({
-    name: 'api_todo',
+    name: 'todos',
     columns: [
       new Column({ name: 'list_id', type: ColumnType.TEXT }),
+      new Column({ name: 'photo_id', type: ColumnType.TEXT }),
       new Column({ name: 'created_at', type: ColumnType.TEXT }),
       new Column({ name: 'completed_at', type: ColumnType.TEXT }),
       new Column({ name: 'description', type: ColumnType.TEXT }),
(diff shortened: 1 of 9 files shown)


demos/django-react-native-todolist/app/index.tsx
```diff
@@ -10,34 +8,19 @@ import Logger from 'js-logger';
...
-const App = observer(() => {
-  const { djangoConnector } = useSystem();
-
+const App = () => {
   React.useEffect(() => {
     Logger.useDefaults();
     Logger.setLevel(Logger.DEBUG);
 
-    const getSession = async () => {
-      const response = await fetch('http://127.0.0.1:8000/api/get_session', {
```
stevensJourney, line 21: Previously the URL was hardcoded, but besides that the `get_session` endpoint always returns a valid session. Which makes this logic not really have an affect. 

The user should now be able to register and login - although that is not strictly required by the backend (it will create tokens for PowerSync either way). This login screen process does at least keep this demo in line with other Todolist demos. Correct session management can be added later. Users will unfortunately have to signin each time the app boots. Alternatively signin could just be dropped altogether.

demos/django-react-native-todolist/library/powersync/AppSchema.ts
```diff
@@ -2,20 +2,26 @@ import { Column, ColumnType, Index, IndexedColumn, Schema, Table } from '@powers
 
 export const AppSchema = new Schema([
```
DominicGBauer, line 3: can this be changed to use the V2 table?

stevensJourney: Yeah can do that. This now uses `TableV2`

DominicGBauer (approved):

PR #220, opened 2024-07-09
chore: update readme and update to PowerSyncDatabase

DominicGBauer:
## Description
* Improve React README
* Make changes to use PowerSyncDatabase that were missed

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -7,6 +7,8 @@ Configure a PowerSync DB connection and add it to a context provider.
 ```JSX
 // App.jsx
 import { PowerSyncDatabase } from '@powersync/react-native';
+// or for Web
+// import { PowerSyncDatabase } from '@powersync/web';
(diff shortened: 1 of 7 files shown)


packages/react/README.md
```diff
@@ -7,6 +7,8 @@ Configure a PowerSync DB connection and add it to a context provider.
 ```JSX
 // App.jsx
 import { PowerSyncDatabase } from '@powersync/react-native';
+// or for Web
+// import { PowerSyncDatabase } from '@powersync/web';
```
Chriztiaan, line 11: Nit, but as this is for the `react` package, I would prioritise the `/web` import and comment out the RN one.

Chriztiaan (approved):

packages/react/README.md
DominicGBauer: I was thinking the same thing but wasn't sure. This has been the decider will make the same change

stevensJourney (approved):

PR #221, opened 2024-07-09
[Feature] Prebundling packages

Chriztiaan:
This PR shows an initial stab at prebundling packages so that end users don't need to install polyfills. In the context of web projects that would be `Buffer` and in the case of React Native the list is a [bit longer](https://github.com/powersync-ja/powersync-js/tree/main/packages/react-native#install-polyfills). This basically includes the polyfills as part of the package.

1. After getting `@powersync/react-native`'s Rollup setup working (against both monorepo and standalone copies of the todo app), I could confirm that this allows a RN project to keep working without having to apply the respective polyfills.

2. Next focus was `@powersync/web`, kept getting stuck with bundling the workers (code not included for now). Decided to park Rollup build here for now.

3. Moved to `@powersync/common`, which is upstream to both `react-native` and `web` and should be good enough to drop the `Buffer` polyfill. Got to a point where I could setup `common` such that it works for the existing non-rollup web build (confirmed that buffer isn't needed), but not with the rollup react-native build.

# PoC Summary:
- `React` works with `rollup-common`(for browser) + existing web (No more buffer polyfill) 
- `React native` works with `existing-common` + `rollup react-native`
--Either get @ps/common issue with Text column error or "globalObject.fetch.bind is not a function (it is undefined), js engine: hermes"
- React Native breaks with `rollup-common` + `rollup react-native`

So if we can figure out `rollup common > existing web` working alongside `rollup common > rollup react-native`, this should be good enough for now. Might need to outputs in the `rollup common` config.

## Further todos

- [ ] Update README entries wrt polyfills
- [ ] Update todo apps affected (should be most if not all - showcasing changes in react/RN in this) polyfill deps and code


packages/react-native/src/sync/stream/ReactNativeRemote.ts
```diff
@@ -31,55 +31,6 @@ class ReactNativeFetchProvider extends FetchImplementationProvider {
   }
 }
 
-type PolyfillTest = {
```
Chriztiaan, line 34: Because Rollup's inject plugin replaces calls to these members instead of supporting global usages, these tests always fail.

PR #223, opened 2024-07-10
SupabaseConnector upload queue error code fix

Chriztiaan:
It seems that the example SupabaseConnector files have morphoed a bit over time and that a bug slipped through to a few of them which caused table operations with errors never being removed from the upload queue.

This PR ensures that the thrown error contains the postgres error code (which was left out when making a new `Error`) which is then evaluated. Verified that bad operations (like RLS violations) are removed from the upload queue.

To easily reproduce insert a list/todo with a random owner UUID not related to the logged in user.

# Side note: 
Is there merit in having an importable SupabaseConnector or flag that can be enabled on a PS instance that uses a bundled SB connector? Or is having these explicit copies of the Connector good for illustration?


rkistner (dismissed):

rkistner:
> Is there merit in having an importable SupabaseConnector or flag that can be enabled on a PS instance that uses a bundled SB connector? Or is having these explicit copies of the Connector good for illustration?

This is something to investigate for the future.

Right now I prefer the explicit copies, to make it clear that it is the responsibility of the app to do these updates, not the SDK. The example connectors do not handle cases such as a column removed on the server, table/column name being different on the client and server, or using compound primary keys on the server. So rather than having those reported as SDK bugs, the cases should be handled directly in the app code currently.

In the future we can look into options for having a standard but configurable connector, with well-documented behavior and limitations.

rkistner (approved):

PR #224, opened 2024-07-12
chore(BREAKING): rename final @journeyapps packages to @powersync

DominicGBauer:
## Description
* Find and replace all @journeyapps references with @powersync in package.json, README's and pods

Before this goes in we should update all the reference docs in gitbook or anywhere else you can think of @benitav

Todos: 
- [ ] Update reference docs in gitbook


stevensJourney (dismissed):

Chriztiaan:
Closing this for now. We are still using `@journeyapps/wa-sqlite` over `@powersync/wa-sqlite`.

PR #226, opened 2024-07-16
Fix web lock issues

rkistner:
`WASQLiteDBAdapter` did not correctly lock read statements (`get`, `getAll` and `getOptional`). This meant these statements could be run in the middle of another transaction and return inconsistent results.

This also changes the global lock in `open-db.ts` to a local Mutex, since it is only relevant for that specific connection. This may give a minor performance improvement - navigator.locks can be slow when tabs/workers are busy.


packages/kysely-driver/tests/sqlite/sqlite-connection.test.ts
```diff
@@ -35,7 +34,7 @@ describe('PowerSyncConnection', () => {
 
       expect(rows.length).toEqual(1);
       expect(rows[0].name).toEqual('John');
-      expect(getAllSpy).toHaveBeenCalledTimes(1);
+      expect(getAllSpy).toHaveBeenCalledWith('SELECT * From users', []);
```
rkistner, line 37: The watch from `updateHasSynced` (`SELECT 1 FROM ps_buckets WHERE last_applied_op > 0 LIMIT 1`) interfered with the `toHaveBeenCalledTimes(1)` test. The test worked previously, but the lock fix changed the timing of the call, which caused this test to break.

This now just tests that it has been called correctly once, and ignores any other calls.

stevensJourney (approved):

PR #229, opened 2024-07-16
Add a debugMode flag to log queries to the performance timeline

rkistner:
This displays all SQL queries on the performance timeline. Currently only implemented in the Web SDK.

This excludes the time waiting for the global transaction lock, but includes all overhead in worker communication. This means you won't see concurrent queries in most cases.

This includes internal statements from PowerSync, including queries saving sync data, and begin/commit statements. It does not include internal statements from powersync-sqlite-core.

TODO:
 * ~The flag currently defaults to false if `process.env.NODE_ENV == 'production'`, similar to what React does. All other cases default to true. Should we perhaps be more conservative and require explicit opt-in, since the queries could be considered semi-sensitive data?~ Updated default to false, requiring explicit opt-in for now.

Example of the diagnostics app during initial sync:

image

### DIFF
--- a/packages/common/src/client/SQLOpenFactory.ts
+++ b/packages/common/src/client/SQLOpenFactory.ts
@@ -13,8 +13,11 @@ export interface SQLOpenOptions {
   /**
    * Enable debugMode to log queries to the performance timeline.
    *
-   * Defaults to false if `process.env.NODE_ENV == 'production'`,
-   * true otherwise.
+   * Defaults to false.
+   *
+   * To enable in development builds, use:
+   *
+   *    debugMode: process.env.NODE_ENV !== 'production'
    */
   debugMode?: boolean;
 }
--- a/packages/common/src/client/SQLOpenFactory.ts
+++ b/packages/common/src/client/SQLOpenFactory.ts
@@ -9,6 +9,14 @@ export interface SQLOpenOptions {
    * Directory where the database file is located.
    */
   dbLocation?: string;
+
+  /**
+   * Enable debugMode to log queries to the performance timeline.
+   *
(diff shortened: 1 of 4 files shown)


packages/common/src/client/SQLOpenFactory.ts
```diff
@@ -9,6 +9,14 @@ export interface SQLOpenOptions {
...
    */
   dbLocation?: string;
+
+  /**
+   * Enable debugMode to log queries to the performance timeline.
+   *
+   * Defaults to false if `process.env.NODE_ENV == 'production'`,
+   * true otherwise.
+   */
+  debugMode?: boolean;
```
rkistner, line 19: @stevensJourney Is this okay, or is there a better place to define this option?

stevensJourney: This is where I'd recommend to put it.

stevensJourney (approved):
LGTM . This is an awesome feature.

stevensJourney (approved):

guillempuche:
Is it documented?

rkistner:
@guillempuche It's not properly documented yet, but you can see how it's used in the diagnostics app:

https://github.com/powersync-ja/powersync-js/blob/05f3dbd25b697dee0214e5bafe852d0055c9b736/tools/diagnostics-app/src/library/powersync/ConnectionManager.ts#L26-L32

I'd recommend using something like this to only enable debugMode in development:

```js
debugMode: process.env.NODE_ENV != 'production'
```

The feature just displays queries in the "Performance" tab of Chrome's developer tools (after recording) - see the screenshots above.

guillempuche:
Thanks Ralf @rkistner

PR #230, opened 2024-07-16
[Experimental] Log profiling output to the console

rkistner:
Builds on #229.

This outputs any queries over a certain threshold to the console when `debugMode` is enabled.

If the runtime was over another (higher) threshold, EXPLAIN QUERY PLAN output is additionally added.

This can be quite verbose, and the thresholds are rather arbitrary. We'd need to consider what should be logged by default, and what should be configurable.

Sample output for the diagnostics app:
image
---
image


rkistner:
Superseded by #760

PR #234, opened 2024-07-22
Isolated Demo Tests

stevensJourney:
# Overview

We've noticed that demo sometimes behave differently outside the monorepo. This seems to be due to our monorepo package hoisting configuration.

This adds an automated test which will move the demos to standalone test projects for testing.

The current workspace packages are published to a locally hosted NPM registry. These packages are used when installing the demo's dependencies.

This tests if the demo's `install` script and `test:build` script. A `test:build` script has been added to some demos which specifies the test criteria for building the demo.

Only some of the Demos currently have functional build tests. Other demos can be added in future PRs.

The action shows a summary of all demos which were tested. A warning is shown if the demo does not have a `test:build` script.

image


.github/workflows/test-isolated.yaml
```diff
@@ -0,0 +1,68 @@
...
+          sleep 10 # Give Verdaccio some time to start
+
+      - name: Prepare For Test Publish
+        run: npx tsx scripts/test-publish-helper.ts
+
+      - name: Registry login
+        run: npx npm-cli-login -u test -p test -e test@example.com -r http://localhost:4873 
+
+      - name: Config Temporary Registry
+        run: echo "registry=http://localhost:4873" >> ~/.npmrc
```
rkistner, line 59: Is it feasible to use this registry only for the `@powersync` scope, instead of proxying? Not sure if that will make a big difference, but it feels a little cleaner.

stevensJourney: Yes I think that should be possible. I agree that would feel cleaner.

stevensJourney: This has been implemented now

Chriztiaan (dismissed):

rkistner (approved):

Chriztiaan (approved):

Issue #231, opened 2024-07-17
Conflicting peer dependencies of @powersync/common


hut36:
When installing both @powersync/web and @powersync/attachments as dependencies, I ran into this error:

```
dependencies:
+ @powersync/attachments 2.0.10
+ @powersync/web 1.2.1

 WARN  Issues with peer dependencies found
.
├─┬ @powersync/web 1.2.1
│ └── ✕ missing peer @powersync/common@^1.13.01.13.0
└─┬ @powersync/attachments 2.0.10
  └── ✕ missing peer @powersync/common@^1.9.0
✕ Conflicting peer dependencies:
  @powersync/common
```

The latest good @powersync/web version is 1.0.2. I'm using pnpm 8.15.8.

hut36:
Installing only @powersync/web results the same error. The `peerDependencies` in file package.json (https://www.npmjs.com/package/@powersync/web?activeTab=code) looks apparently wrong:

```
  "peerDependencies": {
    "@journeyapps/wa-sqlite": "~0.2.0",
    "@powersync/common": "^1.13.01.13.0"
  },
```

DominicGBauer:
Hi @hut36 apologies for the delay. I am investigating this issue now.

DominicGBauer:
Hi @hut36 this is now resolved in the latest package versions e.g. https://www.npmjs.com/package/@powersync/web?activeTab=code, thanks for pointing this issue out.

PR #235, opened 2024-07-22
chore: update @powersync/common peer dep range

DominicGBauer:
## Description
There appears to be an issue with pnpm update (https://github.com/pnpm/pnpm/pull/7633/files) which has resulted in `workspace` changing behaviour when publishing. As a result it duplicates the package version resulting in something like `1.3.0^1.3.0`. We are therefore changing the `peerDeps` range to use `^` and removing the version number i.e. no longer `^1.3.0` and just `^`.

When the packages are updated this should resolve this https://github.com/powersync-ja/powersync-js/issues/231

### DIFF
--- a/.changeset/healthy-insects-push.md
+++ b/.changeset/healthy-insects-push.md
@@ -1,0 +1,10 @@
+---
+'@powersync/kysely-driver': patch
+'@powersync/react-native': patch
+'@powersync/attachments': patch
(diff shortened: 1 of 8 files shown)


stevensJourney (approved):
Looks good to me

Chriztiaan (approved):

PR #243, opened 2024-07-25
Docker image for diagnostics-app

rkistner:
Automates a docker image build for the diagnostics app, using nginx to serve the static assets.

See discussion on https://github.com/powersync-ja/self-host-demo/issues/23.

### DIFF
--- a/.github/workflows/diagnostics-image-release.yaml
+++ b/.github/workflows/diagnostics-image-release.yaml
@@ -1,0 +1,56 @@
+#  Publishes the Docker image to DockerHub
+#  This is triggered whenever the `service` is versioned and tagged
+name: Docker Image Release
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - '@powersync/diagnostics-app*'
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  release-docker-image:
+    name: Build and Release diagnostics-app Docker Image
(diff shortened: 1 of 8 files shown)


.github/workflows/diagnostics-image-release.yaml
```diff
@@ -0,0 +1,53 @@
+#  Publishes the Diagnostics App Docker image to DockerHub
+#  This is triggered whenever the `diagnostics-app` package is versioned and tagged
+name: Diagnostics Image Release
+
+on:
+  workflow_dispatch:
+  push:
```
stevensJourney, line 7: We should probably remove the `push` here. Or update it to only run when the "package" is tagged.

The main issue is that DockerHub will let you overwrite the existing version when a publish is performed. 

Running on every push would update the published image with dev content.

Running on pushes to just the `main` branch are also not ideal since the Changesets versioning process will only contain versions bumps after the `Version` PR is merged to `main`. This means that if there are code changes that have been merged, but not versioned, then the image will be updated with the next version's content.

rkistner: > We should probably remove the push here. Or update it to only run when the "package" is tagged.

That was just temporary to test the workflow before merging. Reverted that change now.

stevensJourney: 👍

stevensJourney (approved):
LGTM!

PR #211, opened 2024-06-14
[Feature]: Improve client constructors

stevensJourney:
# Overview

This POC updates the `PowerSyncDatabase` constructor to behave a bit more similarly to the Dart SDK (within some limitations). 

The constructor previously relied on an instance of `DBAdapter` to be supplied. This caused the creation of a `PowerSyncDatabase` to require being wrapped inside a factory which generated the `DBAdapter`. Typically this required the use of `RNQSPowerSyncDatabaseOpenFactory` in React Native and `WASQLitePowerSyncDatabaseOpenFactory` in Web to create the `DBAdapter` and return an instance of `PowerSyncDatabase`. 

It's now possible to construct `PowerSyncDatabase` using the options from the `AbstractPowerSyncOpenFactory` directly.

This is achieved by moving out the logic from `AbstractPowerSyncOpenFactory` (and it's platform variants) which created SQLite connections to a light-weight `SQLOpenFactory` which just opens a `DBAdapter`. The PowerSync open factories still use this same logic internally, but are now deprecated in favour of using the  `PowerSyncDatabase` constructor directly.

A PowerSyncDatabase client can now be constructed by either passing it a `DBAdapter` which will directly be used internally; a `SQLOpenFactory` which will be used to create a `DBAdapter`; or by passing paramaters which will be passed to a default `SQLOpenFactory`.

This results in users being able to create clients as follows:

```diff
- export const db = new WASQLitePowerSyncDatabaseOpenFactory({
-  dbFilename: 'example.db',
-  schema: AppSchema
- }).getInstance();
+ export const db = new PowerSyncDatabase({
+ schema: AppSchema,
+ database: {
+  dbFilename: 'example.db'
+ }
+ });
```

## Testing

All demos have been updated to use the new constructor method. The demos were started locally to verify everything works correctly. 

Development packages for the SDKs have been published and installed inside an isolated copy of the React Todolist application. The latest development`@powersync/web` package SDK client was shown to be compatible with the latest stable `@powersync/react` package. 

All unit tests were updated to use the constructor method for creating PowerSync clients. An additional test has been added to ensure that the deprecated factory methods still function correctly.

PR #252, opened 2024-07-29
[Fix] SQL options

stevensJourney:
# Overview

This fixes a regression from https://github.com/powersync-ja/powersync-js/pull/211. 

The PR above introduced the ability to directly construct a `PowerSyncDatabase`. The `AbstractPowerSyncDatabase` accepts either an instance of `DBAdapter`, a SQLite open factory which will be used to open a `DBAdapter` or options to be used with a default SQLite open factory.

If options are provided the `AbstractPowerSyncDatabase` constructor will call the abstract `openDBAdapter` method with `SQLOpenOptions`. 

Web implementations require additional `flags` when opening SQLite connections. The implementation previously assumed the `resolvedFlags` were available in the `openAdapter` method. This is not the case since the `super()` method is called before the resolved flags are assigned. This would cause the supplied flags to be ignored when creating DB connections. In practise this disregarded the `useWebWorker` flag.

This PR updates the internal and `protected` `openAdapter` method to have access to all the `options` provided to the `PowerSyncDatabase` constructor. 

Unit tests have been added to ensure workers are correctly used given the relevant `flags`.


demos/example-electron/package.json
```diff
@@ -19,6 +19,8 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@emotion/react": "^11.13.0",
```
stevensJourney, line 22: I noticed this was required in a standalone demo during testing.

rkistner (approved):

DominicGBauer (approved):

PR #253, opened 2024-08-01
[Fix] isolate test for react-supabase-todolist

Chriztiaan:
So I played around with fixing the `react-supabase-todolist` isolate test and it seems for the standalone version it has an issue which a few of the other demos have in that its tsconfig has a "references: ../../packages/web"  entry which you need to remove to make the standalone copy work (which is one more step for people setting up their own projects based on our demos).
The current plan is to remove it from the codebase, and have entries like these be re-added via a post-install script when developing the demos/SDKs inside the monorepo. 

To get this working for the isolate test workflow, the post install script is ignored.


Chriztiaan:
This still leaves concerns about having to modify the tsconfigs in the demo which need to be discarded. Ideally it could be done with a config that the IDE picks up in the monorepo - no success yet.

Alternative option so far would be to remove the references entirely and not have the the benefits of the references (which feels lame).

Issue #247, opened 2024-07-25
Memory leak when `transaction.complete()` is not awaited


byzyk:
Failure to call `await transaction.complete()` inside `uploadData()` results in an endless loop and eventually hangs the browser.

I understand that it's not intended behavior, so I would appreciate it if we could find a way to detect the memory leak and throw an exception instead. It would potentially save many dev hours for those who'll make the same mistake as I just did 😃

PS. Amazing library, I enjoy it quite a lot the past few days!

stevensJourney:
Thanks for logging this issue. We'll investigate by reproducing on our side.

stevensJourney:
Hi @byzyk, I tried to reproduce a leak and browser hang on my side by removing [this](https://github.com/powersync-ja/powersync-js/blob/8fe44fbe7a880350afd4bd7518fe5845e83cb98d/demos/react-supabase-todolist/src/library/powersync/SupabaseConnector.ts#L141) line in our demo.

I did notice that the upload function remains in a _non-blocking_ loop, but I didn't notice indefinite increasing memory usage or a hang. Attached below is the profile of memory usage which did remain stable.
![image](https://github.com/user-attachments/assets/824347c3-a57d-4c54-ab66-9370545cd1ce)

We are still investigating ways to potentially detect if the transaction is not completed and avoid the loop. It would be helpful if you could share some more context on how the hang or leak occurred on your side.

stevensJourney:
We've added some additional checks in these versions https://github.com/powersync-ja/powersync-js/pull/255 to delay infinite loops if no uploads are performed in the `uploadData` method of the `BackendConnector`. This should help a lot with these kinds of issues. Generally an infinite loop could cause large memory spikes depending on what the client code is doing in `uploadData`. I'm going to close this for now since I could not reproduce a memory leak on my side.

byzyk:
hi @stevensJourney, thanks for the update.

I just wanted to clarify from my side - you are correct, what was happening here was an infinite loop. In my case, I had a memory leak probably due to some additional logic inside `uploadData`. Thanks for patching this!

PR #254, opened 2024-08-01
[Feature] Add warning if crud transactions are not completed

stevensJourney:
# Overview

Currently when processing CRUD items the `PowerSyncDatabase` client will call the backend connector's `uploadData` function whenever `CrudEntry` items are present in the `ps_crud` internal table.

The `uploadData` callback has access to the `PowerSyncDatabase` client which it can use to get the pending `CrudEntry`s. The connector can consume the CRUD events in numerous different ways. 

Typically the `uploadData` callback will either call `getNextCrudTransaction` or `getCrudBatch` on the `PowerSyncDatabase`. 

- `getNextCrudTransaction` will return the next (Single) group of `CrudEntry`s which occurred in the same transaction as an instance of `CrudTransaction`. If an  operation did not occur in a transaction then only a single operation will be returned. 
- `getCrudBatch` returns `n` `CrudEntry`s as an instance of `CrudBatch` . These entires are not grouped by transaction.

The `CrudBatch` and `CrudTransaction` classes each have a `complete` method which will remove the `CrudEntry`s from the CRUD queue. This should be executed in the `uploadData` method after the items have successfully been uploaded.

Given the nature above, there may be more items in the queue than what was consumed by a single invocation of the  `uploadData` method. In this case the SDK will continuously check if `CrudEntry`s are present in the queue and call `uploadData` until the queue is empty. 

Issues can occur if the `uploadData` callback fetches CRUD items and uploads them without calling and awaiting the `CrudTransaction` or `CrudBatch` `complete` method.

Failing to `await` a `complete()` call in the `uploadData` method is mostly guarded in the SDK due to limits on concurrent SQLite operations. Previously the `SqliteBucketStorage` implementation used `.execute` methods for checking if CRUD items were available. Internal queues would (luckily) ensure the next upload iteration would essentially await for the `complete` action to complete. 

Failing to call the `complete()` method altogether is a much worse scenario. This would result in `CrudEntry`s not being removed from the queue (after they have been uploaded). The SDK previously would blindly check if `CrudEntry`s were still present and immediately call `uploadData` to upload them - this will result in an infinite loop of repeated uploads.

This PR replaces the use of the `hasCrud` check with getting the `CrudEntry` which is first in line to be uploaded (the behaviour of `hasCrud` is matched if the entry exists). The SDK can then keep track of this `CrudEntry` in the upload loop. If it sees the same entry that it already processed previously a warning will be printed and the upload loop will be delayed in order to slow down the infinite loop. This does not currently prevent the loop from continuing. 

This should help for issues such as https://github.com/powersync-ja/powersync-js/issues/247.

### DIFF
--- a/packages/common/src/client/sync/bucket/SqliteBucketStorage.ts
+++ b/packages/common/src/client/sync/bucket/SqliteBucketStorage.ts
@@ -1,5 +1,7 @@
 import { Mutex } from 'async-mutex';
+import Logger, { ILogger } from 'js-logger';
 import { DBAdapter, Transaction, extractTableUpdates } from '../../../db/DBAdapter';
+import { BaseObserver } from '../../../utils/BaseObserver';
 import {
   BucketState,
   BucketStorageAdapter,
@@ -8,12 +10,10 @@ import {
   PSInternalTable,
   SyncLocalDatabaseResult
 } from './BucketStorageAdapter';
-import { OpTypeEnum } from './OpType';
 import { CrudBatch } from './CrudBatch';
-import { CrudEntry } from './CrudEntry';
+import { CrudEntry, CrudEntryJSON } from './CrudEntry';
+import { OpTypeEnum } from './OpType';
 import { SyncDataBatch } from './SyncDataBatch';
-import Logger, { ILogger } from 'js-logger';
-import { BaseObserver } from '../../../utils/BaseObserver';
 
 const COMPACT_OPERATION_INTERVAL = 1_000;
 
@@ -51,10 +51,10 @@ export class SqliteBucketStorage extends BaseObserver<BucketStorageListener> imp
 
   async init() {
     this._hasCompletedSync = false;
-    const existingTableRows = await this.db.execute(
+    const existingTableRows = await this.db.getAll<{ name: string }>(
       `SELECT name FROM sqlite_master WHERE type='table' AND name GLOB 'ps_data_*'`
     );
-    for (const row of existingTableRows.rows?._array ?? []) {
+    for (const row of existingTableRows ?? []) {
       this.tableNames.add(row.name);
     }
   }
@@ -72,10 +72,10 @@ export class SqliteBucketStorage extends BaseObserver<BucketStorageListener> imp
   startSession(): void {}
 
   async getBucketStates(): Promise<BucketState[]> {
-    const result = await this.db.execute(
+    const result = await this.db.getAll<BucketState>(
       'SELECT name as bucket, cast(last_op as TEXT) as op_id FROM ps_buckets WHERE pending_delete = 0'
     );
-    return result.rows?._array ?? [];
+    return result;
   }
 
   async saveSyncData(batch: SyncDataBatch) {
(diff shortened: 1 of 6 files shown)


packages/web/tests/multiple_instances.test.ts
```diff
@@ -184,34 +186,44 @@ describe('Multiple Instances', () => {
...
+      retryDelayMs: 100,
+      flags: {
+        broadcastLogs: true
+      }
+    });
 
     // Generate the second streaming sync implementation
     const connector2 = new TestConnector();
-    const spy2 = vi.spyOn(connector2, 'uploadData');
+    // The second connector will be called first to upload, we don't want it to actually upload
```
stevensJourney, line 205: This test was actually previously vulnerable to a race condition where the `dispose` of `stream2` would happen during the invocation of the upload function. This would cause `connector2` to complete the upload transaction at indeterministic times. 

Note that these `uploadData` functions aren't actually awaited. The new checks in `SqliteBucketStorage` seem to catch these issues better now.

packages/web/tests/multiple_instances.test.ts
```diff
@@ -184,34 +186,44 @@ describe('Multiple Instances', () => {
...
       remote: new WebRemote(connector1),
       uploadCrud: async () => {
         triggerUpload2();
         connector2.uploadData(db);
       },
-      identifier
-    };
-    const stream2 = new SharedWebStreamingSyncImplementation(syncOptions2);
+      identifier,
+      retryDelayMs: 100,
```
stevensJourney, line 222: Updated this delay since if the second connector doesn't actually complete the transaction then there will be a delay in the upload loop.

rkistner (approved):

packages/common/src/client/sync/bucket/SqliteBucketStorage.ts
```diff
@@ -318,10 +327,10 @@ export class SqliteBucketStorage extends BaseObserver<BucketStorageListener> imp
       return null;
     }
 
-    const crudResult = await this.db.execute('SELECT * FROM ps_crud ORDER BY id ASC LIMIT ?', [limit]);
+    const crudResult = await this.db.getAll<CrudEntryJSON>('SELECT * FROM ps_crud ORDER BY id ASC LIMIT ?', [limit]);
```
DominicGBauer, line 330: same here

packages/common/src/client/sync/bucket/SqliteBucketStorage.ts
```diff
@@ -304,9 +305,17 @@ export class SqliteBucketStorage extends BaseObserver<BucketStorageListener> imp
     });
   }
 
+  async nextCrudItem(): Promise<CrudEntry | undefined> {
+    const next = await this.db.getOptional<CrudEntryJSON>('SELECT * FROM ps_crud ORDER BY id ASC LIMIT 1');
```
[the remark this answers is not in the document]
rkistner: Yes, the ids are an auto-incrementing sequence defining the queue order (first-in-first-out queue):
https://github.com/powersync-ja/powersync-sqlite-core/blob/5704c7fb689b84f8679b006023f3fd0858fd75ff/crates/core/src/view_admin.rs#L164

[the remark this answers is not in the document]
DominicGBauer: Epic thanks!

DominicGBauer (approved):

Chriztiaan (approved):

PR #261, opened 2024-08-12
[Fix] hasSynced not resetting after a disconnectAndClear() call

Chriztiaan:
This is a minor iteration on the work done [here](https://github.com/powersync-ja/powersync-js/pull/257). With this change the hasSynced value exposed by the status hooks/listeners should be correctly reset to undefined after doing a disconnectAndClear. 

This can be verified by the following flow in the react supabase todo app:
1. Login
2. Verify hasSynced state goes from undefined/falsey to truthy.
3. Logout
4. Login
5. Verify hasSynced state goes from undefined/falsey to truthy. (This would previously start on truthy).

Todo
- [x] Remove watch usage


packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -299,36 +297,13 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
...
-      [],
-      {
-        onResult: (result) => {
-          const hasSynced = !!result.rows?.length;
-
-          if (hasSynced != this.currentStatus.hasSynced) {
-            this.currentStatus = new SyncStatus({ ...this.currentStatus.toJSON(), hasSynced });
-            this.iterateListeners((l) => l.statusChanged?.(this.currentStatus));
-          }
+    const result = await this.getAll<any>(syncedSQL);
```
stevensJourney, line 300: Nit: you could use `getOptional` for this. It doesn't look like `syncedSQL` is being used anywhere else but in this call. The SQL could be in the call.

stevensJourney (approved):

Issue #262, opened 2024-08-12
Kysely support for react-native-web


jasonlewicki:
Running on:
```
"@powersync/common": "0.0.0-dev-20240812065227",
"@powersync/react": "0.0.0-dev-20240812065227",
"@powersync/kysely-driver": "0.0.0-dev-20240812065227",
"@powersync/react-native": "0.0.0-dev-20240812065227",
"@powersync/attachments": "0.0.0-dev-20240812065227",
"@powersync/web": "0.0.0-dev-20240812065227"
```

Looks like the powersync kysely driver package is pointing to the esm build instead of cjs
```
remapping @powersync/web
error: node_modules/kysely/dist/esm/migration/file-migration-provider.js: node_modules/kysely/dist/esm/migration/file-migration-provider.js:Invalid call at line 32: import( /* webpackIgnore: true */this.#props.path.join(this.#props.migrationFolder, fileName))
Error: node_modules/kysely/dist/esm/migration/file-migration-provider.js:Invalid call at line 32: import( /* webpackIgnore: true */this.#props.path.join(this.#props.migrationFolder, fileName))
    at transformJS (/Users/*****/node_modules/@expo/metro-config/build/transform-worker/metro-transform-worker.js:251:23)
    at transformJSWithBabel (/Users/*****/node_modules/@expo/metro-config/build/transform-worker/metro-transform-worker.js:352:18)
remapping @powersync/web
```

stevensJourney:
I tried adding a CommonJS export to our Kysely driver package, but in testing I noticed that the Metro bundler was still attempting to bundle the ESM export of the `kysely` package. We might be able to make this easier by bundling Kysely into our driver package.

For now the easiest solution is to configure the Metro bundler to use the CommonJS version of Kysely

in the `metro.config.js` file:
```js
config.resolver.resolveRequest = (context, moduleName, platform) => {
  if (platform === 'web') {
    if (['react-native-prompt-android', '@powersync/react-native'].includes(moduleName)) {
      return {
        type: 'empty'
      };
    }
    const mapping = {
      'react-native': 'react-native-web',
      '@powersync/web': '@powersync/web/dist/index.umd.js',
      kysely: 'kysely/dist/cjs/index.js'
    };
    if (mapping[moduleName]) {
      console.log('remapping', moduleName);
      return context.resolveRequest(context, mapping[moduleName], platform);
    }
  } else {
    if (['@powersync/web'].includes(moduleName)) {
      return {
[...]

PR #263, opened 2024-08-13
[Feature] Kysely CommonJs Export

stevensJourney:
# Overview 

This adds a CommonJS export for the `@powersync/kysely-driver` package. Previously only a ESM export was provided. 

React Native typically transpiles `import` statements to `require` statements, but this might be fickle in some cases - potentially in React Native web. 

## How this works

See [this](https://nodejs.org/api/packages.html#conditional-exports) for reference. The output of this module has changed to:

image

The cjs export will be used if the module is required e.g. `const {} = require('@powersync/kysely-driver')`. The ESM export will be used if the module is imported e.g. `import {} from '@powersync/kysely-driver';`

The CommonJS export is transpiled via an additional `tsconfig.json` file. The contents of this file use `require` statements for internal modules. This means this package will require CJS versions of those modules if they provide them. Kysely provides CJS [exports](https://github.com/kysely-org/kysely/blob/43ae280bf8d3bcc0ed10461ba2a2a1e6e577de76/package.json#L15).


stevensJourney:
This should help with https://github.com/powersync-ja/powersync-js/issues/262

stevensJourney:
Closing this for now. The original purpose of this was to help support React Native web, but React Native web was incorrectly importing the `kysely` package (not this driver package).

PR #266, opened 2024-08-13
[web] debounce update notifications

rkistner:
Firing update notifications for each individual row adds a massive performance overhead.

This change debounces the notifications, only sending a single event per table changed (we never exposed the row ids of changes to the end-user API anyway).

In some test scenarios this gives a 20-25x speedup in initial sync times, for example going from 240s -> 10s for 40k operations.


stevensJourney (approved):
This approach looks good to me. I can't think of anything this might break. It's an amazing performance increase.

PR #267, opened 2024-08-13
Update Supabase demo to warn about `js-logger` when running Vite

guillempuche:
Prevent this issue:
```
Uncaught SyntaxError: The requested module 'http://localhost:3000/@fs/.../<repo>/node_modules/js-logger/src/logger.js?v=eb4a8b9e' doesn't provide an export named: 'default'
```

Discussion started on Discord https://discord.com/channels/1138230179878154300/1272981660589363293


romatallinn:
I needed to include `js-logger`, even having it installed in `package.json`. Can you please double check this?

guillempuche:
In my case, I didn't install it.

romatallinn:
> In my case, I didn't install it.

Yeah well I mean pharsing `Only when 'js-logger' isn't installed in package.json` suggests that you don't need it when `js-logger` installed. But it seems that it's needed even when it's installed. So maybe just recommend to always include it? Or when you install it, it's not breaking without `include`?

guillempuche:
How about now?

Chriztiaan:
So turns out you need to install js-logger and configure/apply it, alternatively you can have the include rule. I think we should just add the include rule everywhere for the moment. @guillempuche would you mind applying it to all the vite projects when you have a moment?

guillempuche:
@Chriztiaan, done.

Chriztiaan:
Looks like there is work being done on this [here](https://github.com/vitejs/vite/pull/17837) to resolve the underlying exclude issue.

Chriztiaan (approved):

stevensJourney (approved):

PR #274, opened 2024-08-19
[Feature] Added PowerSyncDatabase options validation

stevensJourney:
# Overview

This adds some very basic checks for invalid options to the `PowerSyncDatabase` constructor. 

JavaScript users might incorrectly supply required options or fail to supply them at all. This should help those users identify issues. 

The warnings here are not detailed. But should be better than cryptic errors which could occur down-the-line if invalid arguments are provided.

### DIFF
--- a/packages/common/src/client/AbstractPowerSyncDatabase.ts
+++ b/packages/common/src/client/AbstractPowerSyncDatabase.ts
@@ -182,7 +182,12 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
   constructor(options: PowerSyncDatabaseOptions); // Note this is important for extending this class and maintaining API compatibility
   constructor(protected options: PowerSyncDatabaseOptions) {
     super();
-    const { database } = options;
+
+    const { database, schema } = options;
+
+    if (false == schema instanceof Schema) {
+      throw new Error('The `schema` option should be provided and should be an instance of `Schema`.');
+    }
 
     if (isDBAdapter(database)) {
       this._database = database;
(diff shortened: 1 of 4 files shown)


Chriztiaan (approved):

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -182,21 +182,28 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
...
+    }
 
     if (isDBAdapter(database)) {
       this._database = database;
     } else if (isSQLOpenFactory(database)) {
       this._database = database.openDB();
     } else if (isPowerSyncDatabaseOptionsWithSettings(options)) {
       this._database = this.openDBAdapter(options);
+    } else {
+      throw new Error('The provided `database` option is invalid.');
```
DominicGBauer, line 199: is there no way to give a more specific error?

stevensJourney: I specifically don't want to go into the rabbit hole of validating this - at least for now. There are too many permutations of incorrect options.

DominicGBauer (approved):

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -182,21 +182,28 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
   constructor(options: PowerSyncDatabaseOptions); // Note this is important for extending this class and maintaining API compatibility
   constructor(protected options: PowerSyncDatabaseOptions) {
     super();
-    const { database } = options;
+
+    const { database, schema } = options;
+
+    if (false == schema instanceof Schema) {
```
rkistner, line 188: An `instanceof` check like this can cause issues in edge cases.

One I've seen is someone writing reusable code for both `@powersync/web` and `@powersync/react-native`, with the schema just using one of the above libs. They can then end up with different copies of the Schema class, which will fail the check.

Other similar cases can occur when just having multiple versions of `@powersync/common` installed.

rkistner (approved):

PR #277, opened 2024-08-22
chore(common): add max column check

DominicGBauer:
## Description
Users currently only see an error for maxmium number of columns at runtime. This adds a check so that they would see it on Table creation.

## Work Done
* Added checks to `TableV2` and added testing
* Added a check to `Table` (left out tests since we are going to remove this in next major release)
* Updated to latest `vitest`

### DIFF
--- a/packages/common/src/db/schema/Table.ts
+++ b/packages/common/src/db/schema/Table.ts
@@ -105,7 +105,7 @@ export class Table {
       throw new Error(`Invalid characters in view name: ${this.viewNameOverride}`);
     }
 
-    if(this.columns.length >= MAX_AMOUNT_OF_COLUMNS) {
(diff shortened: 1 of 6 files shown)


packages/common/src/db/schema/Table.ts
```diff
@@ -103,11 +105,15 @@ export class Table {
       throw new Error(`Invalid characters in view name: ${this.viewNameOverride}`);
     }
 
+    if(this.columns.length >= MAX_AMOUNT_OF_COLUMNS) {
```
rkistner, line 108:
```suggestion
    if(this.columns.length > MAX_AMOUNT_OF_COLUMNS) {
```

rkistner (approved):

PR #279, opened 2024-08-23
[fix] Electron Demo Multiple Tabs

stevensJourney:
# Overview

Electron production builds' watched queries currently fail when enabling multiple tabs.

Watches fail to trigger due to multiple Shared WA-SQLite web workers being created in production builds. Usually a single shared WA-SQLite worker exists. This single worker caches a single SQLite connection. Table change notifications are shared by all clients connected to the shared worker. Multiple SQLite connections are created if multiple shared workers are created - these connections do not share table change notifications.

Multiple workers seem to be created due to the app HTML+js assets being served over a `file://` origin. This works around the origin issue by serving the production assets via an Express server. I'm not sure if there are any caveats to this approach, but it works.

## Testing 

This was verified locally with a standalone build of this demo. This was also verified [here](https://discord.com/channels/1138230179878154300/1266087750075940976/1276650512992829601).


DominicGBauer (approved):
Crazy that `file://` casuses multiple worker instances. Nice catch, must have been a mission to figure out image

Chriztiaan (approved):

stevensJourney:
> Crazy that `file://` casuses multiple worker instances. Nice catch, must have been a mission to figure out image

It's even more difficult to catch since you can't (as) easily debug or inspect shared workers in Electron's dev tools.

nicreuss:
@stevensJourney in your PR you referenced [this issue](https://github.com/electron/electron/issues/13952). There is a [comment](https://github.com/electron/electron/issues/13952#issuecomment-414403732) on the issue which suggests that registering a custom protocol may bypass the limitation of `file://`. I'm curious to know if you tried that approach too in your debugging?

stevensJourney:
> @stevensJourney in your PR you referenced [this issue](https://github.com/electron/electron/issues/13952). There is a [comment](https://github.com/electron/electron/issues/13952#issuecomment-414403732) on the issue which suggests that registering a custom protocol may bypass the limitation of `file://`. I'm curious to know if you tried that approach too in your debugging?

@nicreuss I did not investigate that approach much. Some of the other comments on that thread made it seem like using a custom protocol did not help in their use cases.

PR #280, opened 2024-08-26
React native web support via webpack 

Chriztiaan:
This introduces a webpack build step in `web` that takes the `lib` contents created by `tsc` and processes it further to add support for `UMD` (written to the `dist` dir). Most projects will still depend on the `tsc` output (`@powersync/web`, but projects can opt to use `@powersync/web/umd`.

An example use case for this is projects that want to leverage `react-native-web`. The `react-native-web` setup is documented in the new `react-native-web-todo` demo. The demo also illustrates how to handle a few other issues that arise if a project aims to support both mobile and web

I have added a showcase demo under `demos/react-native-web-supabase-todolist` which is based off `react-native-supabase-todolist`. While originally in the same demo, it turned out that too many things that were changed didn't benefit the core `react-native` use case and would just be noise for that path.

# Update
Tested that single tab and multi tab overrides work for both the react-native-web and the normal react (web js) projects


.changeset/two-bats-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/web': patch
```
stevensJourney, line 2: This could be a `minor` change

demos/react-native-supabase-todolist/library/powersync/system.ts
```diff
@@ -18,19 +20,32 @@ export class System {
...
+      });
+    } else {
+      this.powersync = new PowerSyncDatabaseWeb({
+        schema: AppSchema,
+        database: {
+          dbFilename: 'sqlite.db'
+        },
+        flags: {
+          // Only works when `enableMultiTabs` is true.
+          workerPath: '/node_modules/@powersync/web/dist/'
```
stevensJourney, line 45: I think it would be nice to support either passing in the instance of the `Worker`/`SharedWorker` or a URL to that worker. This would make it a bit more flexible for other styles of loading the workers, for example with Vite's worker import syntax. Given that - should this path be for the directory or the URL to individual workers?

demos/react-native-supabase-todolist/library/powersync/system.ts
```diff
@@ -1,7 +1,9 @@
 import '@azure/core-asynciterator-polyfill';
 
 import React from 'react';
-import { PowerSyncDatabase, SyncStreamConnectionMethod } from '@powersync/react-native';
+import { PowerSyncDatabase as PowerSyncDatabaseNative } from '@powersync/react-native';
+import { PowerSyncDatabase as PowerSyncDatabaseWeb } from '@powersync/web/umd';
```
stevensJourney, line 5: We should maybe add a comment here as to why the UMD is required for RN web

stevensJourney (commented):
This looks really nice so far. The docs are quite thorough with good examples.

demos/react-native-supabase-todolist/library/powersync/system.ts
Chriztiaan: Oh this slipped through. It doesn't have to be UMD - it can be. I will rather revert it.

demos/react-native-supabase-todolist/library/powersync/system.ts
Chriztiaan: I follow. To me it makes more sense to add the ability to define all three the workers 🤔

Chriztiaan: Updated the flag definition and readme.

demos/react-native-web-supabase-todolist/CHANGELOG.md
```diff
@@ -0,0 +1,279 @@
+# react-native-supabase-todolist
+
+## 1.3.8
```
stevensJourney, line 3: This changelog should probably start from a fresh history.

demos/react-native-web-supabase-todolist/README.md
```diff
@@ -0,0 +1,289 @@
...
+#### 1.1 Recommended `workers`
+
+With `React Native Web` path to the workers should be configured when instantiating a database factory for `PowerSyncDatabase`, refer to the example [here](./library/powersync/system.ts). It is recommended to configure the workers to point to `/node_modules/@powersync/web/dist/`. If this doesn't work out of the box, try the next section [Copying `dist` and custom `workers` locations](#12-copying-dist-and-custom-workers-locations).
+
+```javascript
+const factory = new WASQLiteOpenFactory({
+  dbFilename: 'sqlite.db',
+  // You can specify paths to the workers
+  sharedSyncWorker: '/node_modules/@powersync/web/dist/worker_SharedSyncImplementation.umd.js',
+  wasqliteDBWorker: '/node_modules/@powersync/web/dist/worker_SharedWASQLiteDB.umd.js'
```
stevensJourney, line 69:
```suggestion
  worker: '/node_modules/@powersync/web/dist/worker_SharedWASQLiteDB.umd.js'
```
I don't think `sharedSyncWorker` should be here. This should just be a SQLite DBAdapter factory.

packages/common/src/client/SQLOpenFactory.ts
```diff
@@ -20,6 +20,22 @@ export interface SQLOpenOptions {
    *    debugMode: process.env.NODE_ENV !== 'production'
```
Chriztiaan, line 20: @stevensJourney 
1. Moved worker config here.

packages/web/src/db/sync/WebStreamingSyncImplementation.ts
```diff
@@ -9,6 +9,12 @@ export interface WebStreamingSyncImplementationOptions extends AbstractStreaming
   flags?: {
```
Chriztiaan, line 9: @stevensJourney 
2. Which would eventually end up here.

demos/react-native-web-supabase-todolist/README.md
stevensJourney: I think we should be able to combine the Shared and dedicated workers into a single JS file which can have it's path specified in a `worker` option which can be of type `worker: string | URL | () => MessagePort`

Returning the message port defers the selection of `new SharedWorker` vs `new Worker` to the function returning the `MessagePort`. If a string path or URL is provided then the adapter should instantiate the correct worker type.

packages/web/src/db/PowerSyncDatabase.ts
```diff
@@ -144,7 +144,7 @@ export class PowerSyncDatabase extends AbstractPowerSyncDatabase {
     const remote = new WebRemote(connector);
 
     const syncOptions: WebStreamingSyncImplementationOptions = {
-      ...this.options,
+      ...(this.options as {}),
```
Chriztiaan, line 147: 3. Because of the typing in `WebStreamingSyncImplementation` and the `this.database` being a filtered down intersection type, it doesn't believe the `database` field is satisfied even though it is.

demos/react-native-web-supabase-todolist/README.md
Chriztiaan: Added the `worker: string | URL | () => MessagePort` part and moved where the db worker and shared worker gets configured.
Working on the combine part now

Chriztiaan: Shared and dedicated db workers have been combined.

demos/react-native-web-supabase-todolist/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+```javascript
+const factory = new WASQLiteOpenFactory({
+  dbFilename: 'sqlite.db',
+  // You can specify a path to the db worker
+  worker: '/node_modules/@powersync/web/dist/worker_WASQLiteDB.umd.js'
+
+  // Or provide factory function to create the worker
+  // worker: () =>
+  //   new SharedWorker(`/node_modules/@powersync/web/dist/worker_WASQLiteDB.umd.js`, {
```
stevensJourney, line 74: We should add an example of conditionally instantiating the `new SharedWorker` vs `new Worker` here given the options. It might be nice to pass the flags/options to the worker generator function.

demos/react-native-web-supabase-todolist/README.md
```diff
@@ -0,0 +1,303 @@
...
+  //     name: `shared-DB-worker-sqlite.db`
```
stevensJourney, line 75: We should document that this name is important. If I remember correctly this name should be unique for the DB filename - although I'm not sure if that restriction is actually required for the DB worker.

demos/react-native-web-supabase-todolist/README.md
```diff
@@ -0,0 +1,303 @@
...
+  schema: AppSchema,
+  database: factory,
+  sync: {
+    // You can specify a path to the sync worker
+    worker: '/node_modules/@powersync/web/dist/worker_SharedSyncImplementation.umd.js'
+
+    // Or provide factory function to create the worker
+    // worker: () =>
+    //   new SharedWorker('/node_modules/@powersync/web/dist/worker_SharedSyncImplementation.umd.js', {
+    //     name: `shared-sync-sqlite.db`
```
stevensJourney, line 89: This name must be unique for the DB filename. This is relevant if multiple clients with different DB's are present.

demos/react-native-web-supabase-todolist/README.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+#### 1.2 Copying `dist` and custom `workers` locations
+
+You can copy the contents of the `dist` directory to somewhere else like `./public`, in which case configure the `workers` accordingly:
+
+```javascript
+const factory = new WASQLiteOpenFactory({
+  dbFilename: 'sqlite.db',
+  worker: '/node_modules/@powersync/web/dist/worker_WASQLiteDB.umd.js'
```
stevensJourney, line 102: I think this URL should be updated to `public/`?

demos/react-native-web-supabase-todolist/README.md
```diff
@@ -0,0 +1,303 @@
...
+});
+
+this.powersync = new PowerSyncDatabaseWeb({
+  schema: AppSchema,
+  database: factory,
+  sync: {
+    worker: '/public/worker_SharedSyncImplementation.umd.js'
```
stevensJourney, line 109: Is it possible to name these outputs to look prettier? This current name might be okay, just something seems weird to me about having to type this.

demos/react-native-web-supabase-todolist/library/powersync/system.ts
```diff
@@ -0,0 +1,96 @@
...
+        //   new SharedWorker(`/node_modules/@powersync/web/dist/worker_WASQLiteDB.umd.js`, {
+        //     name: `shared-DB-worker-sqlite.db`
+        //   }).port
+      });
+      this.powersync = new PowerSyncDatabaseWeb({
+        schema: AppSchema,
+        database: factory,
+        sync: {
+          // You can specify a path to the sync worker
+          worker: '/node_modules/@powersync/web/dist/worker_SharedSyncImplementation.umd.js'
```
stevensJourney, line 55: Is it possible to add exports for the worker files in the `@powersync/web` package and then use `require.resolve()` for this?

For example:
```js
worker: require.resolve('@powersync/web/worker/sync')
```

Chriztiaan: - [ ] Investigate
- [ ] Apply

demos/react-native-web-supabase-todolist/README.md
Chriztiaan: - [ ] Pass along options
- [ ] Conditionally construct in example based on mutlitab flag

demos/react-native-web-supabase-todolist/library/powersync/system.ts
Chriztiaan: As discussed this doesn't work for metro out of the box, I believe the current iteration of the usage is good enough.

demos/react-native-web-supabase-todolist/library/powersync/system.ts
```diff
@@ -0,0 +1,96 @@
...
+        schema: AppSchema,
+        database: {
+          dbFilename: 'sqlite.db'
+        }
+      });
+    } else {
+      const factory = new WASQLiteOpenFactory({
+        dbFilename: 'sqlite.db',
+        // You can specify a path to the db worker
+        worker: '/node_modules/@powersync/web/dist/worker_WASQLiteDB.umd.js'
```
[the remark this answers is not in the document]
Chriztiaan: node_modules didn't work, moving things to the public dir with a sub dir does.

Chriztiaan:
@stevensJourney @kobiebotha I think this PR is ready for scrutiny once again.

demos/react-native-web-supabase-todolist/README.md
```diff
@@ -0,0 +1,288 @@
...
+bucket_definitions:
+  user_lists:
+    # Separate bucket per todo list
+    parameters: select id as list_id from lists where owner_id = request.user_id()
+    data:
+      - select * from lists where id = bucket.list_id
+      - select * from todos where list_id = bucket.list_id
+```
+
+## Configure The App
```
benitav, line 36: What do you think about using the `.env.local.template` approach that we seem to have adopted in our later demos, e.g. https://github.com/powersync-ja/powersync-js/blob/main/demos/react-supabase-todolist/README.md? Seems a bit cleaner and probably also safer.

Chriztiaan: Sorted

stevensJourney (dismissed):

benitav (approved):

PR #282, opened 2024-08-28
chore(common): merge Table and TableV2

DominicGBauer:
## Description
Merge `Table` and `TableV2` to avoid needing to make a breaking change to attachments package.

## Work Done
* Used overload to differentiate between table v1 and v2
* Removed `TableV2` tests in favour of `Table` tests
* Moved `Column.ts` under `schema` as this makes more sense
* Kept `TableV2` but this should be removed in future major release, we should also remove all Table V1 related code in that as well

## Shoutout
To @stevensJourney for doing the majority of this.


packages/common/src/db/schema/Schema.ts
```diff
@@ -28,20 +27,20 @@ export class Schema<S extends SchemaType = SchemaType> {
...
 
   toJSON() {
     return {
-      tables: (this.tables as ClassicTable[]).map((t) => t.toJSON())
+      tables: this.tables.map((t) => t.toJSON())
     };
   }
 
   private convertToClassicTables(props: S) {
     return Object.entries(props).map(([name, table]) => {
```
stevensJourney, line 42: We should make a comment here for why this is still required. Which is due to the `name` field not being present in the "V2" syntax.

packages/common/src/db/schema/Table.ts
```diff
@@ -35,7 +58,7 @@ export class Table {
     return new Table({ ...options, localOnly: false, insertOnly: true });
   }
 
-  static createTable(name: string, table: TableV2) {
+  static createTable(name: string, table: Table) {
     return new Table({
       name,
       columns: Object.entries(table.columns).map(([name, col]) => new Column({ name, type: col.type })),
```
stevensJourney, line 64: This might be incorrect since `table.columns` is an Array.

packages/common/src/db/schema/Table.ts
```diff
@@ -98,41 +181,42 @@ export class Table {
 
   validate() {
     if (InvalidSQLCharacters.test(this.name)) {
-      throw new Error(`Invalid characters in table name: ${this.name}`);
+      throw new Error(`Invalid characters in table`);
```
stevensJourney, line 184: Why change this error?

packages/common/src/db/schema/Table.ts
```diff
@@ -46,8 +69,58 @@ export class Table {
     });
   }
 
-  constructor(options: TableOptions) {
-    this.options = { ...DEFAULT_TABLE_OPTIONS, ...options };
+  constructor(columns: Columns, options?: TableV2Options);
```
stevensJourney, line 72: The autocomplete is a bit bad for this now. The TypeScript language server can only tell that V1 syntax is being used after a `columns` field which is an array is supplied. Otherwise it cannot tell if an empty options object is a Column map or TableOptions object.

Could we add examples and document the different constructors better in comments to help with this?

packages/common/src/db/schema/Table.ts
DominicGBauer: There isn't always a name when validation is run.

packages/common/src/db/schema/Table.ts
```diff
@@ -35,19 +58,127 @@ export class Table {
...
+   * @param {Columns | TableOptions} optionsOrColumns - Either a Columns object (for V2 syntax) or a TableOptions object (for V1 syntax)
+   * @param {TableV2Options} [v2Options] - Optional configuration options for V2 syntax
+   *
+   * @example
+   * <caption>New constructor example</caption>
+   * ```javascript
+   * const table = new Table(
+   *   {
+   *     name: { type: ColumnType.TEXT },
+   *     age: { type: ColumnType.INTEGER }
```
rkistner, line 96: I'd recommend the shorter `name: column.text, age: column.integer` here.

packages/common/src/db/schema/Table.ts
rkistner: Since we're deprecating the V1 syntax, I'm happy with the auto-complete not being great for that.

packages/common/src/db/schema/Table.ts
```diff
@@ -59,13 +190,23 @@ export class Table {
   }
 
   get viewName() {
-    return this.viewNameOverride ?? this.name;
+    return this.viewNameOverride ? this.viewNameOverride : this.name;
```
rkistner, line 193: Why this change?

rkistner (commented):
I quite like this change - what more do we need before getting this out?

packages/common/src/db/schema/Table.ts
stevensJourney: I still think it would be nice to document the constructor(s) with a JS-docs example - we could also mention the deprecation there.

packages/common/src/db/schema/Table.ts
DominicGBauer: After looking into this I have reverted this back and solved the error when the conversion to classic table happens. 

What was happening is because of a circular change happening:

Given these
```
  get viewNameOverride() {
    return this.options.viewName;
  }

  get viewName() {
    return this.viewNameOverride ?? this.name;
  }
```
and
```
  private convertToClassicTables(props: S) {
    return Object.entries(props).map(([name, table]) => {
      const convertedTable = new Table({
        name,
        columns: table.columns,
        indexes: table.indexes,
        localOnly: table.localOnly,
        insertOnly: table.insertOnly,
        viewName: table.viewName
      });
      return convertedTable;
    });
  }
```

1. Initially if viewName is not declared in TableV2 syntax then it is `undefined` so `viewNameOverride` is `undefined` and the `viewName` getter became `""` 
2. When that table is converted to the classic table `viewNameOverride` is now `""` because `viewName` became `""` so when using `this.viewNameOverride ?? this.name` it would evaluate to `""` and not use the `name`.

I've changed it now to avoid this:

```
  private convertToClassicTables(props: S) {
    return Object.entries(props).map(([name, table]) => {
      const convertedTable = new Table({
        name,
        columns: table.columns,
        indexes: table.indexes,
        localOnly: table.localOnly,
        insertOnly: table.insertOnly,
        viewName: table.viewNameOverride || name
      });
      return convertedTable;
    });
  }
  ```
  
However it becomes very confusing having `name`, `viewName` and `viewNamOverride` and we should look at removing `name` or `viewName` in the next major release to simplify this.

File: packages/common/src/db/schema/Table.ts (hunk 4 above)
Comment (DominicGBauer) [reply packages/common/src/db/schema/Table.ts:115]: I've added JSDoc comments for these

File: packages/common/src/db/schema/Table.ts (hunk 3 above)
Comment (DominicGBauer) [reply packages/common/src/db/schema/Table.ts]: Nvm reverted this

File: demos/example-vite/src/index.js [hunk 7]
```diff
@@ -19,9 +19,9 @@ class DummyConnector {
   async uploadData(database) {}
 }
 
-export const AppSchema = new Schema([
-  new Table({ name: 'customers', columns: [new Column({ name: 'name', type: ColumnType.TEXT })] })
-]);
+const customers = new Table({ name: column.text })
+
+export const AppSchema = new Schema(customers);
```
rkistner, line 24:
```suggestion
export const AppSchema = new Schema({ customers });
```

demos/example-webpack/src/index.js
```diff
@@ -19,9 +19,9 @@ class DummyConnector {
   async uploadData(database) {}
 }
 
-export const AppSchema = new Schema([
-  new Table({ name: 'customers', columns: [new Column({ name: 'name', type: ColumnType.TEXT })] })
-]);
+const customers = new Table({ name: column.text })
+
+export const AppSchema = new Schema(customers);
```
rkistner, line 24:
```suggestion
export const AppSchema = new Schema({ customers });
```

rkistner (approved):

PR #283, opened 2024-08-28
chore: added isLoading to docs

stevensJourney:
# Overview

This adds a basic example of using `isLoading` and `isFetching` to the `@powersync/react` README.

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -102,15 +102,15 @@ export const TodoListsDisplayDemo = () => {
         }}>
         Loading todo lists...
       </div>
-      <div
+      <ul
         style={{
           opacity: isLoading ? 0 : 1,
           transition: 'opacity 1s ease-in-out'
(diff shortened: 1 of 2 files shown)


packages/react/README.md
```diff
@@ -81,3 +81,38 @@ export const TodoListDisplay = () => {
     </View>
 }
 ```
+
+#### Query Loading
+
+The result from `useQuery` contains information about the data loading and fetching states. This can be used to show loading spinners or conditional widgets.
```
DominicGBauer, line 87: I think add something like "`useQuery` has the `isLoading` and `isFetching` properties which can be used to see what the current fetch state is ...`. I think this will make it more clear?

stevensJourney: That is better. I updated it now with some minor alterations.

DominicGBauer (approved):

PR #286, opened 2024-09-02
Fixed fast-base64-decode dependency issue in react-native-package.

Chriztiaan:
As part of the prebundling work, `fast-base64-decode` was thought to be bundled in the `crypto.js` vendor file that gets injected in the rollup config. Turns out it is not the case, and sometimes this caused projects to fail startup if the dependency wasn't installed.

This fix takes the small dependency and manually adds its code to the vendor file, instead of trying to figure out how to bundle a dependency that gets injected. Both the crypto and fast-base64-decode libraries are from the same author so this felt okayish.

I have tested this fix on both the monorepo and a standalone React native project.

Dev release available at:
```
"@powersync/react": "0.0.0-dev-20240903063630",
"@powersync/common": "0.0.0-dev-20240903063630",
"@powersync/react-native": "0.0.0-dev-20240903063630",
"@powersync/web": "0.0.0-dev-20240903063630"
```


stevensJourney (approved):
This fixed the issue of `fast-base64-decode could not be resolved` for me in my copy of the Ignite PowerSync template app.

stevensJourney (approved):

PR #289, opened 2024-09-05
powersync-sqlite-core v0.2.1

rkistner:
Update with the changes in https://github.com/powersync-ja/powersync-sqlite-core/pull/21.

When using a self-hosted instance, this requires v0.5.0 or later of the image, to include [these changes](https://github.com/powersync-ja/powersync-service/pull/64). The changes are already rolled out on all hosted instances.

## Changes

1. Add (X-)User-Agent header. Examples: `powersync-js/1.16.2 powersync-web/1.6.0 Chrome/128 windows` / `powersync-js/1.16.2 powersync-react-native/1.11.0 react-native/0.78 android/34`
2. Add `client_id` parameter with an unique uuid per client.
3. Persist `lastSyncedAt`.
4. Emit update notifications on `disconnectAndClear()`.
5. Validate that the powersync-sqlite-core version number is in a compatible range of `^0.2.0`.
6. Always cast `target_op` to ensure it's an integer.
7. Sync optimizations: https://github.com/powersync-ja/powersync-sqlite-core/pull/19

See similar dart SDK changes here: https://github.com/powersync-ja/powersync.dart/pull/148


packages/web/src/db/sync/WebRemote.ts
```diff
@@ -10,6 +10,9 @@ import {
   RemoteConnector
 } from '@powersync/common';
 
+import { version as POWERSYNC_WEB_VERSION } from '../../../package.json';
```
stevensJourney, line 13: Will the work by default with most bundlers like Vite and Webpack?

stevensJourney (dismissed):
I'm very happy with the changes. These are cool features.

packages/web/src/db/sync/WebRemote.ts
rkistner: It works with Vite (tested the diagnostics app). What are all the ones we need to test?

stevensJourney: I think we just have Vite and Webpack examples for now.

rkistner: It works with vite (diagnostics app) and webpack (example-webpack).

angular-supabase-todolist works, but gave this warning:
> ../../packages/web/lib/src/db/sync/WebRemote.js:26:53-74 - Warning: Should not import the named export 'version' (imported as 'POWERSYNC_WEB_VERSION') from default-exporting module (only default export is available soon)

So for now I removed the version from the web and react-native packages, and just kept it in common where it's bundled. Can fined a proper solution catering for all environments later.

stevensJourney (approved):

Issue #292, opened 2024-09-06
Support for Kysely plugins


bviebahn:
Using Kysely plugins doesn't seem to be properly supported by the react package. I am using the `ParseJSONResultsPlugin`, and it works when I call `execute` on the query, but not when I pass the query to the useQuery hook.

I guess this is a bug rather than a feature request, because:
- The types of `wrapPowerSyncWithKysely` suggest kysely can be configured as usual using the second config parameter. (On another note: I think the config parameter should be changed to `Omit<KyselyConfig, 'dialect'>`, since `dialect` is set by `wrapPowerSyncWithKysely` and I'm not sure it ever makes sense to override it. At least it should be optional)
- Kysely seems to assume specific plugins to be used. For example `jsonArrayFrom` and `jsonObjectFrom` assume to be used with `ParseJSONResultsPlugin`, so the types will be wrong if the plugin doesn't work.

stevensJourney:
This is probably due to us calling `.compile` on the supplied Kysely query to get the SQL and query parameters. We don't explicitly call `.execute` on the Kysely query. I'm assuming the `.execute` call should allow the plugins to map column data. We can investigating supporting `.execute`

guillempuche:
This means I can't use https://kysely-org.github.io/kysely-apidoc/classes/CamelCasePlugin.html for now, right?

stevensJourney:
> This means I can't use https://kysely-org.github.io/kysely-apidoc/classes/CamelCasePlugin.html for now, right?

Most likely. This will be fixed soon.

guillempuche:
I love the team's speed for improvements!

guillempuche:
Reminder: this will help to convert [Temporal datetimes](https://github.com/fullcalendar/temporal-polyfill) via a plugin. More here https://discord.com/channels/1138230179878154300/1282584951656550473/1282671991005909072

stevensJourney:
Closing this issue as plugins should be supported from `@powersync/kysely-driver@0.4.0`

PR #295, opened 2024-09-09
[Fix] Custom Query Executor Hooks

stevensJourney:
# Overview

Packages such as the Kysely driver provide a `.execute()` method on query objects. This execute method could perform pluggable modifications based on installed plugins. 

This PR allows the `useQuery` hooks and composibles from `@powersync/react` and `@powersync/vue` to use the `execute` method from compatible queries. 

Related issue: https://github.com/powersync-ja/powersync-js/issues/292

The `dialect` option in the  `wrapPowerSyncWithKysely` has been made optional since the method provides a PowerSync dialect by default.

## Testing

This was verified with unit tests and by implementing a simple Kysely plugin which overwrites the query result.

```TypeScript
const k = wrapPowerSyncWithKysely<Database>(db, {
  plugins: [
    {
      transformQuery: (p) => p.node as any,
      transformResult: async (args) => {
        return {
          rows: [{ hello: 's' }]
        };
      }
    }
  ]
});

k.selectFrom('lists')
  .selectAll()
  .execute()
  .then((r) => console.log(`Got test`, r));
```
image

### DIFF
--- a/.changeset/orange-points-speak.md
+++ b/.changeset/orange-points-speak.md
@@ -3,4 +3,4 @@
 '@powersync/vue': patch
 ---
 
-React and Vue helpers should execute queries from capatible query executor methods. This should allow Kysely queries with plugins to function correctly.
+React and Vue helpers should execute queries from compatible query executor methods. This should allow Kysely queries with plugins to function correctly.
--- a/.changeset/orange-points-speak.md
+++ b/.changeset/orange-points-speak.md
@@ -1,0 +1,6 @@
+---
+'@powersync/react': patch
+'@powersync/vue': patch
+---
+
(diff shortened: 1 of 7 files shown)


.changeset/orange-points-speak.md
```diff
@@ -0,0 +1,6 @@
+---
+'@powersync/react': patch
+'@powersync/vue': patch
+---
+
+React and Vue helpers should execute queries from capatible query executor methods. This should allow Kysely queries with plugins to function correctly.
```
rkistner, line 6: Should this be `compatible`?

rkistner (approved):

.changeset/orange-points-speak.md
stevensJourney: Jip. Thanks for spotting.

Chriztiaan (approved):

PR #296, opened 2024-09-09
[Fix] SSR Mode Detection

stevensJourney:
# Overview

This fixes an issue where manually specifying `ssrMode: true` in the `PowerSyncDatabase` constructor `flags` could cause unexpected syncing behaviour to occur in the browser client.

Previously the `WASQLiteOpenFactory` did not account for the `ssrMode` flag. This would create a `DBAdapter` based on the detected environment. Specifying `ssrMode: true` would result in a functional `WASQLiteDBAdapter` being created (on the client). 

Previously the `SSRStreamingSyncImplementation` extended `AbstractStreamingSyncImplementation` which contains logic for processing sync commands from the server. Sync commands would be a `no-op` if the suppled `DBAdapter` was a placeholder SSR DBAdapter, but as described above this was not the case if `ssrMode: true` was specified.

The locks in `SSRStreamingSyncImplementation` are not exclusive, which could result in multiple clients processing sync commands into a single SQLite DB. This PR removes the ability for `SSRStreamingSyncImplementation` to perform any actual syncing. This implementation is just meant to be a placeholder for cases where the PowerSyncDatabase client is accidentally used or exposed on the server.

### DIFF
--- a/.changeset/four-birds-raise.md
+++ b/.changeset/four-birds-raise.md
@@ -1,0 +1,6 @@
+---
+'@powersync/common': minor
+'@powersync/web': minor
+---
+
+Fixed SSR Mode detection for DB adapters. Removed the potential for SSR Web Streamining implementations from to perform syncing operations.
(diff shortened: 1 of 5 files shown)


rkistner (approved):

guillempuche:
I don't find this mode in the docs. What is it about?

stevensJourney:
> I don't find this mode in the docs. What is it about?

We have a TODO for adding SSR docs. 

In short: Our web SDK requires web APIs. WebAPIs are not available when the server might be rending components for SSR. Using web APIs on the server typically throws exceptions for missing methods. Our web SDK is not meant to be used on the server-side, but we provide these helpers which will avoid actually performing and DB or sync operations if a developer accidentally ends up attempting to use the SDK on the server-side.

PR #301, opened 2024-09-12
Remove lodash dependency

rkistner:
`lodash/throttle` was the last remaining function used, and it's simple enough to provide a custom implementation.

This is to reduce issues related to ES modules going forward, and may slightly reduce the bundle size.


stevensJourney (approved):
Looks good to me.

PR #309, opened 2024-09-17
[Demo] Optional Sync Supabase Todo App 

Chriztiaan:
This adds a demo project that aims to demonstrate how an app could be built starting from a local-only state and later being "upgraded" to a version that has sync capabilities. It started off as a copy of the existing todo app that was stripped of some of the features. It works off the existing react-supabase-todo demo and the optional-sync demo added to the Dart monorepo [here](https://github.com/powersync-ja/powersync.dart/pull/138).

The readme has a basic description of how everything works, with diagrams to show how the app switches from the local-only mode to sync mode.

Added tests to the [web package](https://github.com/powersync-ja/powersync-js/pull/309/files#diff-afb1b89e2eb253e26cf5b602322c6413b947de50801560ed3fb9e584688d8e68) to confirm that `updateSchema` works. A refresh schema implementation doesn't seem to be required based on looking at the tests and updating the schema in the demo.


demos/react-supabase-todolist-optional-sync/src/library/powersync/AppSchema.ts
```diff
@@ -0,0 +1,107 @@
+import { AbstractPowerSyncDatabase, column, ColumnsType, Schema, Table, TableV2Options } from '@powersync/web';
+import { setSyncEnabled } from './SyncMode';
+
+export const LISTS_TABLE = 'lists';
+export const TODOS_TABLE = 'todos';
+
+const todosDef = {
```
Chriztiaan, line 7: I struggled to get the `options` out of an already constructed table. This inline object instantiation works but there could be a cleaner way.

demos/react-supabase-todolist-optional-sync/src/library/powersync/AppSchema.ts
```diff
@@ -0,0 +1,107 @@
+import { AbstractPowerSyncDatabase, column, ColumnsType, Schema, Table, TableV2Options } from '@powersync/web';
+import { setSyncEnabled } from './SyncMode';
```
benitav, line 2: Would be good to add an overview comment like we have in dart https://github.com/powersync-ja/powersync.dart/blob/main/demos/supabase-todolist-optional-sync/lib/models/schema.dart#L4

demos/react-supabase-todolist-optional-sync/src/library/powersync/AppSchema.ts
```diff
@@ -0,0 +1,107 @@
...
+      viewName: localName(listsDef.name)
+    })
+  });
+}
+
+export async function switchToSyncedSchema(db: AbstractPowerSyncDatabase, userId: string) {
+  await db.updateSchema(makeSchema(true));
+
+  // needed to ensure that watches/queries are aware of the updated schema
+  // await db.refreshSchema();
```
benitav, line 75: Should we just remove this entriely?

Chriztiaan: - [ ] Remove

demos/react-supabase-todolist-optional-sync/src/library/powersync/AppSchema.ts
Chriztiaan: Ported comment.

stevensJourney (dismissed):

stevensJourney (approved):

PR #313, opened 2024-09-20
Fix auto docker image release for diagnostics app

rkistner:
The workflow was triggered on tag push, but then the job is ignored since `github.ref` isn't `main`. This fixes it.


rkistner:
This does not actually fix the issue - releasing manually for now.

Issue #323, opened 2024-09-29
[Feature/React] Support for dynamic dependencies in `useQuery`


guillempuche:
Currently, `useQuery` in PowerSync throws the following warning when the dependencies change dynamically:
```
Warning: The final argument passed to useMemo changed size between renders. The order and size of this array must remain constant. Previous: [] Incoming: [01923dca-8707-744a-a214-d69e099b9a83]
```
This issue arises when `useQuery`'s dependency changes from `null` to a valid value, which is a common pattern in React when dealing with conditional data fetching based on asynchronous or optional data sources.

#### **React Hooks Example:**
React allows handling such scenarios using `useEffect` and `useState`:
```tsx
import React, { useEffect, useState } from 'react';

// Simulate an asynchronous fetch function
const fetchData = async (creatorId: string) => {
  return `Data for creatorId: ${creatorId}`;
};

const ExampleComponent = () => {
  const [creator, setCreator] = useState<{ id: string } | null>(null);
  const [data, setData] = useState<string | null>(null);

  // Simulate setting the creator asynchronously
  useEffect(() => {
    setTimeout(() => {
      setCreator({ id: '01923dca-8707-744a-a214-d69e099b9a83' });
    }, 1000);
  }, []);

  useEffect(() => {
    if (creator) {
      fetchData(creator.id).then(setData);
    }
  }, [creator]);

  return (
    <div>
      <p>{data ? data : 'Loading...'}</p>
    </div>
  );
};
```

In this example, React's `useEffect` handles the asynchronous change of `creator` correctly without any warnings when `creator` changes from `null` to an object.

#### **Problem:**
In my case, I use `useQuery` with `creator.value.id` obtained from an authentication hook:
```tsx
const { creator } = useAuth();

const { data: platforms, isLoading: isLoadingPlatforms } = useQuery(
  Option.isSome(creator)
    ? RuntimeClient.runSync(RepoPlatform.getList({ creatorId: creator.value.id }))
    : undefined, // or ' '
);
```
When `creator` is initially `null` and later resolves to a valid object, `useQuery` throws the warning because its dependency array changes size from `[]` to `[creator.value.id]`.

PowerSync's current implementation does not handle such dynamic dependencies correctly, limiting its flexibility in handling real-world scenarios where dependencies can resolve asynchronously or conditionally.

#### **Proposed Solution:**
[...]

PR #325, opened 2024-10-01
[Fix] `packages/react` issue with `useQuery` not supporting dynamic query parameters

Chriztiaan:
Original issue reported at https://github.com/powersync-ja/powersync-js/issues/323.
Dynamic query dependencies would warn with the following if the query parameter array changed in size:
 `Warning: The final argument passed to useMemo changed size between renders. The order and size of this array must remain constant. Previous: [] Incoming: [xxxxxxxxxxxxx]
`

Tested with the following to verify that it works now:
```
const [params, setParams] = React.useState<string[]>([]);
const { data } = useQuery(`select * from lists where id=?`, params);

onclick:: setParams(["some-uuid"])
```

The change from `React.useMemo(() => {}, [...array]);` to `React.useMemo(() => {},[JSON.stringify(array)])` should not introduce any issues, as valid query parameters are expected to be serialisable.

### DIFF
--- a/.changeset/few-jeans-prove.md
+++ b/.changeset/few-jeans-prove.md
@@ -1,0 +1,5 @@
+---
+'@powersync/react': patch
+---
+
+Fixed issue with `useQuery` not supporting dynamic query parameters.
--- a/packages/react/src/hooks/useQuery.ts
+++ b/packages/react/src/hooks/useQuery.ts
@@ -62,7 +62,7 @@ export const useQuery = <T = any>(
   const [isFetching, setIsFetching] = React.useState(true);
   const [tables, setTables] = React.useState([]);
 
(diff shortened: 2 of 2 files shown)


DominicGBauer (approved):

DominicGBauer:
We may want to look at something like this in the future https://github.com/sandiiarov/use-deep-compare#readme.

guillempuche:
Very cool Christian!

fooware:
I just got hit with this today in our process of porting over our codebase from Electric. Spent quite some time trying to figure out which `useMemo` it was, until I found this. Thanks for fixing it @Chriztiaan!

Any idea when this could be merged and released? Is there any canary / nightly one could test after it is merged?

stevensJourney (approved):

Chriztiaan:
@fooware, I'll release this soonest and ping you.

PR #326, opened 2024-10-02
docs(react-native): add note about metro config

DominicGBauer:
## Description
Added a note in the docs about `metro.config.js` changes that need to be made to get the package to work with non-Expo React Native projects. This should hopefully avoid these issues for users in the future https://discord.com/channels/1138230179878154300/1290949501049110623 and https://discord.com/channels/1138230179878154300/1194710422960472175/1290907741568765964

### DIFF
--- a/packages/react-native/README.md
+++ b/packages/react-native/README.md
@@ -73,7 +73,7 @@ module.exports = function (api) {
 
 ## Metro config (optional)
 
-When using a bare react native app without a framework like Expo, the `@powersync/react-native` package does not work well with inline requires.
+When using a bare React Native app without a framework like Expo, the `@powersync/react-native` package does not work well with inline requires.
 
 If you see the following error message
 
--- a/packages/react-native/README.md
+++ b/packages/react-native/README.md
@@ -71,6 +71,26 @@ module.exports = function (api) {
 };
 ```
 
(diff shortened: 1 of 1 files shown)


packages/react-native/README.md
```diff
@@ -71,6 +71,32 @@ module.exports = function (api) {
 };
 ```
 
+# Metro config
+
+The `@powersync/react-native` has issues with inline requires. If you are not using Expo then you may see this error
```
mugikhan, line 76: We should make this message more clear to indicate when it is needed. Something along the lines of

# Metro config (optional)

When using a bare react native app without a framework like Expo, the `@powersync/react-native` package does not work well with inline requires.

packages/react-native/README.md
```diff
@@ -71,15 +71,18 @@ module.exports = function (api) {
 };
 ```
 
-# Metro config
+## Metro config (optional)
+
+When using a bare react native app without a framework like Expo, the `@powersync/react-native` package does not work well with inline requires.
```
benitav, line 76:
```suggestion
When using a bare React Native app without a framework like Expo, the `@powersync/react-native` package does not work well with inline requires.
```

benitav (approved):

PR #328, opened 2024-10-03
feat(diagnostics): update autogenerated schema and description

LucDeCaf:
Updates the schema generated by the diagnostics app to use the current Schema and Table syntax.
Also updates the comment above the schema to be more descriptive.

Below is an example of a before/after.

Before:
```js
// This displays the inferred schema currently used by the diagnostics app.
// This is based on downloaded data, rather than the source database.
// If a table is empty, it will not display here.
// Tables and columns are only added here. Nothing is removed until the database is cleared.
new Schema([
  new Table({
    name: 'profiles',
    columns: [
      new Column({ name: 'household_id', type: ColumnType.TEXT })
    ]
  }),
  new Table({
    name: 'grocery_items',
    columns: [
      new Column({ name: 'household_id', type: ColumnType.TEXT }),
      new Column({ name: 'created_at', type: ColumnType.TEXT }),
      new Column({ name: 'name', type: ColumnType.TEXT }),
      new Column({ name: 'comment', type: ColumnType.TEXT })
    ]
  }),
  new Table({
    name: 'groups',
    columns: [
      new Column({ name: 'household_id', type: ColumnType.TEXT }),
      new Column({ name: 'created_at', type: ColumnType.TEXT }),
      new Column({ name: 'name', type: ColumnType.TEXT }),
      new Column({ name: 'index', type: ColumnType.INTEGER }),
      new Column({ name: 'is_aisle', type: ColumnType.INTEGER })
    ]
  }),
  new Table({
    name: 'groups_grocery_items',
    columns: [
      new Column({ name: 'household_id', type: ColumnType.TEXT }),
      new Column({ name: 'group_id', type: ColumnType.TEXT }),
      new Column({ name: 'grocery_item_id', type: ColumnType.TEXT })
    ]
  })
])
```

After:
```js
/**
 * This is the inferred schema of the data received by the diagnostics app.
 * Because this schema is generated on-the-fly based on the data received by the app, it can
 * be incomplete and should NOT be relied upon as a source of truth for your app schema.
 * If a table is empty, it will not be shown here.
 * Tables and columns are only added here. Nothing is removed until the database is cleared.
 */
export const profiles = new Table(
  {
    // id column (text) is automatically included
    household_id: column.text
  },
  { indexes: {} }
);

export const grocery_items = new Table(
  {
    // id column (text) is automatically included
    household_id: column.text,
    created_at: column.text,
    name: column.text,
    comment: column.text
  },
  { indexes: {} }
);

export const groups = new Table(
  {
    // id column (text) is automatically included
    household_id: column.text,
    created_at: column.text,
    name: column.text,
    index: column.integer,
    is_aisle: column.integer
  },
  { indexes: {} }
);

export const groups_grocery_items = new Table(
  {
    // id column (text) is automatically included
    household_id: column.text,
    group_id: column.text,
    grocery_item_id: column.text
  },
  { indexes: {} }
);

export const AppSchema = new Schema({
  profiles,
  grocery_items,
  groups,
  groups_grocery_items
});

export type Database = (typeof AppSchema)['types'];
```


.changeset/smart-jobs-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/diagnostics-app': minor
+---
+
+Updated schema generated by diagnostics app + increased descriptiveness of the preceding comment"
```
benitav, line 5: Nit: can simplify for the changeset and leave out the comment update since that's quite a trivial update

DominicGBauer:
LGTM, 

We will want to adapt the generators from https://github.com/powersync-ja/powersync-service/blob/d8e8c4d16e649ca45ae1a95980b66f7b453835c3/packages/sync-rules/src/SchemaGenerator.ts#L5 to allow for the case of not using sync rules to generate schema, as is the case in here. This will avoid us duplicating the schema generation in future and would allow us to import the generators from `@powersync/service-sync-rules`.

DominicGBauer (approved):

tools/diagnostics-app/src/app/views/schema.tsx
```diff
@@ -4,10 +4,13 @@ import { Box, Grid, styled } from '@mui/material';
...
-  const docs = `// This displays the inferred schema currently used by the diagnostics app.
-// This is based on downloaded data, rather than the source database.
-// If a table is empty, it will not display here.
-// Tables and columns are only added here. Nothing is removed until the database is cleared.`;
+  const docs = `/**
+ * This is the inferred schema of the data received by the diagnostics app.
+ * Because this schema is generated on-the-fly based on the data received by the app, it can
+ * be incomplete and should NOT be relied upon as a source of truth for your app schema.
+ * If a table is empty, it will not be shown here.
+ * Tables and columns are only added here. Nothing is removed until the database is cleared.
```
Dieterbe, line 12: .. but not columns that only have null values, right? might be good to clarify that, as that's a subtle issue.

LucDeCaf: That's true, good catch 👍 Added a line that explains that.

benitav (approved):

PR #331, opened 2024-10-07
fix: Ensuring that `useQuery`'s `isFetching` becomes true after query changes

Chriztiaan:
An issue was found where if the query or parameters change for `useQuery`, there would be single render cycle where `isFetching` was `false` before it became `true`.

For example:
```
1. isFetching: false // query changed but the variable is still false
2. isFetching: true // state has been updated
3. isFetching: false // query has finished re-executing
```

This fix ensures that `isFetching` is correctly set to `true` as soon as the hook executes due to a query change.
```
1. isFetching: true // query change causes hook to execute, compare current query to new query - 
track with internal `shouldFetch` variable.
2. isFetching: true // state has been updated
3. isFetching: false // query has finished re-executing
```


fooware:
In our case, we had multiple renders with `isFetching` being incorrectly false in combination with the old stale data, probably due to us having multiple `useQuery` hooks firing and re-rendering things.

I have tested @Chriztiaan's changes locally, and the latest version in this PR resolves the issue for us.

packages/react/src/hooks/useQuery.ts
```diff
@@ -139,5 +152,5 @@ export const useQuery = <T = any>(
     };
   }, [powerSync, sqlStatement, memoizedParams, memoizedOptions, tables]);
 
-  return { isLoading, isFetching, data, error, refresh: fetchData };
```
DominicGBauer, line 142: Can we add a test for this?

DominicGBauer: Sorry, for the change*

Chriztiaan: @DominicGBauer, I had to fix a few of the React `useQuery` tests which weren't using `await` with `waitFor`. After doing that I took a stab at testing the new `isFetching` value but there's an issue with how `setXXX` works for React.useState when used inside the test framework. I am going to park this for now, but I do have the test saved when I get a chance to look at this gain.

DominicGBauer (approved):

Issue #322, opened 2024-09-29
Error [Not at a consistent checkpoint - cannot update local db] when running react-native-supabase-todolist


nthtrung09it:
I ran the source **react-native-supabase-todolist**, but saw the error log:

![CleanShot 2024-09-29 at 10 25 12@2x](https://github.com/user-attachments/assets/3aeee9eb-ec41-4e1c-8028-bad4727e35e6)

> DEBUG  [PowerSyncStream] Checkpoint complete {"buckets": [{"bucket": "user_lists[\"d727e9f8-3fa9-4cb3-b14c-715fe8a74da2\"]", "checksum": -1366502408, "count": 2}, {"bucket": "user_lists[\"c021e243-f121-4341-8749-5c9b43c86391\"]", "checksum": 572786632, "count": 6}, {"bucket": "user_lists[\"fab167a0-09db-4b0a-b303-cdc4aea2ac9d\"]", "checksum": 977374437, "count": 1}, {"bucket": "user_lists[\"05d32f2a-39d0-4a79-9c9b-1f0dccde481e\"]", "checksum": 1909264827, "count": 1}], "last_op_id": "270"}
> DEBUG  [SqliteBucketStorage] validateChecksums result item {"result": "{\"valid\":true,\"failed_buckets\":[]}"}
> DEBUG  [SqliteBucketStorage] Not at a consistent checkpoint - cannot update local db
> DEBUG  [PowerSyncStream] Caught exception when uploading. Upload will retry after a delay. Exception: Received 404 -  when getting from /write-checkpoint2.json?client_id=bda2e5be-87c9-4b63-a690-61fecb280c93: {"message":"Route GET://write-checkpoint2.json?client_id=bda2e5be-87c9-4b63-a690-61fecb280c93 not found","error":"Not Found","statusCode":404}}

When I did a fresh install the app, logined and then updated data on supabase tables, changes **synced** to ios app.
But when I started to change something in my ios app (create new list), changes from ios app synced to my supabase tables, but changes from supabase tables did not sync to my ios app.

The web project (react-supabase-todolist) does not encounter that issue.

Here is the video: [powersync_supabase_ios_sync_error](https://www.icloud.com/iclouddrive/0bfhjfabK-XETMRr6TC5tEIRA#powersync%5Fsupabase%5Fios%5Fsync%5Ferror)

stevensJourney:
[...]

PR #333, opened 2024-10-07
Handle typo in powersync_url env variable

HeinrichvonStein:
When a trailing forward slash "/' is present in the POWERSYNC_URL environment variable, it causes the `GET` request  that writes a checkpoint to fail.

This fix addresses issue https://github.com/powersync-ja/powersync-js/issues/322 by throwing an error that lets a user know that they have made a typo. The error message appears as follows:
  - Logs:
    Screenshot 2024-10-23 at 12 40 04

  - Error notification:
  Screenshot 2024-10-23 at 12 38 25

  - Console:
Screenshot 2024-10-23 at 12 38 38

  - In web console:
Screenshot 2024-10-23 at 09 50 06

This fix was tested with the following demos:
 - rn-supabase-group-chat
 - rn-supabase-todolist (demo that highlighted the issue)
 - react-supabase-todolist


packages/common/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -117,18 +118,23 @@ export abstract class AbstractRemote {
...
     if (credentials != null && (credentials.endpoint == null || credentials.endpoint == '')) {
       throw new Error('PowerSync endpoint not configured');
     } else if (credentials?.token == null || credentials?.token == '') {
       const error: any = new Error(`Not signed in`);
       error.status = 401;
       throw error;
+    } else if (credentials?.endpoint.match(POWERSYNC_TRAILING_SLASH_MATCH)) {
+      this.logger.warn('A trailing forward slash "/" was entered after the POWERSYNC_URL environment variable. ' +
+        'Remove the trailing "/" from the variable to get rid of this warning.');
+      endPoint = credentials?.endpoint.replace(POWERSYNC_TRAILING_SLASH_MATCH, "");
```
rkistner, line 131: This check can be moved to the `getCredentials` function, to avoid duplicating it.

I think this should also be a hard error rather than a warning - it will simplify our code, and the developer can just immediately fix the issue.

HeinrichvonStein: Thanks for the feedback @rkistner .

I was following the rule of three, but will avoid the duplicate.

If an error is thrown and it's up to the developer to fix the issue, then the regex replace can be removed?

HeinrichvonStein:
> * introducing a new warning message

packages/common/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -108,6 +109,11 @@ export abstract class AbstractRemote {
       return this.credentials!;
     }
     this.credentials = await this.connector.fetchCredentials();
+    if (this.credentials?.endpoint.match(POWERSYNC_TRAILING_SLASH_MATCH)) {
+      throw new Error(
+        'A trailing forward slash "/" was entered after the POWERSYNC_URL environment variable. Remove the trailing "/" from the variable to fix this error.'
```
stevensJourney, line 114: The error message here is a bit focussed from the perspective of our demos. The PowerSync SDK does not directly interface with environment variables. We should update this message to reference the BackendConnector's `fetchCredentials` response instead.

HeinrichvonStein: Thanks for the feedback @stevensJourney.
This was an oversight on my part. Will make the necessary changes.

stevensJourney (approved):
LGTM

PR #343, opened 2024-10-14
powersync-sqlite-core 0.3.0 for op-sqlite

rkistner:
Follow-up to #334 to also update powersync-sqlite-core with `@powersync/op-sqlite`.


mugikhan (approved):
LGTM

PR #353, opened 2024-10-15
feature: added `useSuspenseQuery` hook to react package

Chriztiaan:
description pending...


DominicGBauer (commented):
Can you add tests for this?

rkistner (commented):
I haven't tested yet, but looks great overall

packages/react/src/WatchedQuery.ts
```diff
@@ -0,0 +1,175 @@
...
+      }
+      this.maybeDispose();
+    };
+
+    const timeoutRelease = () => {
+      if (this.isReady || this.controller == null) {
+        release();
+      } else {
+        // If the query is taking long, keep the temporary hold.
+        timeout = setTimeout(release, 5_000);
```
rkistner, line 59: Is this bit correct?

As I read this:
1. Set initial timeout of 5s.
2. Once that timeout expired, if the query is done executing (`isReady = true`) - all good.
3. If still waiting for the query, wait another 5s.
4. If the component rendered in that time, `clearTimeout` is called - all good.
5. If not, `release()` is called.

So overall, this just gives us a 10-second timeout in which the query must complete?

packages/react/src/hooks/useSuspenseQuery.ts
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import { getQueryStore } from '../QueryStore';
+import { usePowerSync } from './PowerSyncContext';
+import { CompilableQuery, ParsedQuery, parseQuery, SQLWatchOptions } from '@powersync/common';
+import { WatchedQuery } from '../WatchedQuery';
+import { QueryResult } from './useQuery';
+
+export type SuspenseQueryResult<T> = Pick<QueryResult<T>, 'data' | 'refresh'>;
+
+export const useSuspenseQuery = <T = any>(
```
DominicGBauer, line 10: To make this more testable and easier to test you could do something like this:
```
import React from 'react';
import { getQueryStore } from '../QueryStore';
import { usePowerSync } from './PowerSyncContext';
import { CompilableQuery, ParsedQuery, parseQuery, SQLWatchOptions } from '@powersync/common';
import { WatchedQuery } from '../WatchedQuery';

// Extracted function for parsing query
export function parseQuerySafely<T>(query: string | CompilableQuery<T>, parameters: any[]): ParsedQuery {
  try {
    return parseQuery(query, parameters);
  } catch (error) {
    throw new Error(`Failed to parse query: ${error.message}`);
  }
}

// Extracted function for generating key
export function generateQueryKey(parsedQuery: ParsedQuery, options: Omit<SQLWatchOptions, 'signal'>): string {
  return `${parsedQuery.sqlStatement} -- ${JSON.stringify(parsedQuery.parameters)} -- ${JSON.stringify(options)}`;
}

// Extracted function for getting query from store
export function getQueryFromStore(
  store: ReturnType<typeof getQueryStore>,
  key: string,
  query: string | CompilableQuery<any>,
  parsedQuery: ParsedQuery,
  options: Omit<SQLWatchOptions, 'signal'>
): WatchedQuery {
  return store.getQuery(
    key,
    { rawQuery: query, sqlStatement: parsedQuery.sqlStatement, queryParameters: parsedQuery.parameters },
    options
  );
}

// Custom hook for PowerSync setup
export function usePowerSyncSetup() {
  const powerSync = usePowerSync();
  if (!powerSync) {
    throw new Error('PowerSync not configured.');
  }
  return powerSync;
}

// Main hook that uses the extracted functions
export function useSuspenseQuery<T>(
  query: string | CompilableQuery<T>,
  parameters: any[] = [],
  options: Omit<SQLWatchOptions, 'signal'> = {}
) {
  const powerSync = usePowerSyncSetup();

  const parsedQuery = React.useMemo(() => parseQuerySafely(query, parameters), [query, parameters]);
  const key = React.useMemo(() => generateQueryKey(parsedQuery, options), [parsedQuery, options]);
  
  const store = getQueryStore(powerSync);
  const q = React.useMemo(() => getQueryFromStore(store, key, query, parsedQuery, options), [store, key, query, parsedQuery, options]);

  return { parsedQuery, key, q };
}
```

packages/react/src/WatchedQuery.ts
Chriztiaan: Good catch, this should be `timeout = setTimeout(timeoutRelease, 5_000);`.
Which would be:

1. Set initial timeout of 5s.
2. Once that timeout expired, if the query is done executing (isReady = true) - all good.
3. If still waiting for the query, wait another 5s.
4. If the component rendered in that time, clearTimeout is called - all good.
5. If not, go to step 3.

packages/react/src/hooks/useSuspenseQuery.ts
Chriztiaan: I have added a few tests for both the query store and for the useSuspenseQuery hook, I opted to not break most of the hook's code into smaller functions as it might make it harder to follow where the throwing happens. The hook's tests loosely follow the same testing as `useQuery`, except that the suspense and error boundaries are used to capture the values.

DominicGBauer: Nice thanks!

DominicGBauer:
LGTM

DominicGBauer (dismissed):

Chriztiaan:
Looking at tests.

Chriztiaan:
@rkistner @DominicGBauer I made a few changes to the PR to extend the `BaseObserver` class.

packages/vue/README.md
```diff
@@ -1,5 +1,7 @@
 # Vue composables for PowerSync
 
+The `powersync/vue` package package is a Vue-specific wrapper for PowerSync. It provides Vue composables that are designed to support reactivity, and can be used to automatically re-render components when query results update or to access PowerSync connectivity status changes.
```
benitav, line 3:
```suggestion
The `powersync/vue` package is a Vue-specific wrapper for PowerSync. It provides Vue composables that are designed to support reactivity, and can be used to automatically re-render components when query results update or to access PowerSync connectivity status changes.
```

stevensJourney (approved):

PR #356, opened 2024-10-15
chore: add check for navigator locks

DominicGBauer:
## Description 
Currently, apps do not work as expected when users use an insecure context (e.g. http://127.0.0.2) which results in confusing behaviour for users (e.g. https://discord.com/channels/1138230179878154300/1291056749197131796).

This PR adds a check to confirm that navigator locks are present and throws an error otherwise. This should avoid users experiencing the above mentioned issue.

This also includes vitest updates as they are required to avoid false positives see here (https://github.com/vitest-dev/vitest/issues/6187)

## Testing
I've added unit tests to confirm the behaviour


packages/web/src/shared/navigator.ts
```diff
@@ -0,0 +1,50 @@
+import { Mutex } from 'async-mutex';
+
+export const getNavigationLocks = (): LockManager => {
```
rkistner, line 3: This creates a new Map every time it's called, which is probably not the idea? Perhaps it should be called once and assigned to a global var?

rkistner (commented):
What are the typical cases where `navigator.locks` is not available?
Could this cause other issues, such as multiple tabs syncing concurrently?

DominicGBauer:
@rkistner based on this https://discord.com/channels/1138230179878154300/1291056749197131796 the general issue is an unsecure context being used (e.g. 192.168...) which results in hard to diagnose issues for users. The idea with this is to warn users when navigator locks are not available and to then have a fallback so they can still continue albeit with more limited functionality.

rkistner:
Ok, I wasn't aware of the secure context limitation.

Is it feasible to limit this to only "safe" setups, such as when using a SharedWorker? I'm concerned that silent data corruption because of a lack of cross-tab locks could be even more difficult to debug.

stevensJourney:
I did some digging and found that WA-SQLite also uses the `navigator.locks` API internally. From what I could see Navigator locks are currently unavoidable if we use the `IDBBatchAtomicVFS` VFS which uses [WebLocks](https://github.com/powersync-ja/wa-sqlite/blob/59c792fb3b778c9032901d4425f4c17edc597bcc/src/examples/IDBBatchAtomicVFS.js#L105) under the [hood](https://github.com/powersync-ja/wa-sqlite/blob/59c792fb3b778c9032901d4425f4c17edc597bcc/src/examples/WebLocks.js#L183).

rkistner (approved):

gavin-ob:
HI there, is this work complete? And if so do you have any idea when it will be released? I think I am in a situation where this would benefit me, I am building a nextjs application and then converting it to ios and android app using capacitor. Powersync seems to work well when the apps are bundled fully, but during the development process I am pointing to local host and using the nextjs instance that is running there, which enables hot reloading of my changes in xcode and android studio, this method doesn't work with powersync, I get an error that mentions navigator locks. 

Apologies if this is unrelated, but would be a massive help if your changes solve it! happy to test them on this use case if you can point me in the direction of a package with the changes implemented.

rkistner:
This is released already, but it only checks that `navigator.locks` are present, giving a better error message when it isn't.

Even if we removed usage of the locks here, wa-sqlite still uses the same `navigator.locks` internally, so it is something that is required.

You say you're pointing to local host in development - is that `http://localhost:<port>`, or for example using your WiFi IP? localhost should work as far as I know, but the WiFi IP won't. At least for Android you can use `adb reverse` to expose your development server as an actual `http://localhost` URL on the device. Not sure if there is an equivalent for iOS.

gavin-ob:
@rkistner Thanks very much for the response, I will give that a try

PR #357, opened 2024-10-16
Add client parameter definition on Diagnostics App

Omnicpie:
Adds a new area to the diagnostics app that allows the user to add and remove client parameters, allowing them to test more scenarios for buckets and the amount of data that will be synced to a system. 

Users can add new rows, define Keys and Values, and Remove rows. 

When a user presses "Submit" these key-value pairs will be converted to an object and the sync will re-connect with the new params. This is persisted throughout the app using local storage. 

If a user refreshes the page, the client params page will parse the value in local storage and convert it back to an array, allowing users to edit/remove items.


Chriztiaan:
Tested, works on my side!

@Omnicpie, there is minor cleanup I'd ask for to get this over the line if you don't mind? 

Can we put `'currentParams'` into a const in the `ConnectionManager` and move this snippet
```
 const stringifiedParams = localStorage.getItem('currentParams');
    const params = safeParse(stringifiedParams);
```
into some `getParams` helper method in that file so that there is less duplication?

Omnicpie:
@Chriztiaan Updated with the changes

Chriztiaan (approved):

PR #362, opened 2024-10-21
fix: `useQuery()`'s `isFetching` value was stuck as true in react-native.

Chriztiaan:
The `isFetching` value was stuck on true in react-native projects if either the query or parameters changed. It seems state changes trigger in a slightly different order.

Verified that this package works via a dev release on the reporter's `react-native` project and on the `react` project that originally needed improvements for `isFetching`.


mugikhan (approved):
LGTM!

PR #373, opened 2024-10-30
Add attachment storage options for React Native Todo Demo App

HeinrichvonStein:
This adds the functionality for a user to choose between either supabase or AWS S3 for attachment storage by changing an environment variable.

Supabase is still used for authentication.

The README contains updated instructions on how to setup an S3 bucket as well as configuring the env variables.


benitav:
Re: questions/concerns - my feeling is that it would be cleaner (less duplication and less to maintain) if we add the S3 implementation differences/instructions separately. README probably makes sense, because I assume the instructions are heavily tied to this implementation. Alternatively in docs, https://docs.powersync.com/usage/use-case-examples/attachments-files, but that might be difficult to remember to update if we update the example implementation.

PR #376, opened 2024-10-31
Feat: Queue read connections OPSQLite

mugikhan:
## Work done

- 5 concurrent read-only connections
- Use a queue to manage queries when all connections are busy
- Process queued operations as soon as connections become available

## Testing

Tested locally running multiple queries running at the same time with random delays between the queries.

### DIFF
--- a/packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts
+++ b/packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts
@@ -35,9 +35,11 @@ export class OPSQLiteDBAdapter extends BaseObserver<DBAdapterListener> implement
 
   protected initialized: Promise<void>;
 
-  protected readConnections: OPSQLiteConnection[] | null;
+  protected readConnections: Array<{ lockKey: string; connection: OPSQLiteConnection }> | null;
 
   protected writeConnection: OPSQLiteConnection | null;
+
+  private readQueue: Array<() => void> = [];
 
   constructor(protected options: OPSQLiteAdapterOptions) {
     super();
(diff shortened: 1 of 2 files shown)


packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts
```diff
@@ -145,36 +147,45 @@ export class OPSQLiteDBAdapter extends BaseObserver<DBAdapterListener> implement
...
-    const sortedConnections = this.readConnections!.map((connection, index) => ({
-      lockKey: `${LockType.READ}-${index}`,
-      connection
-    })).sort((a, b) => {
-      const aBusy = this.locks.isBusy(a.lockKey);
-      const bBusy = this.locks.isBusy(b.lockKey);
-      // Sort by ones which are not busy
-      return aBusy > bBusy ? 1 : 0;
+    return new Promise(async (resolve, reject) => {
+      const execute = async () => {
```
stevensJourney, line 157: I think we could remove the `execute` function definition here and just do this inline.

packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts
```diff
@@ -145,36 +147,45 @@ export class OPSQLiteDBAdapter extends BaseObserver<DBAdapterListener> implement
...
+        // If we have an available connection, use it
+        if (availableConnection) {
+          availableConnection.busy = true;
+          try {
+            resolve(await fn(availableConnection.connection));
+          } catch (error) {
+            reject(error);
+          } finally {
+            // After query execution, process any queued tasks
+            this.processQueue();
```
stevensJourney, line 170: Should this set `busy` to false?

packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts
stevensJourney: Sorry, I see this uses recursion

mugikhan: It uses the execute function to push it to the queue. Without it it makes the processQueue repeat what is done in the read lock.

stevensJourney (approved):

PR #377, opened 2024-10-31
feat: refresh watched queries on schema changes

Chriztiaan:
This PR introduces fixes for problems encountered when schema changes are made against tables being depended on in watched queries (`watch()` and hooks/composable).

The problem affected web and react-native in different ways. 
- For web the solution is to have the watched queries register a listener which is triggered on schema change.  This fix was applied to `common/AbstractPowerSyncDatabase`'s watch functions, `react`, `tanstack-react-query`, and `vue`.
- For react-native in combination with the web fix, the underlying RNQS or OP-sqlite needs to a pragma call on all read/write connections. The OP-sqlite change is included in this PR, while the RNQS change is available [here](https://github.com/powersync-ja/react-native-quick-sqlite/pull/56).

A similar API addition for `refreshSchema()` was made to the Dart `sqlite-async` project [here](https://github.com/powersync-ja/sqlite_async.dart/pull/57).

This PR also includes a small fix to the optional sync's demo (this demo was used for testing the original bug).

## Testing
 
Originally, after the schema was changed - inserts wouldn't reflect in the UI for watch watched queries. Inserts do reflect now as expected.


packages/common/src/db/DBAdapter.ts
```diff
@@ -101,6 +101,7 @@ export interface DBAdapter extends BaseObserverInterface<DBAdapterListener>, DBG
   readTransaction: <T>(fn: (tx: Transaction) => Promise<T>, options?: DBLockOptions) => Promise<T>;
   writeLock: <T>(fn: (tx: LockContext) => Promise<T>, options?: DBLockOptions) => Promise<T>;
   writeTransaction: <T>(fn: (tx: Transaction) => Promise<T>, options?: DBLockOptions) => Promise<T>;
+  refreshSchema: () => Promise<void>;
```
Chriztiaan, line 104: - [ ] Add documentation making it clear that this is called automatically as part of updateSchema, and users only need to worry about this for schema manipulation outside of that flow (if any).

stevensJourney (approved):

stevensJourney (approved):

PR #380, opened 2024-11-04
feat: drizzle

Chriztiaan:
Introducing an alpha version of the Drizzle ORM package. Following similar structure as the Kysely driver. 

Mintlify docs PR [here](https://github.com/powersync-ja/mintlify-docs/pull/25).

This has been tested across web and react-native (react-native-quick-sqlite and opslite).

# Todos
- [x] Unit tests


ImSingee:
Very happy to see it! BTW, is there any ETA for this? I'm developing a new project and considering waiting on this.

kobiebotha:
> Very happy to see it! BTW, is there any ETA for this? I'm developing a new project and considering waiting on this.

Somewhere in the [days/1-2 weeks range], not months.

packages/drizzle-driver/src/sqlite/db.ts
```diff
@@ -0,0 +1,52 @@
+import { AbstractPowerSyncDatabase, QueryResult } from '@powersync/common';
```
DominicGBauer, line 1: You probably know this is coming but please can we have tests for the new package? The kysely tests should provide a good starting point for the setup.

Chriztiaan: Added the initial test suite, it will be extended when schema inference is added (possibly a follow up PR).

packages/drizzle-driver/src/sqlite/db.ts
```diff
@@ -0,0 +1,52 @@
...
+    config?: PowerSyncSQLiteTransactionConfig
+  ): Promise<T>;
+}
+
+export function wrapPowerSyncWithDrizzle<TSchema extends Record<string, unknown> = Record<string, never>>(
+  db: AbstractPowerSyncDatabase,
+  config: DrizzleConfig<TSchema> = {}
+): PowerSyncSQLiteDatabase<TSchema> {
+  const dialect = new SQLiteAsyncDialect();
+  let logger;
```
DominicGBauer, line 31: I think this can be simplifed to 
`const logger = config.logger || new DefaultLogger();` ?

packages/drizzle-driver/src/sqlite/db.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (config.logger === true) {
+    logger = new DefaultLogger();
+  } else if (config.logger !== false) {
+    logger = config.logger;
+  }
+
+  let schema: RelationalSchemaConfig<TablesRelationalConfig> | undefined;
```
DominicGBauer, line 38: should it not throw an error if there is no schema?

packages/drizzle-driver/src/sqlite/sqlite-query.ts
```diff
@@ -0,0 +1,155 @@
...
+    const { fields, query, logger, customResultMapper } = this;
+    if (!fields && !customResultMapper) {
+      const params = fillPlaceholders(query.params, placeholderValues ?? {});
+      logger.logQuery(query.sql, params);
+      const rs = await this.db.execute(this.query.sql, params);
+      return rs.rows?._array ?? [];
+    }
+
+    const rows = (await this.values(placeholderValues)) as unknown[][];
+    const plain = rows.map((row) => Object.values(row));
```
DominicGBauer, line 56: what does plain mean in this context?

packages/drizzle-driver/src/sqlite/sqlite-query.ts
```diff
@@ -0,0 +1,155 @@
...
+    if (is(field, Column)) {
+      decoder = field;
+    } else if (is(field, SQL)) {
+      decoder = (field as any).decoder;
+    } else {
+      decoder = (field.sql as any).decoder;
+    }
+    let node = result;
+    for (const [pathChunkIndex, pathChunk] of path.entries()) {
+      if (pathChunkIndex < path.length - 1) {
```
DominicGBauer, line 120: there's a lot going on in here, can this be broken up into smaller methods to make it more clear what is happening?

packages/drizzle-driver/src/sqlite/sqlite-session.ts
```diff
@@ -0,0 +1,98 @@
...
+    );
+  }
+
+  override transaction<T>(
+    transaction: (tx: PowerSyncSQLiteTransaction<TFullSchema, TSchema>) => T,
+    config: PowerSyncSQLiteTransactionConfig = {}
+  ): T {
+    const { accessMode = 'read write' } = config;
+
+    if (accessMode == 'read only') {
```
DominicGBauer, line 75: can this use === ?

packages/drizzle-driver/src/sqlite/db.ts
Chriztiaan: Because `config.logger` is of type `DrizzleConfig<TSchema>.logger?: boolean | Logger | undefined`, that shortened version maps to `true | Logger` instead of `Logger | undefined`.

packages/drizzle-driver/src/sqlite/db.ts
Chriztiaan: It could. But if we look at all the reference packages like [Expo's](https://github.com/drizzle-team/drizzle-orm/blob/main/drizzle-orm/src/expo-sqlite/driver.ts) and [OP](https://github.com/drizzle-team/drizzle-orm/blob/main/drizzle-orm/src/op-sqlite/driver.ts), they don't seem to do that - so mostly just trying to stay consistent.

Even without a schema you can still use certain drizzle functions which is cool.

packages/drizzle-driver/src/sqlite/sqlite-session.ts
Chriztiaan: Yarp

packages/drizzle-driver/src/sqlite/sqlite-query.ts
Chriztiaan: It's just a simplified/stripped version of the rows with only the values. It could be called something like values or rowValues?

packages/drizzle-driver/src/sqlite/sqlite-query.ts
Chriztiaan: Broke it up a bit with minor doc comments.
I think taking it further than this might make it harder to follow.

packages/drizzle-driver/src/sqlite/db.ts
DominicGBauer: Does it make sense to use Drizzle without a schema? Sorry I'm not well versed in it so just want to understand how things work.

Chriztiaan: Well I don't think it's the common case. It's more that it is allowed, for example you could still use raw queries against the schemaless Drizzle client if you wanted to.

DominicGBauer: Do we want to allow that given we expect a schema?

packages/drizzle-driver/src/sqlite/sqlite-query.ts
Chriztiaan: Sorted

packages/drizzle-driver/src/sqlite/db.ts
Chriztiaan: I am happy with having consistent behaviour compared to the other packages.
Additionally, you will notice problems in your code much earlier than runtime since something like `db.query.xxx` and `drizzleDb.select().from();` won't give you any options to work with.

packages/drizzle-driver/tests/sqlite/db.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    });
+    const result = await db.select().from(drizzleUsers).get();
+
+    expect(result!.name).toEqual('James Smith');
+  });
+
+  it('should insert a user and update that user within a transaction when raw sql is used', async () => {
+    await db.transaction(async (transaction) => {
+      await transaction.run(sql`INSERT INTO users (id, name) VALUES ('4', 'James');`);
+      await db.update(drizzleUsers).set({ name: 'James Smith' }).where(eq(drizzleUsers.name, 'James'));
```
rkistner, line 56: Shouldn't this be `transaction.update`? Does it actually work with `db.update`?

Chriztiaan: Yea that's mistake. The test does pass though, will check.

Chriztiaan: Works with both `db.update` and `transaction.update`, and the behaviour for putting rollbacks at the end of the transaction is the exact same.

rkistner (approved):

DominicGBauer (approved):

Chriztiaan:
Will apply @benitav 's suggestions in a follow up PR.

PR #384, opened 2024-11-05
chore: add kysely test example for transaction with raw sql and update deps

DominicGBauer:
## Description
Added a test example and updated the docs to show how to use raw sql in a transaction (in case more users have this question https://discord.com/channels/1138230179878154300/1303135609011048551/1303139332730785802)

## Work Done 
* Updated to kysely 0.27.4
* Updated dev deps
* Improved README 
* Cleaned up some imports
* Added a test case 

## Testing
All unit tests are passing

### DIFF
--- a/packages/kysely-driver/README.md
+++ b/packages/kysely-driver/README.md
@@ -147,21 +147,7 @@ const result = await db.selectFrom('users').select('name').executeTakeFirstOrThr
 // 'James Smith'
 ```
 
-- In PowerSync
-
-```js
-  await powerSyncDb.writeTransaction((transaction) => {
-    await transaction.execute('INSERT INTO users (id, name) VALUES(4, ?)', ['James']);
-    await transaction.execute("UPDATE users SET name = ? WHERE name = ?", ['James Smith', 'James']);
-  })
-  const result = await powerSyncDb.get('SELECT name FROM users WHERE name = ?', ['James Smith'])
-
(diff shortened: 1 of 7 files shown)


packages/kysely-driver/README.md
```diff
@@ -164,7 +154,33 @@ const result = await db.selectFrom('users').select('name').executeTakeFirstOrThr
...
+const result = await db.selectFrom('users').select('name').executeTakeFirstOrThrow();
+
+// 'James Smith'
+```
+
+- In PowerSync
+
+```js
+  await powerSyncDb.writeTransaction((transaction) => {
+    await sql`INSERT INTO users (id, name) VALUES ('4', 'James');`.execute(transaction)
```
rkistner, line 180: I assume `sql` is from kysely - does that actually work here?

Either way the plain PowerSync version supports raw SQL directly, so probably don't need this here?

DominicGBauer: This is more a guide for users if they have similar questions to this https://discord.com/channels/1138230179878154300/1303135609011048551/1303139332730785802. From what I could tell when looking at the transaction you need to use compiled kysely query if you use `executeQuery`. I think might be missing something though?

DominicGBauer: Ohhhh sorry misunderstood yes that makes sense

rkistner (approved):

PR #387, opened 2024-11-06
chore: move kysely out of beta

DominicGBauer:
## Description
Move Kysely out of Beta to v1.

This requires the docs to be updated. A PR for that has been created here https://github.com/powersync-ja/mintlify-docs/pull/18/files


packages/kysely-driver/README.md
```diff
@@ -2,10 +2,6 @@
 
 This package (`packages/kysely-driver`) brings the benefits of an ORM through our maintained [Kysely](https://kysely.dev/) driver to [PowerSync](https://powersync.com).
 
-## Beta Release
-
-The `kysely-driver` package is currently in a beta release.
-
 ## Getting started
```
benitav, line 5: @DominicGBauer I'm wondering if we should simply link to docs for getting started instructions and keep this readme minimal? That'll certainly help with maintenance going forward, else we'll need to duplicate everything and that's easy to forget. Maybe a good compromise it to keep the most basic getting started instructions in here (i.e. how to instantiate the DB with kysely, and switch to docs for usage and any other details we might add). You can decide what makes most sense.

DominicGBauer: Yeah that makes a lot of sense!

DominicGBauer: I like the idea of having a getting started but then a link if you want more info. I added a small example of a query in the getting started because I think that would be helpful.

benitav (approved):

PR #395, opened 2024-11-11
powersync-sqlite-core 0.3.6

rkistner:
See: https://github.com/powersync-ja/powersync-sqlite-core/issues/47

Fix issue with remove operations resulting in dangling rows in some cases.

### DIFF
--- a/.changeset/long-onions-double.md
+++ b/.changeset/long-onions-double.md
@@ -1,0 +1,7 @@
+---
+'@powersync/react-native': patch
+'@powersync/diagnostics-app': patch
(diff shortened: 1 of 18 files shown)


.changeset/long-onions-double.md
```diff
@@ -0,0 +1,7 @@
+---
```
Chriztiaan, line 1: Need a Kysley entry?

Chriztiaan: Nvm, dev dependency.

Chriztiaan (approved):

PR #402, opened 2024-11-14
set drizzle-orm as peer dependency

ImSingee:
The latest `drizzle-orm` version is `0.36.1`, but `@powersync/drizzle-driver` now requires a `drizzle-orm@0.35.2` dependency (This indicates `>=0.35.2 & < 0.36`).

This PR:
1. use peer dependency instead of dependency
2. allow a more wide range version of drizzle-orm (`>=0.35.2 & <1`)

### DIFF
--- a/packages/drizzle-driver/package.json
+++ b/packages/drizzle-driver/package.json
@@ -26,10 +26,8 @@
     "test": "vitest"
   },
   "peerDependencies": {
-    "@powersync/common": "workspace:^1.19.0"
-  },
-  "dependencies": {
-    "drizzle-orm": "0.35.2"
+    "@powersync/common": "workspace:^1.19.0",
+    "drizzle-orm": "^0.35.2"
   },
   "devDependencies": {
     "@powersync/web": "workspace:*",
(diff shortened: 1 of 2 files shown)


Chriztiaan:
The caret allows changes that do not modify the left-most non-zero element. So this change would only support `^0.35.2 := >=0.35.2 <0.36.0-0`. I will look into this.

ImSingee:
Oh yes.

Simply `>=0.35.2 <1` works. But I’m off computer, I will update it later (or I believe you can just help me update this)

Chriztiaan:
I am currently testing an alternative that we use for some of our other packages: https://github.com/powersync-ja/powersync-js/pull/404

ImSingee:
Hmm, I don’t think `*` is a good idea in any ways. Only god knows what will happen in the future. If we don’t want limit the lowest version, at least we should use `<1`

Chriztiaan:
Would be willing to test that it resolves the dependency issue on your side @ImSingee ?
I made a dev release at: `@powersync/drizzle-driver@0.0.0-dev-20241114122654`

ImSingee:
@Chriztiaan Yes, it works well with the latest drizzle-orm version

Chriztiaan:
Resolved in https://github.com/powersync-ja/powersync-js/pull/404

PR #408, opened 2024-11-15
feat: Converting Drizzle tables to Powersync tables

Chriztiaan:
# Description
With the current Drizzle integration users need to maintain two table schema definitions (one for Drizzle and one for PowerSync). The initial idea was to infer a Drizzle schema from the PowerSync schema, but based on feedback and some testing it's clear that a Drizzle schema can be more complex and extensive (for example it supports relations) than a PowerSync schema, so the first stab at having a unified schema definition would be to start from a Drizzle schema.

This PR introduces `DrizzleAppSchema` and `toPowerSyncTable`, which allows you to map your Drizzle schema, or tables individually, to a PowerSync AppSchema. The mechanism for defining PowerSync table options are illustrated in the Readme and unit tests. PowerSync table indices can be derived from the Drizzle table indices, so that option is omitted from the options.

# Limitations
- These mapping functions don't support any Drizzle column types beyond `text`, `integer`, `real`.

### DIFF
--- a/packages/drizzle-driver/package.json
+++ b/packages/drizzle-driver/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@powersync/drizzle-driver",
-  "version": "0.0.1",
+  "version": "0.1.0",
   "description": "Drizzle driver for PowerSync",
   "main": "lib/src/index.js",
   "types": "lib/src/index.d.ts",
--- a/packages/drizzle-driver/package.json
+++ b/packages/drizzle-driver/package.json
@@ -26,16 +26,15 @@
     "test": "vitest"
   },
   "peerDependencies": {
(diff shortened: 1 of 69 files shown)


packages/drizzle-driver/package.json
```diff
@@ -29,6 +29,9 @@
     "@powersync/common": "workspace:^1.19.0",
     "drizzle-orm": "<1.0.0"
   },
+  "dependencies": {
```
stevensJourney, line 32: Should this be a peer dependency and a dependency?

Chriztiaan: No

stevensJourney (approved):
LGTM

PR #409, opened 2024-11-18
feat: Web CLI utility for copying assets from node_modules to public

Chriztiaan:
Some platforms/projects like react-native-web require configuration of the webworkers, which include copying assets from the dist directory of `@powersync/web` to somewhere like `public`. This PR introduces a CLI utility, which can in the future house other CLI commands.

### DIFF
--- a/packages/web/bin/powersync.js
+++ b/packages/web/bin/powersync.js
@@ -1,0 +1,41 @@
+#!/usr/bin/env node
+
+const { Command } = require('commander');
+const program = new Command();
+const path = require('path');
+const fsPromise = require('fs/promises');
+
+program
+  .name('powersync-web')
+  .description('CLI for PowerSync Web SDK utilities')
+  .version('0.0.1');
+
+program
+  .command('copy-assets')
+  .description('Copy assets to the specified output directory')
+  .option('-o, --output <directory>', 'output directory for assets', 'public')
+  .action(async(options) => {
+    const outputDir = options.output;
+    
+    console.log(`Target directory: ${outputDir}`);
+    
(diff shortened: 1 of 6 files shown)


packages/web/bin/powersync.js
```diff
@@ -0,0 +1,41 @@
...
+
+const { Command } = require('commander');
+const program = new Command();
+const path = require('path');
+const fsPromise = require('fs/promises');
+
+program
+  .name('powersync-web')
+  .description('CLI for PowerSync Web SDK utilities')
+  .version('0.0.1');
```
stevensJourney, line 11: Can we use the same version as the NPM package here?

packages/web/bin/powersync.js
```diff
@@ -0,0 +1,41 @@
...
+
+
+program.parse(process.argv);
+
+async function copyDirectory(source, destination) {
+  try {
+    await fsPromise.cp(source, destination, { recursive: true });
+    console.log(`Assets copied from ${source} to ${destination}`);
+  } catch (err) {
+    console.error(`Error copying assets: ${err.message}`);
```
stevensJourney, line 39: We might want to exit the process with a non-zero exit code if this fails. This might be useful in cases like CI setups where if the copy fails a deploy should fail.

packages/web/bin/powersync.js
Chriztiaan: Updated and tested with a dev dependency:
image

stevensJourney (dismissed):

stevensJourney (approved):

stevensJourney (approved):

PR #430, opened 2024-12-04
fix: Drizzle query returning an array-in-array type:

Chriztiaan:
Resolved an issue where `.findMany` (and other multiple row results) queries had a type of `[][]` instead of `[]`. 

When checking the `db.watch(query, {onResult})`'s onResult type:
#### find-many:
```
(results: {
    id: string | null;
    assets: {
        id: string | null;
        customer_id: string | null;
    }[];
}[][]) => void
```
#### find-first:
```
results: ({
    id: string | null;
    customer_id: string | null;
    customer: {
        ...;
    } | null;
} | undefined)[]) => void
```

#### find-many:
```
(results: {
    id: string | null;
    assets: {
        customer_id: string | null;
        id: string | null;
    }[];
}[]) => void
```

#### find-first:
```
(results: ({
    id: string | null;
    customer_id: string | null;
    customer: {
        ...;
    } | null;
} | undefined)[]) => void
```

### DIFF
--- a/.changeset/honest-mugs-yawn.md
+++ b/.changeset/honest-mugs-yawn.md
@@ -1,0 +1,5 @@
+---
+'@powersync/drizzle-driver': patch
+---
+
+Fixed a typing issue related to queries returning multiple results when used in `db.watch()`.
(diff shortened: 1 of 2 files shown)


rkistner (approved):

steveninety:
Hi, I think findFirst returning an array is (still) a mistake.

I've noticed that using a PowerSync Drizzle findFirst returns an object (which is correct and expected),
but wrapped inside toCompilableQuery, and also inside onResult as we can see above, a findFirst returns an array (always of length 1 though).

PR #434, opened 2024-12-06
Implementation of 'connecting' sync status from Dart SDK

LucDeCaf:
Implementation of the `SyncStatus.connecting` flag in the Dart SDK as `SyncStatus.connecting()` and `SyncStatusOptions.connecting?: boolean` in SyncStatusOptions.

NB: If `connected === true` then `connecting === false` always.


LucDeCaf:
Putting this here for myself later.

## TODO

* Find all modifications to `SyncStatus.connected` and check if `SyncStatus.connecting` should also be modified
* Find all places where the shape of `SyncStatus` is important and check if it supports `SyncStatus.connecting`

stevensJourney (approved):
I tested this and it's working perfectly. Thanks for this!

PR #435, opened 2024-12-09
make drizzle toPowerSyncTable support mode

ImSingee:
Drizzle schema supports some "mode," which behaves like a type in the JavaScript world and persists data as other types in SQLite.

For example, when using `text({mode: 'json'})`, in JavaScript it's JSON, but it stores data as text in SQLite.

### DIFF
--- a/.changeset/flat-toes-judge.md
+++ b/.changeset/flat-toes-judge.md
@@ -1,0 +1,5 @@
+---
+'@powersync/drizzle-driver': minor
+---
+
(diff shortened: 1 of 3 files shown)


ImSingee:
Another change in this PR: I change the code to use drrizle's `[entityKind]` instead of js class's `.name`

Reason: The JS class name is normally changed by the bundler in the production environment, this will make the code works in development mode but not in production mode

Chriztiaan (approved):

rkistner (approved):

PR #438, opened 2024-12-10
chore: optimize the useStatus hook and underlying isEqual

Frank-vdm:
This prevents the 'spammy' useStatus behaviour
key changes:
- explicitly compare the values in the SyncStatus object, its more performant than the `JSON.stringify` alternative this is important since this is triggered often
- we use this isEqual in the usePowerSyncStatus hook to only update the state only when the value actually changes

### DIFF
--- a/packages/react/src/hooks/usePowerSyncStatus.ts
+++ b/packages/react/src/hooks/usePowerSyncStatus.ts
@@ -21,10 +21,13 @@ export const usePowerSyncStatus = () => {
 
   useEffect(() => {
     const listener = powerSync.registerListener({
-      statusChanged: (status) => {
-        setSyncStatus(status);
+      statusChanged: (newStatus) => {
+        setSyncStatus(prev => {
+          // Only update if the status is actually different
+          return prev.isEqual(newStatus) ? prev : newStatus
+        })
       }
-    });
+    })
 
     return () => listener?.();
   }, [powerSync]);
--- a/.changeset/two-plants-thank.md
+++ b/.changeset/two-plants-thank.md
@@ -1,0 +1,6 @@
+---
+"@powersync/common": patch
+"@powersync/react": patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/react/src/hooks/usePowerSyncStatus.ts
```diff
@@ -21,10 +21,13 @@ export const usePowerSyncStatus = () => {
 
   useEffect(() => {
     const listener = powerSync.registerListener({
-      statusChanged: (status) => {
-        setSyncStatus(status);
+      statusChanged: (newStatus) => {
+        setSyncStatus(prev => {
+          // Only update if the status is actually different
+          return prev.isEqual(newStatus) ? prev : newStatus
```
stevensJourney, line 27: The `isEqual` check should already be implemented [here](https://github.com/powersync-ja/powersync-js/blob/ed5bb49b5a1dc579050304fab847feb8d09b45c7/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts#L634) which should only cause the `statusChanged` listeners to be fired if the status actually changed. Is this necessary? If so there might be a different bug elsewhere.

Frank-vdm: From my experience, when I use the useStatus hook from powersync my app refreshes every 5 seconds even if the status has not actually changed

stevensJourney: @Frank-vdm I tried to reproduce the sync status hook updating periodically with no actual sync status changes or to see if the `statusChanged` listener was being fired when there was infact no change, but I could not really reproduce this. The only time I could (unreliably) reproduce was when making changes to the `usePowerSyncStatus` hook code and hot reloading the application - which seems like an edge case which would not happen for users of the SDK. 
Do you perhaps have any more information on how to reproduce this?

PR #439, opened 2024-12-11
Web database encryption

mugikhan:
## Description

This PR introduces support for database encryption in the Web SDK using `multiple ciphers`.

## Work done

- Add optional encryption key parameter for web
- Dynamically load multiple cipher module factories


packages/web/src/db/adapters/wa-sqlite/WASQLiteConnection.ts
```diff
@@ -146,15 +177,35 @@ export class WASqliteConnection
     return this._dbP;
   }
 
+  protected async executeEncryptionPragma(): Promise<void> {
+    if (this.options.encryptionKey && this._dbP) {
+      await this.executeSingleStatement(`PRAGMA key = "${this.options.encryptionKey}"`);
```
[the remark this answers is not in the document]
mugikhan: It shows this error if you try to execute the `PRAGMA` statement and the multiple cipher vfs hasn't been setup correctly. 
`Uncaught (in promise) Error: Setting key failed. Encryption is not supported by the VFS.`
On success, since we are using the default of sqleet, which doesn't really have any method of checking besides querying for data without key which will error.

packages/web/README.md
```diff
@@ -28,6 +28,54 @@ Install it in your app with:
 npm install @journeyapps/wa-sqlite
 ```
 
+### Encryption with Multiple Ciphers
```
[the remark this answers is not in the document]
mugikhan: I can dd it there as well, all the other sdks (RN and Flutter) both mention it in the README.  @benitav thoughts on were to place this documentation for encryption, only in the docs or both?

packages/web/src/db/PowerSyncDatabase.ts
```diff
@@ -56,14 +58,30 @@ type WithWebSyncOptions<Base> = Base & {
...
 
+export interface WebEncryptionOptions {
+  /**
+   * Encryption key for the database.
+   * If set, the database will be encrypted using Multiple Ciphers.
+   */
+  encryptionKey?: string;
+}
+
+type WithWebEncryptionOptions<Base> = Base & {
```
Chriztiaan, line 69: Could this be `type WithWebEncryptionOptions<Base> = Base & WebEncryptionOptions;`?

packages/web/README.md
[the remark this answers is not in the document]
Chriztiaan: Note that I am currently busy with doing the docs for OPFS, so I can just move this there.

packages/web/src/db/PowerSyncDatabase.ts
mugikhan: It can indeed. Fixed.

Chriztiaan (approved):

packages/web/src/db/PowerSyncDatabase.ts
```diff
@@ -56,14 +58,24 @@ type WithWebSyncOptions<Base> = Base & {
   sync?: WebSyncOptions;
 };
 
+export interface WebEncryptionOptions {
+  /**
+   * Encryption key for the database.
+   * If set, the database will be encrypted using Multiple Ciphers.
+   */
+  encryptionKey?: string;
```
stevensJourney, line 66: It would be nice if we could expose the cipher method here also

packages/web/README.md
```diff
@@ -28,6 +28,54 @@ Install it in your app with:
...
+// If you are using a custom WASQLiteOpenFactory, you need specify the encryption key inside the factory construtor
+export const db = new PowerSyncDatabase({
+  schema: AppSchema,
+  database: new WASQLiteOpenFactory({
+    dbFilename: 'examplsw1se112.db',
+    vfs: WASQLiteVFS.OPFSCoopSyncVFS,
+    // Encryption key for the database.
+    encryptionKey: 'your-encryption-key'
+    flags: {
+      enableMultiTabs: typeof SharedWorker !== 'undefined'
```
stevensJourney, line 60: let's maybe not document this yet. @Chriztiaan mentioned there are still some cases where this is not entirely supported yet.

packages/web/src/db/PowerSyncDatabase.ts
mugikhan: Sticking with the default of sqleet's ChaCha20 as its web and the file compatibility with other sdks is not that important. This can be done in a later PR.

packages/web/README.md
mugikhan: Cleaned up the readme to exclude the flags properties.

PR #440, opened 2024-12-12
fix: supabase integration guide links

DominicGBauer:
## Description
It appears this `https://docs.powersync.com/integration-guides/supabase` link has changed to `https://docs.powersync.com/integration-guides/supabase-+-powersync` so users were being redirected to the overview page. This fixes the links to map to the correct guide.


benitav:
Thanks Dom, I saw this old link in the Dashboard too the other day. Can you maybe also add a redirect on the mintlify docs for this - that should ensure existing old links out there will continue to work: https://github.com/powersync-ja/mintlify-docs/blob/docs/mint.json#L428

benitav (approved):

rkistner:
Perhaps we should add a redirect on the docs site? There may be other sites also using the old link.

Issue #431, opened 2024-12-04
Error: [op-sqlite] SQL execution error: powersync_replace_schema table "ps_data__" already exists


andrewhertog:
I'm getting the following error from powersync whenever I try to read or write from a local SQLite database

```
Error: [op-sqlite] SQL execution error: powersync_replace_schema table "ps_data__" already exists
```

I'm using React-Native with op-sqlite and drizzle orm and trying to sync with a Supabase database. I haven't been able to get a session to connect with powersync yet, as I keep getting the above message.

I've tested with some of your example code which seemed to work, but when I try to integrate it into the project I'm working on I'm getting the above, and I don't see any documentation on the error.

rkistner:
Could you post your client-side schema definition please? There is likely an issue somewhere in the schema definition.

andrewhertog:
[Here](https://github.com/genesis-ai-dev/langquest/blob/powersync/db/powersync/psSchema.ts) is our powersync Schema and [here](https://github.com/genesis-ai-dev/langquest/blob/powersync/db/drizzleSchema.ts) is our client sqlite schema

rkistner:
You need to pass an object to `new Schema({})`, not an array. I'm not sure why this is not picked up by validations though - that could be a bug

andrewhertog:
Thanks. This worked

PR #448, opened 2024-12-18
Add schema validation

HeinrichvonStein:
This PR addresses https://github.com/powersync-ja/powersync-js/issues/431#issuecomment-2522158014.

Add validation to the Schema constructor.
- We need to validate that the tables have a name here because a user could pass in an array Tables that don't have a name because they are using the V2 syntax.  Therefore, 'convertToClassicTables' won't be called on the tables resulting in a runtime error. 

Additional Tests:
- Added a new Schema test which checks that an error is thrown when a Schema is created incorrectly.

### DIFF
--- a/packages/common/tests/db/schema/Schema.test.ts
+++ b/packages/common/tests/db/schema/Schema.test.ts
@@ -5,13 +5,13 @@ import { column, ColumnType } from '../../../src/db/schema/Column';
 import { Column } from '../../../lib';
 
 describe('Schema', () => {
-  it('should fail if array is passed and not a table object', () => {
+  it('should fail if an array of tables using the new syntax is passed to schema', () => {
     const table1 = new Table({ name: column.text });
     const table2 = new Table({ age: { type: ColumnType.INTEGER } });
     expect(() => new Schema([table1, table2])).toThrow();
   });
 
-  it('should create a schema with an array of tables', () => {
+  it('should create a schema with an array of tables using the old syntax', () => {
     const table1 = new Table({
       name: 'table1',
       columns: [new Column({ name: 'name', type: ColumnType.TEXT })]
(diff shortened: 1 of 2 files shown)


HeinrichvonStein:
Converting to a draft for now due to validation oversite

packages/common/tests/db/schema/Schema.test.ts
```diff
@@ -2,14 +2,26 @@ import { describe, it, expect } from 'vitest';
 import { Schema } from '../../../src/db/schema/Schema';
 import { Table } from '../../../src/db/schema/Table';
 import { column, ColumnType } from '../../../src/db/schema/Column';
+import { Column } from '../../../lib';
 
 describe('Schema', () => {
+  it('should fail if array is passed and not a table object', () => {
```
DominicGBauer, line 8: Can you specify that it should fail if an array of tables using the new syntax is passed so "should fail if an array of tables using the new syntax is passed to schema"

packages/common/tests/db/schema/Schema.test.ts
```diff
@@ -2,14 +2,26 @@ import { describe, it, expect } from 'vitest';
...
+    const table1 = new Table({ name: column.text });
+    const table2 = new Table({ age: { type: ColumnType.INTEGER } });
+    expect(() => new Schema([table1, table2])).toThrow();
+  });
+
   it('should create a schema with an array of tables', () => {
```
DominicGBauer, line 14: "should create a schema with an array of tables using the old syntax"
just for clarity for someone who does not have context

DominicGBauer (dismissed):

packages/common/tests/db/schema/Schema.test.ts
```diff
@@ -2,14 +2,26 @@ import { describe, it, expect } from 'vitest';
 import { Schema } from '../../../src/db/schema/Schema';
 import { Table } from '../../../src/db/schema/Table';
 import { column, ColumnType } from '../../../src/db/schema/Column';
+import { Column } from '../../../lib';
```
stevensJourney, line 5: Nit: Importing from `lib` is technically okay for tests, but it's probably nicer to import from `src` for consistency.

stevensJourney (approved):

DominicGBauer (approved):

PR #454, opened 2024-12-23
chore: rename docs headings

DominicGBauer:
## Description
Rename JS Doc headings to use the frameworks nomenclature.

image

### DIFF
--- a/docs/utils/packageMap.ts
+++ b/docs/utils/packageMap.ts
@@ -30,14 +30,14 @@ export const packageMap: PackageMap = {
     id: Packages.ReactNativeSdk
   },
   [Packages.ReactSdk]: {
-    name: 'React SDK',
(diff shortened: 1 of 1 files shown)


docs/utils/packageMap.ts
```diff
@@ -30,14 +30,14 @@ export const packageMap: PackageMap = {
     id: Packages.ReactNativeSdk
```
benitav, line 30: @DominicGBauer this LGTM. I noticed now, can we also update the order: React native SDK, Web SDK, React Hooks, Tanstack Ract Query Hooks, Vue Composables, Attachments SDK

DominicGBauer: Done

benitav (approved):

PR #455, opened 2024-12-23
chore: update docs deps

DominicGBauer:
## Description 
This updates the docs to use docusarus 3.6, typedoc 0.27 and other minor updates. 

From Docusarus 3.6 we can now use a faster flag which improves build time. This has been enabled.

## Testing
Rebuilding and re-running the docs showed no errors and clicking through the docs page appeared the same as previously.

### DIFF
--- a/docs/docusaurus.config.ts
+++ b/docs/docusaurus.config.ts
@@ -153,7 +153,10 @@ const config: Config = {
     prism: {
       theme: prismThemes.github,
       darkTheme: prismThemes.dracula
-    }
+    },
+    future: {
+      experimental_faster: true,
+    },
   } satisfies Preset.ThemeConfig
 };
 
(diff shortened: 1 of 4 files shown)


docs/docusaurus.config.ts
```diff
@@ -153,7 +153,10 @@ const config: Config = {
     prism: {
       theme: prismThemes.github,
       darkTheme: prismThemes.dracula
-    }
+    },
+    future: {
+      experimental_faster: true,
```
HeinrichvonStein, line 158: Is `experimental_faster`  the default name? If everything works as expected I'd change this to `faster` or something similar?

HeinrichvonStein:
LGTM!

HeinrichvonStein (approved):

PR #457, opened 2024-12-24
chore: add concurrency to actions

DominicGBauer:
## Description
Add concurrency to actions to avoid duplicate actions running after new commit. 

More info here https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-using-concurrency-to-cancel-any-in-progress-job-or-run

### DIFF
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -4,7 +4,7 @@ on:
   push:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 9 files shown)


DominicGBauer:
I wouldn't say we need this. Just feels like a waste IMO. Happy to cull it.

HeinrichvonStein:
Probably out of the scope of this PR, but should we not look at having a common workflow that the other workflows can 'inherit'/use?

rkistner:
GitHub Actions are pretty cheap to run (apart from the macos ones), I'd prefer to just keep it.

PR #461, opened 2025-01-08
fix: OP-sqlite write transaction single operation notification

Chriztiaan:
This PR addresses a timing issue in OPsqlite where the SQLite update hook triggers before the commit completes when performing single-write transactions (Resulting in `ps_crud` not being processed). As a result, `nextCrudItem` was returning undefined because the data was not yet committed at the time the hook fired. 

Using `react-native-quick-sqlite` as reference, updates are now batched and only flushed after a write lock is released.

### DIFF
--- a/.changeset/metal-fans-argue.md
+++ b/.changeset/metal-fans-argue.md
@@ -2,4 +2,5 @@
 '@powersync/op-sqlite': minor
 ---
 
-Fixed single write transanaction operations in `ps_crud` not being processed. Batching update notifications per write lock.
+Fixed single write transaction operations in `ps_crud` not being processed. Batching update notifications per write lock.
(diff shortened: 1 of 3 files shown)


rkistner (dismissed):
Changes look good.

I assume this would also affect regular watched queries?

Chriztiaan:
> Changes look good.
> 
> I assume this would also affect regular watched queries?

Using
```const { data: psCrud } = useQuery("Select * from ps_crud")``` initially I could see that it was stuck at `1`, but its definitely becoming 0 now after being processed. Do you mean WRT performance?

rkistner:
I mean that if you had `useQuery('select * from my_table')`, you could potentially get outdated results before because of the same issue, and that should be fixed now.

rkistner (approved):

stevensJourney (approved):

PR #468, opened 2025-01-10
chore: Navigate to login page when accessing pages that require auth

Chriztiaan:
This PR adds an auth guard that navigates the app to the login page if you navigate to a page that requires auth, or when the session expires.

Will apply this pattern to the other demos when this is accepted.

### DIFF
--- a/demos/react-supabase-todolist/src/app/router.tsx
+++ b/demos/react-supabase-todolist/src/app/router.tsx
@@ -1,4 +1,4 @@
-import { Outlet, createBrowserRouter } from 'react-router-dom';
+import { Outlet, createBrowserRouter, useNavigate } from 'react-router-dom';
 import LoginPage from '@/app/auth/login/page';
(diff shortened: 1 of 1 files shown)


demos/react-supabase-todolist/src/app/router.tsx
```diff
@@ -1,18 +1,61 @@
...
+  const navigate = useNavigate();
+  React.useEffect(() => {
+    if (!connector) {
+      console.error(`No Supabase connector has been created yet.`);
+      return;
+    }
+
+    connector.client.auth.onAuthStateChange(async (event, _session) => {
+      if (event === 'SIGNED_OUT') {
+        console.log("here");
```
stevensJourney, line 34: Should probably remove this

Chriztiaan: Sorted.

HeinrichvonStein (approved):

stevensJourney (approved):

PR #469, opened 2025-01-16
feat(op-sqlite): allow extensions to be loaded

DominicGBauer:
## Description
Allow additional sqlite extensions to be loaded by users as requested here https://github.com/powersync-ja/powersync-js/discussions/452#discussioncomment-11746598.

This includes and `extensions` option in `sqliteOptions` which would look something like this to the user
```
    let libPath: string
    if (Platform.OS === 'ios') {
      const bundlePath: string = NativeModules.PowerSyncOpSqlite.getBundlePath();
      libPath = getDylibPath('co.powersync.sqlite', 'powersync-sqlite-core')
    } else {
      libPath = 'libpowersync';
    }

    const factory = new OPSqliteOpenFactory({
      dbFilename: 'sqlite.db',
      sqliteOptions: {
        extensions: [
            { path: libPath, entryPoint: 'sqlite3_powersync_init' }
        ]
      }
    });
```

It also removes the use of `getBundlePath` in favor of [getDylibPath](https://op-engineering.github.io/op-sqlite/docs/api#loading-extensions)

We can't remove the entire Turbo Module because it includes the `powersync-sqlite-core` module for users without them needing to manually include it. 

## Testing
I excluded the powersync extension from being loaded in the `powersync-op-sqlite` library and used the above to include it from the client side in the react native supabase todolist demo. The powersync extension was loaded correctly. I also tested the case where a user does not include any extensions. 

I also ran a dev release and installed into an isolated react-native-supabase-todolist demo for both iOS and Android and it worked.


.changeset/hip-poems-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/op-sqlite': patch
```
Chriztiaan, line 2: Maybe make it minor since it's API enrichment with the additional option.

Chriztiaan (dismissed):

stevensJourney:
We should probably add some notes to the readme or docs about this feature.

For the example in the PR description, we could use an API from OP-SQLite for the iOS extension path:
See [getDylibPath](https://op-engineering.github.io/op-sqlite/docs/api#loading-extensions). We implemented something simillar in our OP-SQLite package, but we should update our package to use the OP SQLite API instead - this would make our adapter package much simpler.

.changeset/hip-poems-repair.md
DominicGBauer: Since we're still on v0 I figured patch meant minor in this context but happy to change it?

packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts
```diff
@@ -124,10 +138,17 @@ export class OPSQLiteDBAdapter extends BaseObserver<DBAdapterListener> implement
...
+        DB.loadExtension(extension.path, extension.entryPoint);
+      }
+    }
+  }
+
+  private loadPowerSyncExtension(DB: DB) {
     if (Platform.OS === 'ios') {
-      const bundlePath: string = NativeModules.PowerSyncOpSqlite.getBundlePath();
-      const libPath = `${bundlePath}/Frameworks/powersync-sqlite-core.framework/powersync-sqlite-core`;
+      const libPath = getDylibPath('powersync-sqlite-core', 'powersync-sqlite-core')
```
rkistner, line 151: Does this change mean we can remove all the native code in this package?

(Can leave that for another PR, it's probably a lot to change and test)

DominicGBauer:
@rkistner and @stevensJourney I needed to revert the change of using `getDylibPath`. I tested it incorrectly last week and after doing a final run through ran into an issue where it can't find the lib path using `getDylibPath`. It requires the `bundleId` but there is no `bundleId` in the powersync-sqlite-core `info.plist` so it is never found using `getDylibPath`. I'm not sure if this is why Mugi opted to do it this way originally.

DominicGBauer:
@stevensJourney, @Chriztiaan  and @rkistner this PR has been updated. Please can you re-review when you get a chance.

rkistner (approved):

PR #470, opened 2025-01-16
fix: Drizzle transactions not using lock context for React-Native

Chriztiaan:
## Description
As part of https://github.com/powersync-ja/powersync-js/pull/463 locks were fixed for `React-Native`, which resulted in Drizzle transactions on RN projects to never be executed. This was due to the original Drizzle transaction implementation not using the correct lock context. I could confirm that this issue was happening for both `RNQS` and `OPSqlite`, but not for web.

## Fix
`PowerSyncSQLiteSession` has been broken into `PowerSyncSQLiteSession` and `PowerSyncSQLiteBaseSession`, which ensures that a transaction uses the constrained LockContext instead of the db context (allowing transactions to be executed in the lock context).

This also makes room to eventually implement nested transactions/save points in `PowerSyncSqliteBaseSession`.

## Test
To verify this works across the board, I tested the fix for `RNQS`, `OPSqlite`, and Web.

------------

The source files have also been renamed to better match class names.

### DIFF
--- a/.changeset/lovely-ligers-sleep.md
+++ b/.changeset/lovely-ligers-sleep.md
@@ -1,0 +1,5 @@
+---
+'@powersync/drizzle-driver': patch
+---
+
+Fixed Drizzle transactions breaking for react-native projects, correctly using lock context for transactions.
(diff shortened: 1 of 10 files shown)


stevensJourney (approved):
Looks good to me.

rkistner (approved):

Issue #471, opened 2025-01-16
(drizzle) add support for custom column types


simonflk:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@powersync/drizzle-driver@0.3.0` for the project I'm working on.

I wanted to add a custom `timestamp` column type to map between ISO-formatted date strings and Date objects like so:

```ts
import { customType } from 'drizzle-orm/sqlite-core';

// https://orm.drizzle.team/docs/custom-types
export const sqliteTimestamp = customType<{ data: Date; driverData: string }>({
  dataType() {
    return 'text';
  },
  fromDriver(value) {
    return new Date(value);
  },
  toDriver(value) {
    return value.toISOString();
  },
});
```

However, I ran into an issue passing this schema to `DrizzleAppSchema`:

```
Error: Unsupported column type: SQLiteCustomColumn
```

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@powersync/drizzle-driver/lib/src/utils/schema.js b/node_modules/@powersync/drizzle-driver/lib/src/utils/schema.js
index 9db8abc..94e46cb 100644
--- a/node_modules/@powersync/drizzle-driver/lib/src/utils/schema.js
+++ b/node_modules/@powersync/drizzle-driver/lib/src/utils/schema.js
@@ -1,7 +1,7 @@
 import { column, Schema, Table } from '@powersync/common';
 import { entityKind, isTable } from 'drizzle-orm';
 import { CasingCache } from 'drizzle-orm/casing';
-import { getTableConfig, SQLiteBoolean, SQLiteInteger, SQLiteReal, SQLiteText, SQLiteTextJson, SQLiteTimestamp } from 'drizzle-orm/sqlite-core';
+import { getTableConfig, SQLiteBoolean, SQLiteCustomColumn, SQLiteInteger, SQLiteReal, SQLiteText, SQLiteTextJson, SQLiteTimestamp } from 'drizzle-orm/sqlite-core';
 export function toPowerSyncTable(table, options) {
     var _a, _b;
     const { columns: drizzleColumns, indexes: drizzleIndexes } = getTableConfig(table);
@@ -14,6 +14,7 @@ export function toPowerSyncTable(table, options) {
             continue;
         }
         let mappedType;
+        MapColumn:
         switch (drizzleColumn.columnType) {
             case SQLiteText[entityKind]:
[...]

PR #475, opened 2025-01-20
feat: Drizzle - added support for custom column types in schema conversion

Chriztiaan:
Adds support for custom types in the Drizzle/PowerSync schema definition - specifically with translating a Drizzle schema to PowerSync's app schema.

Bulk of the code was provided in this issue https://github.com/powersync-ja/powersync-js/issues/471.

### DIFF
--- a/packages/drizzle-driver/src/utils/schema.ts
+++ b/packages/drizzle-driver/src/utils/schema.ts
@@ -12,6 +12,7 @@ import { CasingCache } from 'drizzle-orm/casing';
 import {
   getTableConfig,
   SQLiteBoolean,
+  SQLiteCustomColumn,
   SQLiteInteger,
   SQLiteReal,
   SQLiteText,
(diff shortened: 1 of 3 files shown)


packages/drizzle-driver/src/utils/schema.ts
```diff
@@ -82,6 +66,34 @@ export function toPowerSyncTable<T extends SQLiteTableWithColumns<any>>(
...
+    case SQLiteTextJson[entityKind]:
+      return column.text;
+    case SQLiteInteger[entityKind]:
+    case SQLiteTimestamp[entityKind]:
+    case SQLiteBoolean[entityKind]:
+      return column.integer;
+    case SQLiteReal[entityKind]:
+      return column.real;
+    case SQLiteCustomColumn[entityKind]:
+      const sqlName = (drizzleColumn as any).sqlName; // Not exposed in the public API
```
stevensJourney, line 81: Why can't we use the `dataType` member here?

Chriztiaan: Sadly that gives you `custom` and not the type we need:
image

stevensJourney: After some poking around, it looks like we could do this:
```typescript
  const sqlName = (drizzleColumn as SQLiteCustomColumn<any>).getSQLType();
```

PR #478, opened 2025-01-22
feat: add throttle and delay to connect

DominicGBauer:
## Description
Add `retryDelayMs` and `crudUploadThrottleMs` to `connect` so that the values can be dynamically changed on reconnecting. Deprecate `retryDelay` option in favor of `retryDelayMs`.

## Testing
I ran the todolist react native demo and confirmed that the values used are from the connect method and added tests for web to confirm.

### DIFF
--- a/packages/common/src/client/AbstractPowerSyncDatabase.ts
+++ b/packages/common/src/client/AbstractPowerSyncDatabase.ts
@@ -39,6 +39,10 @@ export interface DisconnectAndClearOptions {
 export interface BasePowerSyncDatabaseOptions extends AdditionalConnectionOptions {
   /** Schema used for the local database. */
   schema: Schema;
+  /**
+   * @deprecated Use {@link retryDelayMs} instead as this will be removed in future releases.
+   */
+  retryDelay?: number;
   logger?: ILogger;
 }
 
--- a/packages/common/src/client/AbstractPowerSyncDatabase.ts
+++ b/packages/common/src/client/AbstractPowerSyncDatabase.ts
@@ -394,7 +394,7 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
 
     this.syncStreamImplementation = this.generateSyncStreamImplementation(connector, {
       crudUploadThrottleMs: options?.crudUploadThrottleMs,
-      retryDelay: options?.retryDelayMs
+      retryDelayMs: options?.retryDelayMs
     });
     this.syncStatusListenerDisposer = this.syncStreamImplementation.registerListener({
       statusChanged: (status) => {
--- a/packages/common/src/client/AbstractPowerSyncDatabase.ts
+++ b/packages/common/src/client/AbstractPowerSyncDatabase.ts
@@ -243,7 +243,11 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
   protected abstract openDBAdapter(options: PowerSyncDatabaseOptionsWithSettings): DBAdapter;
 
   protected abstract generateSyncStreamImplementation(
-    connector: PowerSyncBackendConnector
+    connector: PowerSyncBackendConnector,
+    options?: {
+      retryDelayMs?: number;
+      crudUploadThrottleMs?: number;
+    }
   ): StreamingSyncImplementation;
 
   protected abstract generateBucketStorageAdapter(): BucketStorageAdapter;
@@ -388,7 +392,10 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
       throw new Error('Cannot connect using a closed client');
     }
 
-    this.syncStreamImplementation = this.generateSyncStreamImplementation(connector);
+    this.syncStreamImplementation = this.generateSyncStreamImplementation(connector, {
+      crudUploadThrottleMs: options?.crudUploadThrottleMs,
+      retryDelay: options?.retryDelayMs
+    });
     this.syncStatusListenerDisposer = this.syncStreamImplementation.registerListener({
       statusChanged: (status) => {
         this.currentStatus = new SyncStatus({
--- a/packages/common/src/client/AbstractPowerSyncDatabase.ts
+++ b/packages/common/src/client/AbstractPowerSyncDatabase.ts
@@ -27,7 +27,8 @@ import {
   type AdditionalConnectionOptions,
   type PowerSyncConnectionOptions,
   StreamingSyncImplementation,
-  StreamingSyncImplementationListener
+  StreamingSyncImplementationListener,
+  DEFAULT_RETRY_DELAY_MS
 } from './sync/stream/AbstractStreamingSyncImplementation.js';
 import { runOnSchemaChange } from './runOnSchemaChange.js';
 
@@ -237,7 +238,7 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
 
   protected abstract generateSyncStreamImplementation(
     connector: PowerSyncBackendConnector,
-    options?: AdditionalConnectionOptions
+    options: Required<AdditionalConnectionOptions>
   ): StreamingSyncImplementation;
 
   protected abstract generateBucketStorageAdapter(): BucketStorageAdapter;
@@ -382,9 +383,13 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
       throw new Error('Cannot connect using a closed client');
     }
 
+    // Use the options passed in during connect, or fallback to the options set during database creation or fallback to the default options
+    const retryDelayMs = options?.retryDelayMs ?? this.options.retryDelayMs ?? this.options.retryDelay ?? DEFAULT_RETRY_DELAY_MS;
+    const crudUploadThrottleMs = options?.crudUploadThrottleMs ?? this.options.crudUploadThrottleMs ?? DEFAULT_CRUD_UPLOAD_THROTTLE_MS;
+
     this.syncStreamImplementation = this.generateSyncStreamImplementation(connector, {
-      crudUploadThrottleMs: options?.crudUploadThrottleMs,
-      retryDelayMs: options?.retryDelayMs
+      retryDelayMs,
+      crudUploadThrottleMs,
     });
     this.syncStatusListenerDisposer = this.syncStreamImplementation.registerListener({
       statusChanged: (status) => {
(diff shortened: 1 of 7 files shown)


DominicGBauer:
@stevensJourney this is ready for review

packages/react-native/src/db/PowerSyncDatabase.ts
```diff
@@ -42,9 +42,17 @@ export class PowerSyncDatabase extends AbstractPowerSyncDatabase {
...
+    connector: PowerSyncBackendConnector,
+    // This is used to pass in options on connection instead of only during database creation
+    options?: {
+      retryDelayMs: number;
+      crudUploadThrottleMs: number;
+    }
   ): AbstractStreamingSyncImplementation {
     const remote = new ReactNativeRemote(connector);
+    // Use the options passed in during connect, or fallback to the options set during database creation
+    const retryDelayMs = options?.retryDelayMs || this.options.retryDelay;
```
stevensJourney, line 54: I'm not sure what the intended behaviour for a `0` `retryDelayMs` is, but could we continue allowing that value to be used (if specified) by using `??` instead?

```javascript
const retryDelayMs = options?.retryDelayMs ?? this.options.retryDelay;
```

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -79,7 +79,19 @@ export interface PowerSyncConnectionOptions {
    * These parameters are passed to the sync rules, and will be available under the`user_parameters` object.
    */
   params?: Record<string, StreamingSyncRequestParameterType>;
+  /**
+   * Delay for retrying sync streaming operations
+   * from the PowerSync backend after an error occurs.
+   */
+  retryDelayMs?: number;
```
stevensJourney, line 86: It might be nice to move these `retryDelayMs` and `crudUploadThrottleMs` members to a smaller interface which is used also used for the `options` param in `generateSyncStreamImplementation`.

If we did that we could also introduce something along the lines of `BasePowerSyncConnectOptions` which would removes the need for `Omit` in `PowerSyncConnectionRequiredOptions` e.g.

```TypeScript
export interface BasePowerSyncConnectionOptions {
  /**
   * The connection method to use when streaming updates from
   * the PowerSync backend instance.
   * Defaults to a HTTP streaming connection.
   */
  connectionMethod?: SyncStreamConnectionMethod;

  /**
   * These parameters are passed to the sync rules, and will be available under the`user_parameters` object.
   */
  params?: Record<string, StreamingSyncRequestParameterType>;
}

export type PowerSyncConnectionOptions = BasePowerSyncConnectionOptions & [connection time interface]
export type PowerSyncConnectionRequiredOptions = Required<BasePowerSyncConnectionOptions>

```

packages/react-native/src/db/PowerSyncDatabase.ts
DominicGBauer: Oh yes thanks!

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
DominicGBauer: I agree but I think this is a better approach to get the same outcome

```
export interface PowerSyncConnectionOptionalOptions {
 
}

export interface PowerSyncConnectionRequiredOptions {
  /**
   * The connection method to use when streaming updates from
   * the PowerSync backend instance.
   * Defaults to a HTTP streaming connection.
   */
  connectionMethod?: SyncStreamConnectionMethod;

  /**
   * These parameters are passed to the sync rules, and will be available under the`user_parameters` object.
   */
  params?: Record<string, StreamingSyncRequestParameterType>;
}

export interface PowerSyncConnectionOptions extends PowerSyncConnectionOptionalOptions, PowerSyncConnectionRequiredOptions {}
```

stevensJourney: I don't have a strong opinion on this, but having optional members like `connectionMethod?` in the `PowerSyncConnectionRequiredOptions` interface seems a bit odd.

DominicGBauer: Do you know why they were made optional?

DominicGBauer: Looks like the ? is unnecessary and I have removed it

stevensJourney: IIRC the original `PowerSyncConnectionOptions` is used by the `connect` method. User's are not required to provide client parameters or `connectionMethod` each time they call `.connect`.
Just to clarify, having optional params in the newly renamed `PowerSyncConnectionRequiredOptions` is what seems odd to me.

packages/react-native/src/db/PowerSyncDatabase.ts
```diff
@@ -42,9 +42,17 @@ export class PowerSyncDatabase extends AbstractPowerSyncDatabase {
   }
 
   protected generateSyncStreamImplementation(
-    connector: PowerSyncBackendConnector
+    connector: PowerSyncBackendConnector,
+    // This is used to pass in options on connection instead of only during database creation
+    options?: {
+      retryDelayMs: number;
```
stevensJourney, line 48: This could use `AdditionalConnectionOptions`

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -35,21 +36,9 @@ export interface DisconnectAndClearOptions {
...
-export interface BasePowerSyncDatabaseOptions {
+export interface BasePowerSyncDatabaseOptions extends AdditionalConnectionOptions {
   /** Schema used for the local database. */
   schema: Schema;
-
-  /**
-   * Delay for retrying sync streaming operations
-   * from the PowerSync backend after an error occurs.
-   */
-  retryDelay?: number;
```
stevensJourney, line 46: It's very unfortunate, but I think renaming this to `retryDelayMs` is a breaking change. We might be able to around it, but we should keep the old field for consistency.

DominicGBauer: Yeah you're right. I've updated the PR to avoid the breaking change, deprecated this option in favour of `retryDelayMs` and added a test to confirm it still works.

packages/react-native/src/db/PowerSyncDatabase.ts
```diff
@@ -42,9 +43,14 @@ export class PowerSyncDatabase extends AbstractPowerSyncDatabase {
...
 
   protected generateSyncStreamImplementation(
-    connector: PowerSyncBackendConnector
+    connector: PowerSyncBackendConnector,
+    // This is used to pass in options on connection instead of only during database creation
+    options?: AdditionalConnectionOptions
   ): AbstractStreamingSyncImplementation {
     const remote = new ReactNativeRemote(connector);
+    // Use the options passed in during connect, or fallback to the options set during database creation
+    const retryDelayMs = options?.retryDelayMs ?? this.options.retryDelayMs ?? this.options.retryDelay;
```
stevensJourney, line 52: This is a bit of a Nit. With the growing logic here for resolving `retryDelayMs` and `crudUploadThrottleMs` I noticed this resolving is repeated in both the web and react native implementations. We could do this resolving once in the `AbstractPowerSyncDatabase` and pass the resolved options to `generateSyncStreamImplementation`.

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -388,7 +383,14 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
       throw new Error('Cannot connect using a closed client');
     }
 
-    this.syncStreamImplementation = this.generateSyncStreamImplementation(connector);
+    // Use the options passed in during connect, or fallback to the options set during database creation or fallback to the default options
+    const retryDelayMs = options?.retryDelayMs ?? this.options.retryDelayMs ?? this.options.retryDelay ?? DEFAULT_RETRY_DELAY_MS;
+    const crudUploadThrottleMs = options?.crudUploadThrottleMs ?? this.options.crudUploadThrottleMs ?? DEFAULT_CRUD_UPLOAD_THROTTLE_MS;
```
Chriztiaan, line 388: In `web/src/db/PowerSyncDatabase.ts` we have this resolve function, which is a pattern that might make this cleaner. 
```
export const resolveWebPowerSyncFlags = (flags?: WebPowerSyncFlags): Required<WebPowerSyncFlags> => {
  return {
    ...DEFAULT_POWERSYNC_FLAGS, // put your default options const here
    ...flags, // put your constructor options here
    ...resolveWebSQLFlags(flags) // put your connect function options here
  };
};
```

Your usage might then look like this here:
```
const { retryDelayMs, crudUploadThrottleMs } = resolveConnectOptions(options);
```

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -376,6 +372,15 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
     return this.waitForReady();
   }
 
+  // Use the options passed in during connect, or fallback to the options set during database creation or fallback to the default options
+  resolvedConnectionOptions(options?: PowerSyncConnectionOptions): RequiredAdditionalConnectionOptions {
+    return {
+      retryDelayMs: options?.retryDelayMs ?? this.options.retryDelayMs ?? this.options.retryDelay ?? DEFAULT_RETRY_DELAY_MS,
+      crudUploadThrottleMs:
+        options?.crudUploadThrottleMs ?? this.options.crudUploadThrottleMs ?? DEFAULT_CRUD_UPLOAD_THROTTLE_MS
```
Chriztiaan, line 380:
```
export const DEFAULT_ADDITIONAL_CONNECTION_OPTIONS: RequiredAdditionalConnectionOptions = {
  crudUploadThrottleMs: DEFAULT_CRUD_UPLOAD_THROTTLE_MS,
  retryDelayMs: DEFAULT_RETRY_DELAY_MS
};

  return {
      ...DEFAULT_ADDITIONAL_CONNECTION_OPTIONS,
      ...(this.options.retryDelay != null ? { retryDelayMs: this.options.retryDelay } : {}),
      ...this.options,
      ...options
    };
```

This is closer to what I meant with the original comment, but it's borderline nitty.

DominicGBauer: Mmm not sure I agree that this is better to be honest seems like it's a lot harder to reason about

Chriztiaan (approved):

PR #479, opened 2025-01-25
Remove iOS folders from demos

LucDeCaf:
iOS and android folders shouldn't be necessary anymore, have only checked if removing iOS folders works, not android.


stevensJourney (commented):
I'm happy with the change of removing the native projects from this repo. It should no longer be required to have those folder committed.
The same could be done for the Android folders, I tested this and didn't notice any issues. 
I'd be happy to approve this PR after it's been updated with the `main` branch.

stevensJourney (approved):

Issue #473, opened 2025-01-17
[Drizzle] bug: when doing joins (left join) columns are wrong


polvallverdu:
Posted this on discord.

Something wrong is going on when using drizzle. I'm going to try using raw sql. The first image is the relationed query, but data is totally wrong (id cannot be a date, and task should be null). Second image is the actual data that should be in the event.

![Image](https://github.com/user-attachments/assets/bb7bcf7a-4a4a-4e4d-abaa-f38ffd793e6a)

![Image](https://github.com/user-attachments/assets/8fb3ef73-ad7d-42cf-ae4f-d6dd4081b43b)

![Image](https://github.com/user-attachments/assets/9600734f-ef69-4872-85b7-f3c4411f5a0e)

Chriztiaan:
Resolved as part of tagged PR and confirmed on discord.

PR #482, opened 2025-01-28
Added executeRaw()

Chriztiaan:
## Description
At the moment duplicate column names in a query result are overwritten with only the last column being the one returned. This causes issues with join queries (a simple case is where multiple `id` columns are returned) and is most prominent when doing joins with one of the ORM drivers (Kysely/Drizzle). This issue is covered by the newly added [tests](https://github.com/powersync-ja/powersync-js/pull/482/files#diff-ac2f7287079fb1f152549bfc81461a48f84d4b9808f2772b355705ff9f9694fe).

This PR introduces `executeRaw` which processes the sqlite results differently and allows the Drizzle ORM driver to correctly map the resulting columns.

An existing approach for Drizzle-orm packages can be seen here:
https://github.com/drizzle-team/drizzle-orm/blob/main/drizzle-orm/src/op-sqlite/session.ts#L165
https://github.com/drizzle-team/drizzle-orm/blob/main/drizzle-orm/src/better-sqlite3/session.ts#L156 

## Sqlite Adapters Affected

1. RNQS
Opted to not implement `executeRaw` in the RNQS package for now, the adapter falls back to `execute` - resulting in the initial behaviour. Users who need this functionality are encouraged to migrate to OPSQLite at the moment.
2. OPSQLite
Using OPSQLite `executeRaw` function, which is also used for their Drizzle integration.
3. Web
Updated `WASQLiteConnection` to expose a raw execute function, simply handling the result differently.
4. Node
Using BetterSqlite3's `raw()` - which is also used for the BSQLite3 Drizzle integration.

 
## Original Workaround
For Drizzle users using joins the following example shows how to side-step the problem in the interim:
```
wait db
      .select({ id: drizzleUsers.id, postsId: sql`"posts"."id"`.as('postsId') }) // or postsId: drizzlePosts.id.getSQL().as('postsId')
      .from(drizzleUsers)
      .leftJoin(drizzlePosts, eq(drizzleUsers.id, drizzlePosts.user_id));
```

Reported in https://github.com/powersync-ja/powersync-js/issues/473


packages/node/tests/DrizzleNode.test.ts
Chriztiaan, line 1: This was the easiest way to verify that the node changeset worked. I am not sure if we should put a Drizzle based test for our Node adapter in the `Node` package or in the `Drizzle` package. Feels minor, but I think it's worth keeping this test around either way.

packages/react-native/src/db/adapters/react-native-quick-sqlite/RNQSDBAdapter.ts
```diff
@@ -43,25 +47,39 @@ export class RNQSDBAdapter extends BaseObserver<DBAdapterListener> implements DB
...
   execute(query: string, params?: any[]) {
     return this.baseDB.execute(query, params);
   }
 
+  /**
+   * 'executeRaw' is not implemented in RNQS, this falls back to 'execute'.
+   */
+  async executeRaw(query: string, params?: any[]): Promise<any[][]> {
+    const result = await this.baseDB.execute(query, params);
+    return result.rows?._array ?? [];
```
stevensJourney, line 80: Should this still unwrap the object responses to arrays of column values?

packages/react-native/src/db/adapters/react-native-quick-sqlite/RNQSDBAdapter.ts
```diff
@@ -75,6 +93,28 @@ export class RNQSDBAdapter extends BaseObserver<DBAdapterListener> implements DB
...
+      ...ctx,
+      // 'executeRaw' is not implemented in RNQS, this falls back to 'execute'.
+      executeRaw: async (sql: string, params?: any[]) => {
+        const result = await ctx.execute(sql, params);
+        return result.rows?._array ?? [];
+      }
+    };
+  }
+
+  generateTransactionContext(ctx: RNQSTransactionContext) {
```
stevensJourney, line 107: IIRC `RNQSTransactionContext` should be an extension of `RNQSLockContext`. Is it feasible to define this executeRaw override once?

packages/react-native/src/db/adapters/react-native-quick-sqlite/RNQSDBAdapter.ts
Chriztiaan: With the structure of the raw response explained, it's a bit more clear why this is needed. We don't have the column information available in most of the raw function calls from other drivers.

Chriztiaan: Discussed, we need to strip the columns here.

Which actually raised another issue, we have redundant Object.values() calls in our Drizzle package after switching to executeRaw. Will remove in this PR

Chriztiaan: @stevensJourney
1. I have updated RNQSAdapter to strip the column keys when calling `execute` in `executeRaw`.
2. I removed the redundant Object.values() calls in the Drizzle package.

stevensJourney (approved):

PR #493, opened 2025-02-07
Diagnostics app optimizations

rkistner:
This now disables slow diagnostics queries while downloading any data, so that the diagnostics app can stay fast while re-downloading bulk data.

This also adds tracking for downloaded operations, to view the progress while downloading.

Together with the switch to OPFS in #492, this makes the diagnostics app functional with syncing 1M operations (takes around 2m45s on my machine).


stevensJourney (approved):
The performance boost is really awesome!

PR #494, opened 2025-02-07
feat: Add FetchStrategy option to connect() for WebSockets

Chriztiaan:
This allows the fetch strategy to be configured, with the actual functionality addition being the `sequential` option.

When specifying `sequential`:
Instead of queueing up multiple sync events (10 originally) before applying them, which would lead to less round-trips - we are configuring the queue size to 1, meaning each event will get processed individually before asking for the next.
This results in less overhead processing of the netword events (pending items) and should free up the web socket for better keep alive acknowledgements.

Currently this resolves a known issue on slower devices with large sync datasets.

### DIFF
--- a/packages/common/src/client/sync/stream/AbstractRemote.ts
+++ b/packages/common/src/client/sync/stream/AbstractRemote.ts
@@ -21,13 +21,12 @@ export type RemoteConnector = {
 
 // Refresh at least 30 sec before it expires
 const REFRESH_CREDENTIALS_SAFETY_PERIOD_MS = 30_000;
-const SYNC_QUEUE_REQUEST_N = 1;
 const SYNC_QUEUE_REQUEST_LOW_WATER = 5;
 
 // Keep alive message is sent every period
-const KEEP_ALIVE_MS = 60_000;
+const KEEP_ALIVE_MS = 20_000;
 // The ACK must be received in this period
-const KEEP_ALIVE_LIFETIME_MS = 90_000;
+const KEEP_ALIVE_LIFETIME_MS = 30_000;
 
 export const DEFAULT_REMOTE_LOGGER = Logger.get('PowerSyncRemote');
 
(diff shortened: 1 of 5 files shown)


packages/common/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -216,8 +233,10 @@ export abstract class AbstractRemote {
   /**
    * Connects to the sync/stream websocket endpoint
    */
-  async socketStream(options: SyncStreamOptions): Promise<DataStream<StreamingSyncLine>> {
-    const { path } = options;
+  async socketStream(options: SocketSyncStreamOptions): Promise<DataStream<StreamingSyncLine>> {
+    const { path, fetchStrategy = FetchStrategy.Buffered } = options;
+
+    const syncQueueRequestSize = fetchStrategy == FetchStrategy.Buffered ? 10 : 1;
```
rkistner, line 239: Is there perhaps an in-between value that could work well for both slow and fast devices?

Chriztiaan: At the moment this does kinda feel like a stopgap, since I can't reproduce the issue without the physical device used for the React Native issue.

A similar PowerSync web issue which was related to a slow network connection was a clear fundamental issue with queueing is also solved by this flag - in which case it has nothing to do with hardware (maybe we should change our documentation around that in the changeset). But in this scenario: The connection is very slow and it struggles to download 10 batches of (requested) data. The server is trying to send the batches of data which hogs the socket, not allowing keep-alives to function correctly.

---
In pleasant network/hardware connections you get a little bit more performance out of the buffering, but it would be nice to automatically fallback if we detect something is taking longer than expected.  I would suggest adding that at a later point. 

For cases where a user knows the hardware is slower, this should be good enough for now.

rkistner: Ok, happy as a stopgap, but we should investigate further and properly fix this at some point.

FWIW, I see the keep-alive error even with the web SDK in some cases (but it recovers automatically after that).

rkistner (approved):

PR #496, opened 2025-02-09
fix(react-query): ensure useSuspenseQuery data is nonNullable

### LINKED ISSUE #495: [tanstack-react-query] Data type incorrectly shows as potentially undefined in useSuspenseQuery
Hello! I've been facing the following issue when using the `tanstack-react-query` integration.

# Description
When providing a `queryFn` that returns potentially undefined data the TypeScript definitions for `useSuspenseQuery` show the `data` property as potentially undefined. 
This is incorrect because `useSuspenseQuery`, like React Query's own implementation, guarantees that data will be available - it either returns data or throws an error (which would then be handled at an error boundary).

# Reproduction
Here's an example that uses the kysely-driver integration.

```
  const { data: entries } = useSuspenseQuery({
    queryKey: ['entries', goalId],
    queryFn: () =>
      db
        .selectFrom('entry')
        .selectAll()
        .where('goalId', '=', goalId)
        .orderBy('date', 'asc')
        .executeTakeFirst(),
  });
```
Image

leeluca:
This fixes #495

## Description
When providing a `queryFn` that returns potentially undefined data, the TypeScript definitions for `useSuspenseQuery` show the `data` property as potentially **undefined**. 

Since `useSuspenseQuery`, like React Query's own implementation, guarantees that data will be always available by throwing an error in case it is not, the set of types for `data` should not include **null** or **undefined**, regardless of the `queryFn`'s return value type definition. 

This PR updates the TypeScript definitions for `useSuspenseQuery` to correctly reflect that the returned data is always defined. 

## Changes
- Updated return type of `useSuspenseQuery` to use `NonNullable<TData>`

### DIFF
--- a/packages/tanstack-react-query/src/hooks/useQuery.ts
+++ b/packages/tanstack-react-query/src/hooks/useQuery.ts
@@ -32,7 +32,7 @@ export function useQuery<TData = unknown, TError = Tanstack.DefaultError>(
 export function useSuspenseQuery<TData = unknown, TError = Tanstack.DefaultError>(
   options: UseBaseQueryOptions<Tanstack.UseSuspenseQueryOptions<TData, TError>> & { query?: undefined },
   queryClient?: Tanstack.QueryClient
-): Tanstack.UseSuspenseQueryResult<TData, TError>;
+): Tanstack.UseSuspenseQueryResult<NonNullable<TData>, TError>;
 
 // Overload when 'query' is present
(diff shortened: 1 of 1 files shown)


leeluca:
Narrowing the return type inside the `queryFn` is a much better approach than trying to fix the hook types. 
I'll close this PR since the issue was solved (#495)

PR #498, opened 2025-02-10
[Fix] OPFS Sleeping Tabs Sync Deadlock

stevensJourney:
# Overview

We recently introduced OPFS VFS support for the PowerSync web SDK. Unlike IndexedDB-based VFS, which allows a single shared SQLite connection within a shared web worker across multiple tabs, OPFS requires each tab to have its own dedicated SQLite worker.

To manage concurrency, we use Navigator locks, ensuring exclusive access to SQLite connections. Table change notifications are propagated using a BroadcastChannel.

In multi-tab scenarios, a shared sync worker is responsible for syncing operations. This worker requires access to the SQLite database.

- IndexedDB VFS: A `MessagePort` to the shared SQLite worker is passed to the sync worker, allowing direct DB access.
- OPFS VFS: A `MessagePort` to the last connected tab’s dedicated SQLite worker is passed to the sync worker. This connection is updated as tabs are opened and closed.

The shared sync worker's DB connection is linked to a specific tab. If that tab is frozen or put to sleep (e.g., in Chrome’s Tab Freezing or Edge’s Sleeping Tabs), the connection becomes inaccessible, blocking SQLite queries in the sync worker.

Luckily browsers don't freeze or sleep tabs which are actively holding a web lock. This PR adds a hold on a web lock as soon as the connection is shared to the sync worker. In local testing this prevents tabs from being put to sleep.


packages/web/src/db/adapters/WorkerWrappedAsyncDatabaseConnection.ts
```diff
@@ -44,6 +50,38 @@ export class WorkerWrappedAsyncDatabaseConnection<Config extends ResolvedWebSQLO
...
+
+            // Free the lock when the connection is already closed.
+            if (this.lockAbortController.signal.aborted) {
+              return;
+            }
+
+            // Hold the lock while the shared connection is in use.
+            await new Promise<void>((releaseLock) => {
+              // We can use the resolver to free the lock
+              this.releaseSharedConnectionLock = releaseLock;
```
rkistner, line 78: Could you just wait for the `lockAbortController.signal` here, instead of using a separate callback?

stevensJourney: Yeah we can do that. I initially started testing without the AbortController where just this promise was used, using the AbortController signal is cleaner. This callback will be removed.

Chriztiaan (dismissed):

packages/web/src/db/adapters/WorkerWrappedAsyncDatabaseConnection.ts
```diff
@@ -44,6 +48,39 @@ export class WorkerWrappedAsyncDatabaseConnection<Config extends ResolvedWebSQLO
...
+            // Hold the lock while the shared connection is in use.
+            await new Promise<void>((releaseLock) => {
+              this.lockAbortController.signal.addEventListener('abort', () => {
+                releaseLock();
+              });
+            });
+          }
+        )
+        // We aren't concerned with errors here
+        .catch(() => {})
```
simolus3, line 82: Wouldn't an error in `navigator.locks.request` (before a lock is obtained) cause the promise to never be fullfilled here? So I wonder if forwarding the error to the outer promise might be better?

simolus3 (approved):
LGTM apart from one question (which may just be a misunderstanding on my side).

packages/web/src/db/adapters/WorkerWrappedAsyncDatabaseConnection.ts
Chriztiaan: The .catch was meant to consume the `abort` error triggered from `close`. Beyond that I can't see any other reason for a throw, besides a bad lock name (which is in our control and we provide an identifier).

The more defensive approach would be to check the error for being an abort error, otherwise forward. But I am sure if we need it.

rkistner: I missed this the first time, but looking at this again: If the connection is closed before the lock was granted, the AbortError would be thrown there, caught and ignored, which will cause the `shareConnection` promise to never be resolved. In that case, should it rather resolve or reject the promise?

Chriztiaan: I think calling resolve is fine when it's the abort error.

simolus3 (dismissed):

PR #502, opened 2025-02-17
Support bucket priorities

simolus3:
Support bucket priorities by recognizing partial checkpoint complete messages and forwarding their status to the core extension.

This requires rolling version 0.3.10 of the core extension into this repository first.

### DIFF
--- a/packages/common/src/client/AbstractPowerSyncDatabase.ts
+++ b/packages/common/src/client/AbstractPowerSyncDatabase.ts
@@ -9,7 +9,7 @@ import {
   UpdateNotification,
   isBatchedUpdateNotification
 } from '../db/DBAdapter.js';
-import { SyncStatus } from '../db/crud/SyncStatus.js';
+import { SyncPriorityStatus, SyncStatus } from '../db/crud/SyncStatus.js';
 import { UploadQueueStats } from '../db/crud/UploadQueueStatus.js';
 import { Schema } from '../db/schema/Schema.js';
 import { BaseObserver } from '../utils/BaseObserver.js';
@@ -343,14 +343,31 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
   }
 
   protected async updateHasSynced() {
-    const result = await this.database.get<{ synced_at: string | null }>(
-      'SELECT powersync_last_synced_at() as synced_at'
+    const result = await this.database.getAll<{ priority: number; last_synced_at: string }>(
+      'SELECT priority, last_synced_at FROM ps_sync_state ORDER BY priority DESC'
     );
-    const hasSynced = result.synced_at != null;
-    const syncedAt = result.synced_at != null ? new Date(result.synced_at! + 'Z') : undefined;
+    let lastCompleteSync: Date | undefined;
+    const priorityStatus: SyncPriorityStatus[] = [];
 
-    if (hasSynced != this.currentStatus.hasSynced) {
-      this.currentStatus = new SyncStatus({ ...this.currentStatus.toJSON(), hasSynced, lastSyncedAt: syncedAt });
+    for (const { priority, last_synced_at } of result) {
+      const parsedDate = new Date(last_synced_at + 'Z');
+
+      if (priority === 2147483647) {
+        // This lowest-possible priority represents a complete sync.
+        lastCompleteSync = parsedDate;
+      } else {
+        priorityStatus.push({ priority, hasSynced: true, lastSyncedAt: parsedDate });
+      }
+    }
+
+    const hasSynced = lastCompleteSync != null;
+    if (hasSynced != this.currentStatus.hasSynced || priorityStatus != this.currentStatus.statusInPriority) {
+      this.currentStatus = new SyncStatus({
+        ...this.currentStatus.toJSON(),
+        hasSynced,
+        lastSyncedAt: lastCompleteSync,
+        statusInPriority: priorityStatus
+      });
       this.iterateListeners((l) => l.statusChanged?.(this.currentStatus));
     }
   }
--- a/packages/common/src/client/AbstractPowerSyncDatabase.ts
+++ b/packages/common/src/client/AbstractPowerSyncDatabase.ts
@@ -260,16 +260,30 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
   }
 
   /**
+   * Wait for the first sync operation to complete.
+   *
+   * @argument request Either an abort signal (after which the promise will complete regardless of
+   * whether a full sync was completed) or an object providing an abort signal and a priority target.
+   * When a priority target is set, the promise may complete when all buckets with the given (or higher)
+   * priorities have been synchronized. This can be earlier than a complete sync.
    * @returns A promise which will resolve once the first full sync has completed.
    */
-  async waitForFirstSync(signal?: AbortSignal): Promise<void> {
-    if (this.currentStatus.hasSynced) {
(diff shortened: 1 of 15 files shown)


packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -329,14 +343,31 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
...
+      if (priority === 2147483647) {
+        // This lowest-possible priority represents a complete sync.
+        lastCompleteSync = parsedDate;
+      } else {
+        priorityStatus.push({ priority, hasSynced: true, lastSyncedAt: parsedDate });
+      }
+    }
+
+    const hasSynced = lastCompleteSync != null;
+    if (hasSynced != this.currentStatus.hasSynced || priorityStatus != this.currentStatus.statusInPriority) {
```
stevensJourney, line 364: The check here for `priorityStatus != this.currentStatus.statusInPriority` is doing a comparison on the Array references which seems potentially incorrect?

Maybe this is fine since we only execute `updateHasSynced` during the initialisation of the client? If so we should probably add a comment explaining this behaviour.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -453,6 +466,20 @@ The next upload iteration will be delayed.`);
     this.updateSyncStatus({ connected: false, connecting: false });
   }
 
+  private async collectLocalBucketState(): Promise<[BucketRequest[], Map<string, BucketDescription | null>]> {
+    const bucketEntries = await this.options.adapter.getBucketStates();
+    const req: BucketRequest[] = Array.from(bucketEntries).map((entry) => ({
```
stevensJourney, line 471: Nit: Is the `Array.from` necessary here? It looks like `bucketEntries` is already an Array

packages/common/src/db/crud/SyncStatus.ts
```diff
@@ -60,6 +67,45 @@ export class SyncStatus {
     );
   }
 
+  /**
+   * Partial sync status for involved bucket priorities.
+   */
+  get statusInPriority() {
```
stevensJourney, line 73: The name `statusInPriority` took me a few iterations to understand what it is. It looks like this returns the sorted statuses for all the priorities? I don't have a strong opinion, but maybe we could call this something along the lines of `priorityStatuses`?

packages/common/src/utils/DataStream.ts
```diff
@@ -99,10 +99,6 @@ export class DataStream<Data extends any = any> extends BaseObserver<DataStreamL
    * @returns a Data payload or Null if the stream closed.
    */
   async read(): Promise<Data | null> {
-    if (this.dataQueue.length <= this.lowWatermark) {
```
stevensJourney, line 102: For my interest sake. What was the reason for the changes here, how did this surface?

packages/web/tests/stream.test.ts
```diff
@@ -125,9 +40,6 @@ describe('Streaming', () => {
 
     // A new stream should be requested
     await newStream;
-
```
stevensJourney, line 128: I've made similar changes to the `createConnectedDatabase` calls [here](https://github.com/powersync-ja/powersync-js/pull/497/files#diff-0ac88b67c58af29102456584e268f65a59b337f8e205b346fc5b94b2c85128ea). But I think we should be `disconnectAndClear`ing between tests still?

stevensJourney (commented):
Left some minor comments. I didn't review the specifics of the new sync protocol.

packages/web/tests/stream.test.ts
stevensJourney: Discussed offline. The disposal has been moved internally to the `generateConnectedDatabase` function using `onTestFinished`

packages/common/src/utils/DataStream.ts
simolus3: I don't remember the exact details anymore, but the old implementation had a deadlock that surfaced in the partial priority test that I've added (which may be stressing this a bit more than a real-world use-case because it sends new data directly after the old one has been received).

I think the issue was that `read()` was called and triggered the `lowWater` callback, which then completed before the `processQueue()` in the promise below here was reached. The `lowWater` completion didn't trigger an update before now listeners were added yet, and the `processQueue` in the promise didn't trigger anything because the returned data wasn't added to the queue.

packages/common/src/db/crud/SyncStatus.ts
```diff
@@ -3,12 +3,19 @@ export type SyncDataFlowStatus = Partial<{
...
+  hasSynced?: boolean;
+}
+
 export type SyncStatusOptions = {
   connected?: boolean;
   connecting?: boolean;
   dataFlow?: SyncDataFlowStatus;
   lastSyncedAt?: Date;
   hasSynced?: boolean;
+  prioritStatuses?: SyncPriorityStatus[];
```
stevensJourney, line 18: There's a typo in this field member name :)

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -146,6 +146,11 @@ export const isPowerSyncDatabaseOptionsWithSettings = (test: any): test is Power
   return typeof test == 'object' && isSQLOpenOptions(test.database);
 };
 
+/**
+ * The priority used by the core extension to indicate that a full sync was completed.
+ */
+const fullSyncPriority = 2147483647;
```
stevensJourney, line 152: Just a style nit. We typically name these const members like `FULL_SYNC_PRIORITY`

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -329,14 +348,34 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
   }
 
   protected async updateHasSynced() {
-    const result = await this.database.get<{ synced_at: string | null }>(
-      'SELECT powersync_last_synced_at() as synced_at'
+    const result = await this.database.getAll<{ priority: number; last_synced_at: string }>(
+      'SELECT priority, last_synced_at FROM ps_sync_state ORDER BY priority DESC'
```
stevensJourney, line 352: We should update the relevant SDK packages' SQLite dependency ranges with this PR (RNQS, OPSQLite and WASQLite). This query will fail if we don't use the latest version of `powersync-sqlite-core`.

packages/web/src/worker/db/open-worker-database.ts
```diff
@@ -13,34 +13,55 @@ export function openWorkerDatabasePort(
 ) {
   const needsDedicated = vfs == WASQLiteVFS.AccessHandlePoolVFS || vfs == WASQLiteVFS.OPFSCoopSyncVFS;
 
+  const handleErrors = <T extends AbstractWorker>(instance: T, url: string | URL): T => {
+    instance.onerror = (event) => {
+      throw `Unhandled worker error for ${workerIdentifier} at ${url}: ${event.message}: ${event.error}`;
+    };
+
+    return instance;
+  };
```
simolus3, line 22: I've added this to debug test failures (I suspected that something was going wrong with workers). I don't think that turned out to be the problem in the end, but IMO this is still a check worth having since it looks like we're not handling worker errors elsewhere (and not handling errors will cause timeouts too if we're waiting for messages that we'll never receive).

simolus3: Actuall it causes the next demo to break (for whatever reason). Reverting.

stevensJourney (approved):
I'm happy with the changes here :)

PR #503, opened 2025-02-19
feat: add barebones opsqlite demo

DominicGBauer:
## Description
Create a bare-bones react-native demo application that uses OP-SQLite for interacting with the SQLite db.

## Testing
Ios Screenshot:
image

Android Screenshot:
image


demos/react-native-barebones-opsqlite/metro.config.js
```diff
@@ -0,0 +1,49 @@
...
+];
+// #3 - Force resolving nested modules to the folders below
+config.resolver.disableHierarchicalLookup = true;
+config.resolver.unstable_enableSymlinks = true;
+
+// Needed to make `@powersync/web/umd` imports work
+config.resolver.unstable_enablePackageExports = true;
+
+config.resolver.resolveRequest = (context, moduleName, platform) => {
+  if (platform === 'web') {
```
stevensJourney, line 25: This demo doesn't seem to run on web? We should probably not clutter the barebones example with web setup if it's not being used.

stevensJourney:
Is there some requirement to commit the `ios` and `android` folders for this demo? Usually we don't require manual changes to those folders anymore. Some work has been done on other demos to remove some of the native project folders.

DominicGBauer:
Given this is bare bones and there isn't a framework I don't see how we would be able to do that? It's my understanding that using expo is what allows us to avoid adding those folders.

stevensJourney:
> Given this is bare bones and there isn't a framework I don't see how we would be able to do that? It's my understanding that using expo is what allows us to avoid adding those folders.

Okay that makes sense. I didn't realize the bare React Native library had that behaviour.

stevensJourney (approved):

Chriztiaan (approved):

Chriztiaan (approved):

stevensJourney (approved):

PR #505, opened 2025-02-20
Opfs uploads

stevensJourney:
# Overview

A bug was identified when using OPFS as the SQLite VFS. When reconnecting with different client parameters, the PowerSync client failed to trigger CRUD uploads, preventing data from being uploaded. This issue occurred only when multiple tab support was enabled (the default).

The root cause was the shared sync worker using a stale database connection after a reconnect. The sync manager's logic for obtaining a database connection has been updated to ensure a valid connection is always used when a client disconnects or reconnects.

Relevant unit tests were updated to include OPFS.

## TODOs
- [ ] Investigate weird formatting changes

### DIFF
--- a/.changeset/forty-moose-yell.md
+++ b/.changeset/forty-moose-yell.md
@@ -1,0 +1,5 @@
+---
+'@powersync/web': patch
+---
+
+Fixed bug where using OPFS and reconnecting would cause upload triggers to fail.
--- a/packages/web/src/worker/sync/SharedSyncImplementation.ts
+++ b/packages/web/src/worker/sync/SharedSyncImplementation.ts
@@ -71,10 +71,7 @@ export type WrappedSyncPort = {
 /**
  * @internal
  */
-export type RemoteOperationAbortController = {
-  controller: AbortController;
-  activePort: WrappedSyncPort;
-};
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   test:
     name: Test Packages
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-xl
```
stevensJourney, line 10: The web tests are much more stable with this faster runner. Tests often timeout on the standard runner. Increasing the timeout doesn't really help. This might be caused by a Vitest browser mode issue.

packages/web/vitest.config.ts
```diff
@@ -46,10 +46,10 @@ const config: UserConfigExport = {
       instances: [
         {
           browser: 'chromium'
-        },
-        {
-          browser: 'firefox'
         }
+        // {
```
stevensJourney, line 50: Firefox tests don't work on the Ubuntu XL runner. These tests were only recently introduced. Prioritising more stable Chrome tests for now.

Chriztiaan (approved):

PR #512, opened 2025-02-25
Adopt bucket priorities in yjs demo

simolus3:
This adopts bucket priorities in the yjs demo, where the effect of buckets with `priority: 0` is most obvious. In particular, this:

1. Splits the `global` buckets into one bucket for documents (with the default priority) and one for document updates (with priority 0).
2. Adds a small switch to the editor to disable uploading local changes (useful to build up a queue of local crud entries that would prevent sync operations from the server without priority 0).

Running the demo with updated sync rules requires a self-hosted sync service for bucket priorities to be supported. The effect of priority 0 can easily be seen by:

1. Opening the same document in two browsers.
2. Disabling local uploads in browser A, and making a few changes in that browser.
3. Starting to type in browser B.

Despite of the local crud queue in A, changes from B are synchronized and made visible :)


rkistner (approved):
This is great! Works much better now.

I updated the demo to use Supabase anonymous sign-ins, instead of custom auth Edge functions. This simplifies the setup quite a bit. I also removed some unused dependencies.

I wanted to remove the RPC function used for updates, instead just using Postgrest batch insert support. Unfortunately, doing base64 -> bytea conversion with Postgrest becomes complicated. I almost had it working using [custom domains](https://postgrest.org/en/v12/references/api/domain_representations.html#custom-domain), but the PowerSync service does not handle those correctly for syncing down yet.

Chriztiaan (approved):

PR #516, opened 2025-03-04
Add NodeJS implementation

simolus3:
This adds a NodeJS implementation based on the amazing work done by @spinda in [this fork](https://github.com/spinda/powersync-js/). I've followed up on that work by:

1. Offloading queries to workers, which allows concurrent readers.
2. Automatically downloading the prebuilt PowerSync core extension in an `install` script, and loading it when databases are opened (that seemed like the simplest option for now, bundling the shared libraries with the package would probably be a better approach).
3. Adding a small demo and beginning with a few simple tests.


packages/node/src/db/AsyncBetterSqlite.ts
```diff
@@ -0,0 +1,141 @@
...
+} else if (platform === 'darwin') {
+  extensionPath = 'libpowersync.dylib';
+}
+
+const loadExtension = (db: Database) => {
+  const resolved = url.fileURLToPath(new URL(`../${extensionPath}`, import.meta.url));
+  db.loadExtension(resolved, 'sqlite3_powersync_init');
+};
+
+function toComlink(port: MessagePort): Comlink.Endpoint {
```
stevensJourney, line 132: For my interest sake: why is this required?

packages/node/src/db/BetterSQLite3DBAdapter.ts
```diff
@@ -0,0 +1,329 @@
...
+      const worker = new Worker(new URL('./AsyncBetterSqlite.js', import.meta.url));
+      const listeners = new WeakMap<EventListenerOrEventListenerObject, (e: any) => void>();
+
+      const comlink = Comlink.wrap<BetterSqliteWorker>({
+        postMessage: worker.postMessage.bind(worker),
+        addEventListener: (type, listener) => {
+          let resolved: (event: any) => void =
+            'handleEvent' in listener ? listener.handleEvent.bind(listener) : listener;
+
+          // Comlink wants message events, but the message event on workers in Node returns the data only.
```
stevensJourney, line 64: Oh, I see this is required for NodeJS worker message ports to be compatible with Comlink? The Comlink [docs](https://github.com/GoogleChromeLabs/comlink/blob/main/docs/examples/06-node-example/main.mjs) mention that they might export a helper method for Node.js

stevensJourney: FWIW If Comlink is troublesome with Node.js it should be feasible to remove the dependency altogether.

packages/node/src/db/BetterSQLite3DBAdapter.ts
```diff
@@ -0,0 +1,329 @@
...
+        get: (query, params) => connection.get(query, params),
+        getAll: (query, params) => connection.getAll(query, params),
+        getOptional: (query, params) => connection.getOptional(query, params),
+        commit,
+        rollback
+      });
+      await commit();
+      return result;
+    } catch (ex) {
+      await rollback();
```
stevensJourney, line 228: In some cases it can be necessary to swallow errors for this rollback. For [example](https://github.com/powersync-ja/powersync-js/blob/197a37a5fd1a6d74600a495234b575ed4e629d5b/packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts#L284).

packages/node/src/db/BetterSQLite3DBAdapter.ts
```diff
@@ -0,0 +1,329 @@
...
+  async refreshSchema() {
+    await this.writeConnection.refreshSchema();
+
+    for (const readConnection of this.readConnections) {
+      await readConnection.refreshSchema();
+    }
+  }
+}
+
+class RemoteConnection implements BetterSQLite3LockContext {
```
stevensJourney, line 262: Nit: this file is quite large. It would be nice to split these classes into separate files.

stevensJourney (commented):
I made some very minor comments. Overall this looks great to me.

packages/node/src/db/BetterSQLite3DBAdapter.ts
simolus3: I couldn't find a way to import this helper module. It's part of the package they publish, but I couldn't import it (probably because they have a single main module in their `package.json`). So it was easier to effectively reimplement that helper.

> FWIW If Comlink is troublesome with Node.js it should be feasible to remove the dependency altogether.

Well, we do need the RPC channel to the background worker, and Comlink is quite convenient for that :D

packages/node/src/db/AsyncBetterSqlite.ts
simolus3: For this one I don't think it's required, the `.bind` essentially casts the function to `any`. So I've replaced it by just casting the `parentPort` to `Comlink.Endpoint` directly.

demos/example-node/README.md
```diff
@@ -0,0 +1,8 @@
+## NodeJS Demo
+
+This demonstrates a small NodeJS client opening a database and connecting PowerSync.
+
+To get started, set up a PowerSync service instance with the default rules for the todolist examples.
+Create a development token, and use it to implement `DemoConnector.fetchCredentials` in `main.ts`.
+
+Then, use `pnpm run run` to open a database, connect to PowerSync, wait for a first sync, and run a simple query.
```
stevensJourney, line 8: Could we rename the script to something like `start`? Then the command could be `pnpm run start` or `pnpm start`

demos/example-node/README.md
```diff
@@ -0,0 +1,8 @@
+## NodeJS Demo
+
+This demonstrates a small NodeJS client opening a database and connecting PowerSync.
+
+To get started, set up a PowerSync service instance with the default rules for the todolist examples.
```
benitav, line 5: For someone new to PowerSync this might not be clear enough. Are rules referring to Sync Rules? As I understand this they need to basically follow what we covered in this section: https://docs.powersync.com/integration-guides/supabase-+-powersync#configuring-powersync - is that correct? We could then link to that.

demos/example-node/README.md
```diff
@@ -0,0 +1,8 @@
...
+Create a development token, and use it to implement `DemoConnector.fetchCredentials` in `main.ts`.
```
benitav, line 6:
```suggestion
Create a [development token](https://docs.powersync.com/installation/authentication-setup/development-tokens#development-tokens), and use it to implement `DemoConnector.fetchCredentials` in `main.ts`.
```

packages/node/README.md
```diff
@@ -0,0 +1,39 @@
...
+```bash
+npm install @powersync/node
+```
+
+Both `@powersync/node` and the `better-sqlite3` packages have install scripts that need to run to compile
+or download sqlite3 and PowerSync binaries.
+
+# Getting Started
+
+You can follow along our [web SDK reference](https://docs.powersync.com/client-sdk-references/js-web) for the web
```
benitav, line 27:
```suggestion
You can follow along our [web SDK reference](https://docs.powersync.com/client-sdk-references/js-web)
```

demos/example-node/README.md
simolus3: I've added the link 👍

simolus3:
I ran the web tests locally, they still pass.

benitav (approved):

packages/node/tests/PowerSyncDatabase.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  await database.execute('INSERT INTO todos (id, content) VALUES (uuid(), ?)', ['fourth']);
+  await expect.poll(() => fn).toHaveBeenCalledTimes(2);
+});
+
+databaseTest.skip('can watch queries', async ({ database }) => {
+  const query = await database.watch('SELECT * FROM todos;', [])[Symbol.asyncIterator]();
+  expect((await query.next()).value.rows).toHaveLength(0);
+
+  await database.execute('INSERT INTO todos (id, content) VALUES (uuid(), ?)', ['first']);
+  // TODO: There is a race condition somewhere, this reports now rows sometimes.
```
stevensJourney, line 62: Is this still the case?

simolus3: I can still reproduce it. I've tried investigating it but I don't see what we're doing wrong in the Node SDK (I initially suspected a leaking statement keeping an earlier transaction open but couldn't find anything looking in the `sqlite_stmt` VTAB).

I wonder if this might be another variant of https://github.com/powersync-ja/powersync-js/pull/525.

stevensJourney: I see the `onChange` callback test above does pass, it might be a race condition in the async iterator implementation. I don't think many people use the async iterator version of this API, but we should properly investigate at some point. For the initial alpha release this can be a known issue.

stevensJourney (dismissed):

PR #519, opened 2025-03-07
Updated demo's RNQS version and changed SyncStatus.ts to not depend on `toSorted`.

Chriztiaan:
The demos didn't correctly resolve to RNQS 2.4.1 on my side despite the PowerSync packages using them, at least when using `pnpm` on my side.

For `toSorted` in SyncStatus, it isn't available in React-Native without a polyfill, so I just replaced the shorthand with a similar implementation that is supported.


stevensJourney (approved):
LGTM

PR #528, opened 2025-03-11
Fix uploadData calling fetchCredentials when token has not expired

Omnicpie:
Currently, as part of the flow to build a request for uploadData, we call `getCredentials()` within AbstractRemote. This getCredentials function does not check / use the currently stored set of credentials unless there is a time in which the developer provides `expiresAt` as part of their `fetchCredentials()` response, Since this is not included in the typing, its relatively unlikely that this would be the case, so for most users we will be making excess calls to fetchCredentials when the current set of stored creds is still valid. 

This PR expands the `getCredentials()` request to get the current token stored, decode the JWT data to get the current `exp` field (if available), and perform the same check that would happen for a manually provided `expiresAt` field. 

### Testing

I tested this using the `example-vite` demo with some minor alterations, along with a self-hosted powersync instance. 

Below is a code snippet of the main changes to the example, mostly just giving a valid token and endpoint to the return of `fetchCredentials`. There was also an addition of a button to trigger an insert.
 
image

#### Before changes

Before the change showed the issue well, when we called uploadData (via pressing the button to insert an action), we could see that it would log `insert` for the `uploadData` method, but also that it would log `fetch` for the `fetchCredentials` method:

Screenshot 2025-03-12 at 09 29 50

#### After changes

Post changes, when calling the endpoint with a valid token, we could see that `insert` was still logged, but the `fetch` log was no longer present, due to `exp` on the token still being valid.

Screenshot 2025-03-12 at 09 36 48


Chriztiaan:
Hey @Omnicpie, I think that while this PR does work - it's possible that we might want to approach it slightly differently.
If consider what we do in our Dart SDK ([example 1](https://github.com/powersync-ja/powersync.dart/blob/main/packages/powersync_core/lib/src/streaming_sync.dart#L552-L555)), we interrogate the status when setting up the stream and invalidate the credentials on a `401`. It looks like we do this check/invalidate in a few places ([streaming_sync.dart](https://github.com/powersync-ja/powersync.dart/blob/main/packages/powersync_core/lib/src/streaming_sync.dart) and [connector.dart](https://github.com/powersync-ja/powersync.dart/blob/101c0b8f242ab19979c2f382cdd0eb70244d875f/packages/powersync_core/lib/src/connector.dart)).
This results in the same behaviour as what you have here.

You are welcome to update this PR to match the Dart implement, it might be a bit of work though. Otherwise we do have this work in our backlog and will get to this in the near future.

Omnicpie:
hey @Chriztiaan, if I'm understanding those dart implementations correctly, does that mean that we always assume that credentials are OK if they are present (in dart land), and only in the case that we have 401'd and wiped the creds, then we generate new ones on the next request?

if that is the case, what 401 would that be? Is it only the 401 from the sync service itself, or would an error from the uploadData function also trigger that functionality?

Chriztiaan:
> hey @Chriztiaan, if I'm understanding those dart implementations correctly, does that mean that we always assume that credentials are OK if they are present (in dart land), and only in the case that we have 401'd and wiped the creds, then we generate new ones on the next request?

Yes, we depend on the 401 to determine whether they are invalid or not.

> if that is the case, what 401 would that be? Is it only the 401 from the sync service itself, or would an error from the uploadData function also trigger that functionality?

There's a deprecated connector in dart that actually does do invalidate on a 401. But from our side we are concerned about the powersync stream 401s. Users would normally manage their own session invalidation for uploads - but it could be the same, just that we don't enforce it. One possible challenge is detecting this status code for websockets. I haven't looked into this as of yet.

Chriztiaan:
Just an update on this @Omnicpie, I am working on a new PR for this issue.

Omnicpie:
functionality completed in #604

PR #532, opened 2025-03-13
Compile node package as CommonJS module

simolus3:
In the `@powersync/node` package, this adds `rollup` to bundle the ESM modules emitted by `tsc` modules into a CommonJS module. Two places in the package are using `import.meta.url` which I had to replace since the CJS bundle is at a different path. We use a custom entry in `import.meta` (also set by rollup) to detect when the package has been compiled to a CJS module.
As a fallback (e.g. when users are applying their own bundler that might not set the `import.meta` entry), I've also added options to customize how the worker is loaded.

I've used this file to test this:

```JavaScript
// demos/example-node/src/test.cjs
const { PowerSyncDatabase, Schema, Table, column } = require('@powersync/node');

const lists = new Table({
    created_at: column.text,
    name: column.text,
    owner_id: column.text
  });

const main = async () => {
  const db = new PowerSyncDatabase({
    schema: new Schema({lists}),
    database: {
      dbFilename: 'test.db'
    }
  });

  console.log(await db.get('SELECT powersync_rs_version();'));
  await db.close();
};

main();
```


packages/node/src/db/options.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * The {@link SQLOpenOptions} available across all PowerSync SDKs for JavaScript extended with
+ * Node.JS-specific options.
+ */
+export interface NodeSQLOpenOptions extends SQLOpenOptions {
+  /**
+   * The Node.JS SDK will use one worker to run writing queries and additional workers to run reads.
+   * This option controls how many workers to use for reads.
+   */
+  readWorkers?: number;
```
stevensJourney, line 13: Just a nit: Maybe this could be called `readWorkerCount` or `numReadWorkers`? It could just be me, but just reading `readWorkers` seems like it could be a boolean setting (ofcoarse this is if I disregard the typing).

packages/node/src/db/SqliteWorker.ts
```diff
@@ -96,21 +103,43 @@ class BetterSqliteWorker implements AsyncDatabaseOpener {
...
+  extensionPath: () => string;
+}
 
-  const resolved = url.fileURLToPath(new URL(`../${extensionPath}`, import.meta.url));
-  db.loadExtension(resolved, 'sqlite3_powersync_init');
-};
+export function startPowerSyncWorker(options?: Partial<PowerSyncWorkerOptions>) {
+  const resolvedOptions: PowerSyncWorkerOptions = {
+    ...options,
+    extensionPath() {
```
stevensJourney, line 118: Is this always going to override the `extensionPath` method potentially provided by `options`?

packages/node/src/db/SqliteWorker.ts
```diff
@@ -96,21 +103,43 @@ class BetterSqliteWorker implements AsyncDatabaseOpener {
...
+      const isCommonJsModule = import.meta.isBundlingToCommonJs ?? false;
```
stevensJourney, line 119: One potential alternative might to check if `typeof __dirname == 'undefined'` during runtime, this should be false for CJS and true for ESM. Although there might be weird cases where users might globally overwrite `__dirname` (but I can't really think of a good reason to do this).

Another alternative might also be to just write the code for the ESM use cases and have Rollup handle the `import.meta.url` calls by replacing it (maybe using the replace plugin) with `'import.meta.url': 'require("url").pathToFileURL(__dirname).href',`

But I do see the complexities with handling `import.meta.url` here. I'd be happy with the current implementation.

simolus3: > One potential alternative might to check if `typeof __dirname == 'undefined'` during runtime, this should be false for CJS and true for ESM

I'm not sure if vitest is applying something to change that, but the tests seem to fail when I use that check.

> calls by replacing it (maybe using the replace plugin) with `'import.meta.url': 'require("url").pathToFileURL(__dirname).href'`

Another issue is that the path of the worker is different depending on whether we use CommonJS or ESM (`DefaultWorker.cjs` vs `DefaultWorker.js`). So given that this check works by default I'm leaning towards leaving it as-is.

stevensJourney: 👍  That makes sense.

packages/node/tests/PowerSyncDatabase.test.ts
```diff
@@ -1,5 +1,43 @@
...
+        readWorkerCount: 0,
+      }
+    });
+    await database.init();
+  }).rejects.toThrowError('Needs at least one worker for reads');
+});
+
+test('can customize loading workers', async () => {
+  const directory = await createTempDir();
+  const openFunction = vi.fn((...args: ConstructorParameters<typeof Worker>) => new Worker(...args));
```
stevensJourney, line 23: Might want to update this to the `OpenWorker` type

stevensJourney (approved):
Happy with these changes :)

stevensJourney (approved):

PR #533, opened 2025-03-17
Update README.md

benitav:
Link to docs we have available for Node.js


simolus3 (approved):
Oops 😬 Thanks for fixing the link!

PR #545, opened 2025-03-26
Introduced functionality for releasing the navigator lock

Chriztiaan:
In some cases for web with multiple tabs support enabled, multiple 'connect()` calls in a single tab happening in succession can cause a navigator lock to be stuck in an unreleased state. This blocks the sync worker from getting a connection and syncing/connecting breaks. 

Two typical scenarios that can cause this is switching client parameters and React's StrictMode which does multiple calls of hooks like useEffect.

[Cherrypicked commit](https://github.com/powersync-ja/powersync-js/compare/main...hack/sleep-locks) from @stevensJourney.


packages/web/src/worker/sync/SharedSyncImplementation.ts
```diff
@@ -259,6 +259,7 @@ export class SharedSyncImplementation
       }
 
       // Clearing the adapter will result in a new one being opened in connect
+      await trackedPort.clientProvider.releaseSharedConnection();
```
simolus3, line 262: For my own understanding: Where would we release the lock otherwise? Through `trackedPort.db.close()`?

I don't have a good mental model of how we're sharing connections here, but I'm worried about whether it's possible for us to release the shared connection here and someone else then re-obtaining the lock on a different connection? (so we'd have two of them when it sounds like we probably only want to have one).

If the lock is only held to avoid the tab being put to sleep as the comment suggests, why aren't we just requesting a randomly-named lock?

Chriztiaan:
Closed in favour of https://github.com/powersync-ja/powersync-js/pull/546

PR #549, opened 2025-03-31
Node.js install error documentation

stevensJourney:
# Overview

Some machine configurations might case the `install` script to fail when building the Better SQLite3 Node Addon module. 

This can most likely occur due to multiple versions of `node-gyp` being present in a project. 

This PR adds some documentation for common install issues.

## Demos

We currently have multiple versions of `node-gyp` present in this monorepo. The Angular demo has been updated to use Angular 19 - which uses a newer version of `node-gyp`.

image

The Electron demos still use `node-gyp` `9.4.1`. These demos have not been updated.

Running `node-gyp -v` on the main branch in the repo root produces `v9.4.1` while running on this branch produces `v11.1.0`


Chriztiaan (approved):

simolus3 (approved):
Thanks for investigating this 👍

PR #550, opened 2025-03-31
feat: Exposing upload and download errors in SyncStatus

Chriztiaan:
Added extra information fields to SyncStatus, similar to dart API.


stevensJourney (commented):
Overall this looks good to me so far. 

It should be possible add unit tests, at least for the upload error reporting.

Chriztiaan:
> Overall this looks good to me so far.
> 
> It should be possible add unit tests, at least for the upload error reporting.

@stevensJourney, I've added 2 simple sync status tests which are just a minor extensions of the existing stream tests we have. They are under web for the moment, but I think we can do some work/mocking in the future to be able to move these kinds of tests to `common`.

stevensJourney (approved):

PR #555, opened 2025-04-02
Sync progress

simolus3:
This implements the sync progress API for the JavaScript SDKs. This is currently work in progress as it's waiting for the releevant core extensions to be released.

The overall API is very similar to the one exposed by the other SDKs:

1. `SyncDataFlowStatus`, which needs to be transferrable from workers, adds a new `InternalProgressInformation` field tracking progress across different buckets. Users are encouraged not to use that directly though.
2. Instead, a new `downloadProgress` getter is added to `SyncStatus` with the same interface we expose for Dart and Kotlin: An `untilCompletion` getter and a `untilPriority(priority)` method. They both return a structure making the total and downloaded operations available (as well as a relative progress indicator).

I've added unit tests for this through the Node SDK, because that's easier to test with a local core extension build than the Web / RN SDKs that need subsequent builds. I think we should eventually share tests between the SDKs to be sure we're not missing anything, but I think this already points towards the implementation working :)


.changeset/sharp-singers-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/common': minor
+---
+
+Report progress information about downloaded rows.
```
stevensJourney, line 5: It's not technically required, but it's sometimes nice to explicitly include the SDK packages as part of this changeset. The changeset message will directly be mentioned in the relevant SDK changelogs then.

packages/common/src/client/sync/bucket/BucketStorageAdapter.ts
```diff
@@ -30,6 +30,14 @@ export interface SyncLocalDatabaseResult {
   checkpointFailures?: string[];
 }
 
+export type BucketOperationProgress = Record<
+  string,
+  {
```
stevensJourney, line 35: Nit: It would be nice to export this as its own type then use that in `BucketOperationProgress`

packages/common/src/client/sync/bucket/SqliteBucketStorage.ts
```diff
@@ -91,6 +92,13 @@ export class SqliteBucketStorage extends BaseObserver<BucketStorageListener> imp
     return result;
   }
 
+  async getBucketOperationProgress(): Promise<BucketOperationProgress> {
+    const rows = await this.db.getAll<{ name: string; count_at_last: number; count_since_last: number }>(
+      'SELECT name, count_at_last, count_since_last FROM ps_buckets'
+    );
+    return Object.fromEntries(rows.map((r) => [r.name, { atLast: r.count_at_last, sinceLast: r.count_since_last }]));
```
stevensJourney, line 99: I like the use of `Object.fromEntries` here. It might not be warranted, but it would be good to confirm that this works in React Native's JS engine.  `fromEntries` seems to have been available for a long time, but I just get suspicious sometimes about compatibility - React native sometimes requires polyfills for the most common functions.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -164,21 +165,23 @@ export abstract class AbstractStreamingSyncImplementation
...
   syncStatus: SyncStatus;
+  private syncStatusOptions: SyncStatusOptions;
   triggerCrudUpload: () => void;
 
   constructor(options: AbstractStreamingSyncImplementationOptions) {
     super();
     this.options = { ...DEFAULT_STREAMING_SYNC_OPTIONS, ...options };
 
-    this.syncStatus = new SyncStatus({
+    this.syncStatusOptions = {
```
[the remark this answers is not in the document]
stevensJourney: Thanks for the detailed response. That makes sense. I'm happy with this approach, I think maybe just add a comment in the code explaining the above.

If I understand this correctly, another Approach might be to extend the `SyncStatus` class with an internal implementation which has access and tracks these internal sync progress values. The public facing API could still be restricted to exposing the `SyncStatus` class.

But either way, I'm happy with this here if we add a comment.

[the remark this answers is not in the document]
simolus3: I just realized that `SyncStatus` does actually expose the internal counts, through `downloadFlow.downloadProgress`. I think this is required because the shared web worker sync implementations have to use `toJson()` on a `SyncStatus` to serialize it over ports. Using that here as well simplifies the implementaiton.

stevensJourney (approved):
I'm quite happy with the changes here.

stevensJourney (approved):

PR #558, opened 2025-04-03
Fix checkpoints during uploads not being applied

simolus3:
Under our consistency guarantee, we can't make data from new checkpoints visible if there are local changes that haven't been uploaded yet. Our current implementation for this case (recognizable by `ready = false`) was to simply do nothing. That's not unreasonable: After all, the pending local changes are supposed to be uploaded eventually, which should trigger new checkpoint messages that we'd then apply.

However, there is a potential race condition when a pending upload is currently in progress. While the connector is uploading changes (and before finishing), some data may already be uploaded to the database, triggering the sync service to send new checkpoints. These checkpoints would be entirely ignored, because we're already uploading data and there is no subsequent upload to trigger another one.

To fix that issue, this PR improves the checkpoint handling logic. If we get a checkpoint complete message that didn't apply because of local data or a `write_checkpoint` mismatch, _and_ we're currently in the process of uploading changes or waiting for a `write-checkpoint2.json` request, then await that before trying the same checkpoint again. Either,

- the same checkpoint message applies now (in case it was emitted before the connector completes or the `write-checkpoint2.json` returns). Or,
- we can expect another checkpoint to come up soon (since we've waited for the response and the old one didn't work). Doing nothing is the correct behavior here.

### DIFF
--- a/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
+++ b/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
@@ -1,9 +1,9 @@
 import Logger, { ILogger } from 'js-logger';
 
-import { SyncPriorityStatus, SyncStatus, SyncStatusOptions } from '../../../db/crud/SyncStatus.js';
+import { SyncStatus, SyncStatusOptions } from '../../../db/crud/SyncStatus.js';
 import { AbortOperation } from '../../../utils/AbortOperation.js';
 import { BaseListener, BaseObserver, Disposable } from '../../../utils/BaseObserver.js';
-import { throttleLeadingTrailing } from '../../../utils/throttle.js';
+import { onAbortPromise, throttleLeadingTrailing } from '../../../utils/async.js';
 import { BucketChecksum, BucketDescription, BucketStorageAdapter, Checkpoint } from '../bucket/BucketStorageAdapter.js';
 import { CrudEntry } from '../bucket/CrudEntry.js';
 import { SyncDataBucket } from '../bucket/SyncDataBucket.js';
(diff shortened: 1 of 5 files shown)


packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -707,26 +697,13 @@ The next upload iteration will be delayed.`);
...
-              const result = await this.options.adapter.syncLocalDatabase(targetCheckpoint!);
-              if (!result.checkpointValid) {
-                // This means checksums failed. Start again with a new checkpoint.
-                // TODO: better back-off
-                await new Promise((resolve) => setTimeout(resolve, 50));
+              const result = await this.applyCheckpoint(targetCheckpoint!, signal);
+              if (result.endIteration) {
+                // TODO: Why is this one retry: false? That's the only change from when we receive
+                // the line above?
                 return { retry: false };
```
stevensJourney, line 704: Good catch. I think this was ported from the Dart SDK [here](https://github.com/powersync-ja/powersync.dart/blob/773cbba05ba242a31a97fe13daab23fd3eee17ad/packages/powersync_core/lib/src/streaming_sync.dart#L423). However it seems like we don't actually use that return value in [Dart](https://github.com/powersync-ja/powersync.dart/blob/773cbba05ba242a31a97fe13daab23fd3eee17ad/packages/powersync_core/lib/src/streaming_sync.dart#L150), we just always retry. We should probably do the same here.

simolus3: Given that that was the only place returning `retry: false`, I'll just remove the return value altogether and retry unconditionally.

stevensJourney: Thanks!

stevensJourney (approved):
Looks good to me! Thanks finding and fixing this!

Issue #540, opened 2025-03-21
add --input flag to powersync-web copy-assets command


jasonlewicki:
We're on a monorepo (nx), so our `node_modules` folder could be _(is)_ in a different location. I am using a custom script at the moment to copy those files so this doesnt affect me right now, but others may run into this.

Thanks!

PR #561, opened 2025-04-07
chore: Resolving `node_modules` via require.resolve for `copy-assets`

Chriztiaan:
Aims to resolve https://github.com/powersync-ja/powersync-js/issues/540.
In monorepos the node_modules directory is not always at `cwd`.

### DIFF
--- a/packages/web/bin/powersync.js
+++ b/packages/web/bin/powersync.js
@@ -11,14 +11,17 @@ program.name('powersync-web').description('CLI for PowerSync Web SDK utilities')
 program
   .command('copy-assets')
   .description('Copy assets to the specified output directory')
+  .option('-i, --input <directory>', 'node modules dependency directory', 'node_modules')
   .option('-o, --output <directory>', 'output directory for assets', 'public')
   .action(async (options) => {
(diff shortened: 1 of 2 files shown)


packages/web/bin/powersync.js
```diff
@@ -11,14 +11,17 @@ program.name('powersync-web').description('CLI for PowerSync Web SDK utilities')
...
   .action(async (options) => {
+    const inputDir = options.input;
     const outputDir = options.output;
 
+    console.log(`Input directory: ${inputDir}`);
     console.log(`Target directory: ${outputDir}`);
 
     const cwd = process.cwd();
-    const source = path.join(cwd, 'node_modules', '@powersync', 'web', 'dist');
+    const source = path.join(cwd, inputDir, '@powersync', 'web', 'dist');
```
stevensJourney, line 24: Is it feasible here to just use `require.resolve` to find the source directory?

Chriztiaan: Yea, works much better thanks!

stevensJourney (approved):

PR #565, opened 2025-04-08
feat: Logging cleanup in packages (web, node, attachments, react, vue)

Chriztiaan:
This is the first of 3 planned PRs for logging cleanup.

This PR sweeps some of our logging across multiple packages which either didn't respect log levels or were using `console.xxx()` to log information. The most noteworthy changes  were in `BroadcastLogger.ts`.

For the `attachments`, `vue`, and `react` packages the easiest way to access the logger instance was to expose it on the PowerSync database instance.


packages/web/src/db/sync/WebStreamingSyncImplementation.ts
```diff
@@ -26,7 +27,9 @@ export class WebStreamingSyncImplementation extends AbstractStreamingSyncImpleme
 
   obtainLock<T>(lockOptions: LockOptions<T>): Promise<T> {
     const identifier = `streaming-sync-${lockOptions.type}-${this.webOptions.identifier}`;
-    lockOptions.type == LockType.SYNC && console.debug('requesting lock for ', identifier);
+    if (lockOptions.type == LockType.SYNC) {
+      (this.logger ?? Logger.get('WebStreamingSyncImplementation')).debug('requesting lock for ', identifier);
```
stevensJourney, line 31: It might be nicer to set `this.logger` in the constructor to a Required ILogger.

packages/web/src/worker/sync/SharedSyncImplementation.worker.ts
```diff
@@ -20,7 +20,7 @@ _self.onconnect = function (event: MessageEvent<string>) {
   port.addEventListener('message', (event) => {
     const payload = event.data as ManualSharedSyncPayload;
     if (payload?.event == SharedSyncClientEvent.CLOSE_CLIENT) {
-      console.log('closing shared for port', port);
+      console.debug('closing shared for port', port);
```
stevensJourney, line 23: I imagine one could use the global Logger context for this? Either way: if it really is difficult to use a logger for this, then we can just remove it. This log is useful in some debugging scenarios, but we can always just add it again.

packages/web/src/worker/db/WASQLiteDB.worker.ts
```diff
@@ -6,8 +6,15 @@ import '@journeyapps/wa-sqlite';
...
 import { WASqliteConnection } from '../../db/adapters/wa-sqlite/WASQLiteConnection';
-import { ResolvedWASQLiteOpenFactoryOptions } from '../../db/adapters/wa-sqlite/WASQLiteOpenFactory';
+import {
+  ResolvedWASQLiteOpenFactoryOptions,
+  WorkerDBOpenerOptions
+} from '../../db/adapters/wa-sqlite/WASQLiteOpenFactory';
 import { getNavigatorLocks } from '../../shared/navigator';
+import Logger from 'js-logger';
+
+Logger.useDefaults();
```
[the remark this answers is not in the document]
Chriztiaan: Yes, just didn't want to stack the diffs.

packages/web/src/worker/sync/SharedSyncImplementation.worker.ts
Chriztiaan: Will remove then.The problem is that the global logger in this worker context is different to the global logger from the main context and as such this would ignore the log level set unless you can figure out a way to pass the level in on construction.

packages/web/src/db/sync/WebStreamingSyncImplementation.ts
Chriztiaan: Agree

Chriztiaan: Seems like this was already "definitely defined".

simolus3 (approved):
Probably best to let Steven take another look too, but these changes look good to me.

PR #570, opened 2025-04-17
feat: Logger API/Helpers

Chriztiaan:
A common pain point for logging is that users need to depend on js-logger to manage the logging for PowerSync.
This wraps and exports the logging into a simple utility available from common, web, and react-native.  

To ensure consumers of the web package no won't need the vite config for `js-logger`, `js-logger` is now bundled in `common`. This adds 3KB to the final web bundle. 

This bundling work, and the vite config not being needed, have been verified via a dev release of this branch. 

Will remove `js-logger` from the demos/tools, and update the packages in the next PR.

Example usage:
```
import { createBaseLogger, createLogger, LogLevel, PowerSyncDatabase } from '@powersync/web';

const defaultLogger = createBaseLogger();
defaultLogger.useDefaults();
defaultLogger.setLevel(LogLevels.OFF);

const myLogger = createLogger("my-little-logger");
myLogger.setLevel(LogLevels.ERROR);

  schema: AppSchema,
  database: {
    dbFilename: 'example.db'
  },
  logger: myLogger,
});
```

### DIFF
--- a/packages/common/src/utils/Logger.ts
+++ b/packages/common/src/utils/Logger.ts
@@ -1,0 +1,32 @@
+import Logger, { type ILogger, type ILogLevel } from 'js-logger';
+
+export { GlobalLogger, ILogger, ILoggerOpts, ILogHandler, ILogLevel } from 'js-logger';
+
+const TypedLogger: ILogger = Logger as any;
+
+export const LogLevel = {
+  TRACE: TypedLogger.TRACE,
+  DEBUG: TypedLogger.DEBUG,
+  INFO: TypedLogger.INFO,
+  TIME: TypedLogger.TIME,
+  WARN: TypedLogger.WARN,
+  ERROR: TypedLogger.ERROR,
+  OFF: TypedLogger.OFF
+};
+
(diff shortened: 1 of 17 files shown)


packages/common/src/utils/Logger.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export interface CreateLoggerOptions {
+  logLevel?: ILogLevel;
+}
+
+export function createBaseLogger() {
+  return Logger;
+}
+
+export function createLogger(name: string, options: CreateLoggerOptions = {}): ILogger {
```
stevensJourney, line 25: It would be nice to add comments/doc for these methods. It's not very clear from a glance what the difference between `createLogger` and `createBaseLogger` are.

stevensJourney (dismissed):

packages/common/src/utils/Logger.ts
Chriztiaan: Sorted

stevensJourney (approved):

PR #590, opened 2025-05-07
feat: Detox simulator test runner

Chriztiaan:
# Descriptions
This PR introduces an OP-SQLITE adapter test suite that works in a similar way to [RNQS](https://github.com/powersync-ja/react-native-quick-sqlite/blob/main/tests/tests/sqlite/rawQueries.spec.ts). This project's setup differs quite a bit though as it is Detox based, but still runs tests against a simulator. At the moment only the iOS simulator is being used.
 
The tests are pulled into a single Detox test, an interaction in the UI kicks off all the tests and we pull the state of each test out of the UI which then gets reported - any failures will cause the single Detox test to fail.

image

# Considerations
I am not sure whether this and future simulator/emulator Github actions should run on `PR` as they do take a bit longer than our other actions, it's good to have them regardless. They also seem to be nondeterministically unstable.

 # Future work
- Also run tests against the RNQS adapter, might need a second workflow so that you can do a pod install with a separate set of dependencies (without op-sqlite)

### DIFF
--- a/.changeset/cool-yaks-allow.md
+++ b/.changeset/cool-yaks-allow.md
@@ -1,5 +1,0 @@
----
-'@powersync/web': minor
----
-
-To support the upstream credentials management changes from `@powersync/common`, the sync worker now communicates credentials invalidation to tabs.
--- a/.changeset/green-buckets-talk.md
+++ b/.changeset/green-buckets-talk.md
@@ -1,5 +1,0 @@
----
-'@powersync/op-sqlite': patch
----
-
-Promoting package to Beta release.
--- a/.changeset/honest-melons-laugh.md
+++ b/.changeset/honest-melons-laugh.md
@@ -1,5 +1,0 @@
----
-'@powersync/web': patch
----
-
-Fixed issue where broadcast logger wasn't being passed to WebRemote, causing worker remote logs not to be broadcasted to the tab's logs.
--- a/.changeset/shiny-rules-invent.md
+++ b/.changeset/shiny-rules-invent.md
@@ -1,5 +1,0 @@
----
-'@powersync/common': minor
----
-
-Improved credentials management and error handling. Credentials are invalidated when they expire or become invalid based on responses from the PowerSync service. The frequency of credential fetching has been reduced as a result of this work.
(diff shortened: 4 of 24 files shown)


tools/powersynctests/.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
stevensJourney, line 1: We should probably keep the same prettier config from the repo root here.

tools/powersynctests/README.md
```diff
@@ -0,0 +1,97 @@
+This is a new [**React Native**](https://reactnative.dev) project, bootstrapped using [`@react-native-community/cli`](https://github.com/react-native-community/cli).
+
+# Getting Started
```
stevensJourney, line 3: We should probably replace this template README?

tools/powersynctests/src/tests/queries.test.ts
```diff
@@ -0,0 +1,662 @@
...
+        );
+      }
+      await db.execute('PRAGMA wal_checkpoint(RESTART)');
+      let end = performance.now();
+      let duration = end - start;
+
+      expect(duration).lessThan(2000);
+    });
+
+    //     // TODO: Behaviour differs between RNQS and OP-sqlite. Might need to fix this one in the OP-sqlite package
```
stevensJourney, line 624: After taking a quick look at the OPSqliteAdapter, we should be able to reject the promises in the `readQueue` when `close` is called. 

However this test tests if execute/writeLock's reject if the DB is closed. The `async-lock` lib we use for `writeLock ` queuing does not seem to accept an abort controller to cancel and reject write lock requests, but we might be able to keep a single `AbortController` on the adapter class which we abort when the `close` method is called. We could then register a listener in the Promise handler for the `writeLock` method which rejects that promise - if we set a default timeout for the async lock request and check if the abort controller has been aborted before executing the writeLock callback inside the lock callback- we should get better behaviour if the connection is closed.

stevensJourney (commented):
I left some minor comments. Overall this is pretty cool.
Android tests should be possible as discussed offline.

tools/powersynctests/src/tests/queries.test.ts
Chriztiaan: This should be handled now, the test also passes.

stevensJourney (approved):
Happy with these changes from my side.

stevensJourney (approved):

PR #591, opened 2025-05-08
poc: Expo Fetch

stevensJourney:
# Overview

We currently use `react-native-fetch-api` in the React Native SDK to perform `fetch` operations. This is required since the standard React Native `fetch` implementation does not support HTTP text streaming. 

Some users recently reported errors of the form 
```
[PowerSyncStream] Caught exception when uploading. Upload will retry after a delay. Exception: require(_dependencyMap[11], "rea(...)/BlobManager").createFromOptions is not a function (it is undefined)
```
Which most likely is caused by `react-native-fetch-api`.

Expo 52 introduced their own [implementation](https://docs.expo.dev/versions/latest/sdk/expo/#expofetch-api) for `fetch` which supports HTTP streaming. 

This PR attempts to require and use Expo's fetch implementation (if available) - we default to `react-native-fetch-api` if Expo >= 52 is not installed.

### DIFF
--- a/.changeset/many-fishes-camp.md
+++ b/.changeset/many-fishes-camp.md
@@ -1,0 +1,5 @@
+---
+'@powersync/react-native': minor
+---
+
+Added ability to use Expo's fetch API if available.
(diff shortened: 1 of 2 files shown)


stevensJourney:
Closing this since the root issue was resolved in https://github.com/powersync-ja/powersync-js/pull/595

PR #599, opened 2025-05-13
Add sync implementation from core extension

simolus3:
This adds an option to use a new sync implementation from the core extension instead of the JavaScript implementation. The core implementation is marked as experimental, and we default to the old implementation.

The implementation changes are surprisingly compact:

1. In `AbstractRemote.ts`, this introduces "raw" variant of the sync stream methods. Those don't parse sync lines, instead returning them as strings (HTTP streaming) or buffers (RSocket).
2. In `AbstractStreamingSyncImplementation.ts`, copy the existing sync implementation into a method and add a new one for the core sync implementation.

See also:

- The implementation PR for the core extensions: github.com/powersync-ja/powersync-sqlite-core/pull/70
- The equivalent PR for the Kotlin SDK: https://github.com/powersync-ja/powersync-kotlin/pull/192


packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -519,229 +603,378 @@ The next upload iteration will be delayed.`);
...
+    async function connect(instr: EstablishSyncStream) {
+      const syncOptions: SyncStreamOptions = {
+        path: '/sync/stream',
+        abortSignal: abortController.signal,
+        data: instr.request
+      };
 
-        this.logger.debug('Stream established. Processing events');
+      if (resolvedOptions.connectionMethod == SyncStreamConnectionMethod.HTTP) {
+        controlInvocations = await remote.postStreamRaw(syncOptions, (line) => ['line_text', line]);
```
stevensJourney: This is a bit of a nit: It took me a little bit to realize what the different `map` function variants of `postStreamRaw` and `socketStreamRaw` were and why they were necessary. I think some additional comments describing what `line_text` and `line_binary` are would help make this a little bit easier to read. Perhaps adding an Enum for the different Sync operations and changing the `control` method to receive an object parameter will also help (since the parameters are labeled when mapping)

```typescript
enum SyncOperation {
    PROCESS_TEXT_COMMAND = 'line_text', // these don't have to be the same
    PROCESS_BINARY_COMMAND = 'line_binary', 
    STOP = 'stop',
    START = 'start'
}

controlInvocations = await remote.postStreamRaw(syncOptions, (line) => ({operation: SyncOperation.PROCESS_TEXT_COMMAND, payload: line}));
//...
    const controlEvent = await controlInvocations.read();
    await control(controlEvent)
```

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -188,6 +246,7 @@ export abstract class AbstractStreamingSyncImplementation
       }
 
       this.pendingCrudUpload = new Promise((resolve) => {
+        this.notifyCompletedUploads?.();
```
stevensJourney: Should we be doing the notification before the following upload has completed?

stevensJourney:
Overall I think the work done here is pretty cool and well documented. I left some minor comments, but this looks good to me.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
simolus3: Nope, good catch!

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
simolus3: I've added an enum and an object (instead of the array) for this now, let me know if that makes it better.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -519,229 +609,392 @@ The next upload iteration will be delayed.`);
...
+    try {
+      await control(
+        PowerSyncControlCommand.START,
+        JSON.stringify({
+          parameters: resolvedOptions.params
+        })
+      );
+
+      this.notifyCompletedUploads = () => {
+        controlInvocations?.enqueueData({ command: PowerSyncControlCommand.NOTIFY_CRUD_UPLOAD_COMPLETED });
```
stevensJourney, line 996: Note that the `enqueueData` call here and when handling the `FetchCredentials` instruction might need to be wrapped in a `try...catch` since the `enqueueData` method can throw an exception if trying to enqueue data after the DataStream has been closed.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
stevensJourney: Thanks! Looks good to me :D

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
simolus3: I'm setting `controlInvocations` back to null before closing it now, that should cover these.

PR #604, opened 2025-05-16
feat: Fetch credentials on token expiry/401

Chriztiaan:
Improved credentials management and error handling. Credentials are invalidated when they expire or become invalid based on responses from the PowerSync service. The frequency of credential fetching has been reduced as a result of this work.

For web we need to communicate over the comlink from the worker to the main tab that invalidation needs to happen.

Follows a similar pattern applied to PowerSync Dart.

Alternative approach to https://github.com/powersync-ja/powersync-js/pull/528

### DIFF
--- a/packages/common/src/client/sync/stream/AbstractRemote.ts
+++ b/packages/common/src/client/sync/stream/AbstractRemote.ts
@@ -396,9 +396,17 @@ export abstract class AbstractRemote {
         syncQueueRequestSize, // The initial N amount
         {
           onError: (e) => {
-            if (e.message.includes('Authorization failed') || e.message.includes('PSYNC_S21')) {
-              this.invalidateCredentials();
+            if (e.message.includes('PSYNC_')) {
+              if (e.message.includes('PSYNC_S21')) {
+                this.invalidateCredentials();
+              }
+            } else {
+              // Possible that connection is with an older service, always invalidate to be safe
+              if (e.message !== 'Closed. ') {
+                this.invalidateCredentials();
+              }
             }
+
             // Don't log closed as an error
             if (e.message !== 'Closed. ') {
               this.logger.error(e);
--- a/packages/common/src/client/sync/stream/AbstractRemote.ts
+++ b/packages/common/src/client/sync/stream/AbstractRemote.ts
@@ -163,8 +163,6 @@ export abstract class AbstractRemote {
    *
    * This should always fetch a fresh set of credentials - don't use cached
    * values.
-   *
-   *
    */
   async fetchCredentials() {
     const credentials = await this.connector.fetchCredentials();
@@ -223,6 +221,10 @@ export abstract class AbstractRemote {
       body: JSON.stringify(data)
     });
 
+    if (res.status === 401) {
+      this.invalidateCredentials();
+    }
+
     if (!res.ok) {
       throw new Error(`Received ${res.status} - ${res.statusText} when posting to ${path}: ${await res.text()}}`);
     }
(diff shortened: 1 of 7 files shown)


packages/common/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -335,6 +396,9 @@ export abstract class AbstractRemote {
         syncQueueRequestSize, // The initial N amount
         {
           onError: (e) => {
+            if (e.message.includes('Authorization failed')) {
+              this.invalidateCredentials();
+            }
```
rkistner, line 401: This is likely to miss some authorization failures, either now or in the future. For example, connecting with an expired token gives me `[PSYNC_S2106] Authentication required`.

The ideal would be for the sync protocol to include the 401 status in the error - not sure if that is happening currently.

As a workaround/fallback, you could check for messages containing `PSYNC_S21`.

Chriztiaan: Thanks for this! I wasn't sure whether it was fine to depend on the "PSYNC" codes.
We don't currently get the 401 status codes for the websocket connection method, but this should do for now.

rkistner: It's not great to rely on those codes, but it's better than checking the messages.

Chriztiaan:
I have had the dev release of this work tested by @Omnicpie. Feedback was positive.

packages/common/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -335,6 +396,9 @@ export abstract class AbstractRemote {
         syncQueueRequestSize, // The initial N amount
         {
           onError: (e) => {
+            if (e.message.includes('Authorization failed') || e.message.includes('PSYNC_S21')) {
```
stevensJourney, line 399: I think this kind of check becomes very tricky to cover all edge cases. Note that the `PSYNC` error codes were only introduced in PowerSync service version [1.3.7](https://github.com/powersync-ja/powersync-service/pull/182), I'm assuming the `Authorization failed` check should cover previous versions? 

Also, should this check also request new credentials for [PSYNC_22](https://github.com/powersync-ja/powersync-service/pull/182) errors?

Chriztiaan: I don't think so, see https://github.com/powersync-ja/powersync-service/blob/5b390393f3d03fc5422fbb00c1c5a4cd74b15c3a/packages/service-errors/src/codes.ts#L349-L373
Those are all config/integration based, which is less of a "session is expired" issue imo.

rkistner: Agreed with @Chriztiaan on the PSYNC_22 errors - those typically require intervention on the server to fix, and generating a new token won't make a difference. PSYNC_21 are specifically for auth errors due to something the client has control over.

Chriztiaan: @rkistner @stevensJourney It feels like the safest approach would be to change this check to something that definitely supports older services:
Check if we are getting modern error codes, if we do, evaluate. Otherwise, just always invalidate which means SDKs connecting to older services should still work and if they want the optimization they need a service bump.
```
if (e.message.includes('PSYNC_')) {
  if (e.message.includes('PSYNC_S21')) {
    this.invalidateCredentials();
  }
} else {
  // Possible that connection is with an older service, always invalidate to be safe
  this.invalidateCredentials();
}
```

stevensJourney (approved):
Looks good to me

PR #607, opened 2025-05-22
Improve Connection Locks

stevensJourney:
# Overview

**TL;DR:** Improves connection performance when performing multiple calls to `connect` in rapid succession. 

We currently protect the logic in `connect` calls with an exclusive mutex/lock. This mutex causes invocations to `connect` to be queued.

Certain use-cases such as updating client parameters can result in rapid calls to `connect` when client parameters change. The mutex queuing can cause undesirable delays to this process. Each call to `connect` results in a connection attempt which needs to be awaited before proceeding to the next attempt. 

The work here retains a single connection attempt being active at a time - which prevents possible race conditions. However, **instead of queuing all requests sequentially, the system now buffers and consolidates multiple calls, always using the most recent connection parameters**.

# Implementation

## Key Changes

The connection logic has been moved to a `ConnectionManager` implementation which improves tracking pending asynchronous connect and disconnect operations. 

- **`pendingConnectionOptions`** - Buffers the latest connection parameters
- **`connectingPromise`** - Tracks active connection attempts to prevent races  
- **`disconnectingPromise`** - Coordinates disconnect operations
- **Smart consolidation logic** - Recursive connection checking for queued requests

## Connection Behavior

The timing of `connect` calls greatly affects the outcome of resultant internal `connect` operations. Calling `connect` periodically where the period is longer than the time required to connect will still result in a serial queue of connection attempts. 

Connect calls made with a period less than the time taken to connect or disconnect will be compacted into less requests. 

## Scenario 1: Rapid Successive Calls During Initialization

Calling `.connect` multiple times while the SyncStream implementation is initializating will result in the SyncStream implementation making a single request with the latest parameters. Subsequent calls to `connect`, after the SyncStream has been initialized, will result in a reconnect with the latest (potentially compacted) options. 

```mermaid
sequenceDiagram
    participant Client
    participant PowerSync as PowerSync Client
    participant Buffer as pendingConnectionOptions
    participant SyncStream
    
    Note over Client,SyncStream: SyncStream is initializing asynchronously
    
    Client->>PowerSync: connect(optionsA)
    PowerSync->>Buffer: store optionsA
    PowerSync->>SyncStream: start async initialization
    
    Client->>PowerSync: connect(optionsB)
    PowerSync->>Buffer: update to optionsB
    
    Client->>PowerSync: connect(optionsC)
    PowerSync->>Buffer: update to optionsC
    
    Note over SyncStream: Initialization completes
    SyncStream->>Buffer: get latest options
    Buffer-->>SyncStream: optionsC
    SyncStream->>SyncStream: connect with optionsC
    
    Note over Client,SyncStream: Result: Only optionsC is used, A & B are discarded
```

## Scenario 2: Updates During Active Disconnect

`disconnect` operations are shared internally. Calling `connect` multiple times re-uses a single disconnect operation internally. The latest options are buffered while the disconnect is in progress. 

``` mermaid
sequenceDiagram
    participant Client
    participant PowerSync as PowerSync Client  
    participant Buffer as pendingConnectionOptions
    participant SyncStream as Active SyncStream
    
    Note over SyncStream: SyncStream is connected and running
    
    Client->>PowerSync: connect(newOptionsA)
    PowerSync->>SyncStream: start disconnect()
    PowerSync->>Buffer: store newOptionsA
    
    Client->>PowerSync: connect(newOptionsB)
    PowerSync->>Buffer: update to newOptionsB
    
    Client->>PowerSync: connect(newOptionsC)  
    PowerSync->>Buffer: update to newOptionsC
    
    Note over SyncStream: Disconnect completes
    SyncStream-->>PowerSync: disconnected
    
    PowerSync->>Buffer: get latest options
    Buffer-->>PowerSync: newOptionsC
    PowerSync->>SyncStream: initialize & connect with newOptionsC
    
    Note over Client,SyncStream: Result: Seamless transition to latest options
```

## Edge Cases Handled

- **Concurrent multi-tab scenarios** - The initialization of the shared sync implementation is guarded by navigator locks in the web implementation of `runExclusive`
- **Connect during disconnect** - Proper serialization with promise coordination
- **Rapid parameter updates** - Latest-wins semantics ensure current user intent

# Performance Comparison

## Test Results: 21 Rapid `connect()` Calls

| Metric | Before (main) | After (this PR) | Improvement |
|--------|---------------|-----------------|-------------|
| **Total Time** | 2.5 seconds | 0.4 seconds | **6.25x faster** |
| **Sync Implementations Created** | 21 | 3 | **7x fewer resources** |
| **Connection Attempts** | 21 sequential | 3 consolidated | **Optimal efficiency** |

### Before (main branch)
image
*21 sync stream implementations created, each attempting connection before disposal*

### After (this PR)  
image
*Only 3 sync implementations created with intelligent parameter consolidation*

# Breaking Changes
None - this is a pure performance and efficiency improvement with identical public API behavior.

# Testing

This can be tested by using the following development packages.

```
@powersync/common@0.0.0-dev-20250529141956
@powersync/node@0.0.0-dev-20250529141956
@powersync/op-sqlite@0.0.0-dev-20250529141956
@powersync/react-native@0.0.0-dev-20250529141956
@powersync/tanstack-react-query@0.0.0-dev-20250529141956
@powersync/web@0.0.0-dev-20250529141956
```


packages/web/tests/utils/MockStreamOpenFactory.ts
```diff
@@ -81,7 +81,9 @@ export class MockRemote extends AbstractRemote {
         if (this.errorOnStreamStart) {
           controller.error(new Error('Mock error on stream start'));
         }
-
+        if (signal?.aborted) {
```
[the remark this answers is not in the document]
stevensJourney: Thanks!

obezzad:
Thanks a lot @stevensJourney. Very exciting work! Just tested the dev packages and wanted to share some preliminary observations: after expanding nodes, it seems that one stream keeps reconnecting every few seconds without any action from my end. Is this expected?

https://github.com/user-attachments/assets/2860ed8a-a822-4588-8159-334506fc3c28

packages/web/src/db/sync/SharedWebStreamingSyncImplementation.ts
```diff
@@ -187,9 +187,6 @@ export class SharedWebStreamingSyncImplementation extends WebStreamingSyncImplem
    */
   async connect(options?: PowerSyncConnectionOptions): Promise<void> {
     await this.waitForReady();
-    // This is needed since a new tab won't have any reference to the
```
stevensJourney, line 190: This is not needed anymore. Each connect internally is guaranteed to do a disconnect.

packages/common/src/client/ConnectionManager.ts
```diff
@@ -0,0 +1,234 @@
...
+        if (!this.pendingConnectionOptions) {
+          this.logger.debug('No pending connection options found, not creating sync stream implementation');
+          // A disconnect could have cleared this.
+          return;
+        }
+
+        if (this.disconnectingPromise) {
+          return;
+        }
+
```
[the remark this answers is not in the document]
stevensJourney: I'm not sure what the bot is mentioning here. There is a check for `this.pendingConnectionOptions` before destructuring.

packages/common/src/client/sync/stream/WebsocketClientTransport.ts
```diff
@@ -0,0 +1,70 @@
...
+      const errorListener = (ev: ErrorEvent) => {
+        removeListeners();
+        reject(ev.error);
+      };
+
+      /**
+       * In some cases, such as React Native iOS, the WebSocket connection may close immediately after opening
+       * without and error. In such cases, we need to handle the close event to reject the promise.
+       */
+      const closeListener = () => {
```
stevensJourney, line 54: This is a tag-along fix.

While regression testing I noticed some weird behaviour in React Native on iOS. 

Stopping my local PowerSync service while my iOS client was connected would result in the active connection aborting and retrying a connection as expected, but the first attempt after stopping the service would never resolve or reject. This effectively froze the sync implementation. From testing I noticed that the WebSocket would change from the `opening` state to the `closed` state without emitting any error. The standard implementation of `WebsocketClientTransport` only listens to the `open` and `close` events. This fork here now also listens to the `close` event and rejects this connection attempt - this allows the sync implementation to continue retrying connections.

Interestingly enough, only the first connection attempt (after closing the PowerSync service) has this behaviour. Subsequent connection attempts to emit and `error` event.

Chriztiaan (approved):

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -467,7 +474,9 @@ The next upload iteration will be delayed.`);
         });
 
         // On error, wait a little before retrying
-        await this.delayRetry();
+        if (delay) {
+          await this.delayRetry();
+        }
```
rkistner: Not a regression in this PR, but I've noticed two issues related to this retry delay in the diagnostics app (test with any endpoint you can't connect to):
1. `connected: false` isn't reported until the delay has passed. This means the `connect()` call Promise always has a 5-second delay if the connection fails, instead of failing immediately.
2. If you call `connect()` again within that 5 seconds, it waits until the 5 seconds are over before attempting another connection attempt, which is typically not expected. This is quite visible in the diagnostics app if we fix the point above.

The first one can be worked around by just reporting the `connected: false` status earlier, but not sure if that has further impacts on how the status is used.

For the retry delay, it should work to just end the delay early when disconnecting.

rkistner (approved):

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
stevensJourney: This should be fixed now. The delay is now abortable, which should resolve the connecting promise earlier when `disconnect` is called. 

I've updated and tested the Diagnostics app to ensure there are no delays when attempting a connection after a failed attempt.

tools/diagnostics-app/src/app/views/layout.tsx
```diff
@@ -199,6 +193,7 @@ export default function ViewsLayout({ children }: { children: React.ReactNode })
         </Drawer>
       </Box>
       <S.MainBox sx={{ flexGrow: 1, p: 3, width: { md: `calc(100% - ${drawerWidth}px)` }, marginTop: '50px' }}>
+        {syncError ? <Alert severity="error">Sync error detected: {syncError.message}</Alert> : null}
```
stevensJourney, line 196: Tag-along feature. This now alerts if there are sync errors on all the relevant views. E.g. if the provided token expires this banner will now show. 
image

packages/common/src/db/crud/SyncStatus.ts
```diff
@@ -171,7 +171,22 @@ export class SyncStatus {
...
+        return {
+          name: value.name,
+          message: value.message,
+          stack: value.stack
+        };
+      }
+      return value;
+    };
+
+    return JSON.stringify(this.options, replacer) == JSON.stringify(status.options, replacer);
```
stevensJourney, line 189: Tag-along fix: While testing the Diagnostics app I noticed that changes to download Errors were not propagated. The standard `JSON.stringify` serializes the `downloadError` `Error` instance to `'{}`. This causes the `isEqual` test to falsely identify changes in the Error as no change. 

Before:
image

After:
image

PR #614, opened 2025-05-30
Incremental Watched Queries

stevensJourney:
# Overview

Our current Watched query implementations emit results whenever a change to a dependant SQLite table occurs. The table changes might not affect the query result set, but we still query and emit a new result set for each table change. The result sets typically contain the same data, but these results are new Array/object references which will cause re-renders in certain frameworks like React.

This PR overhauls, improves and extends upon the existing watched query implementations by introducing incremental watched queries. 

Incrementally Watched Queries can be constructed with varying behaviour. This PR introduces the concept of comparison and differential watched queries. 

Comparison based queries behave similar to standard watched queries. These queries still query the SQLite DB under the hood on each dependant table change, but they compare the result set and only incrementally yield results if a change has been made. The latest query result is yielded as the result set.

Differential queries watch a SQL query and report detailed information on the changes between result sets. This gives additional information such as the `added`, `removed`, `updated` rows between result set changes. 

## Implementation

The logic required for incrementally watched queries requires additional computation and introduces additional complexity to the implementation. For these reasons a new concept of a `WatchedQuery` class is introduced, along with a  new `query ` method allows building a instances of `WatchedQuery`s via `watch` and `differentialWatch` methods. 

``` typescript
// Create an instance of a WatchedQuery.
const listsQuery = powersync
  .query<EnhancedListRecord>({
    sql: /* sql */ `
      SELECT
        ${LISTS_TABLE}.*,
        COUNT(${TODOS_TABLE}.id) AS total_tasks,
        SUM(
          CASE
            WHEN ${TODOS_TABLE}.completed = true THEN 1
            ELSE 0
          END
        ) as completed_tasks
      FROM
        ${LISTS_TABLE}
        LEFT JOIN ${TODOS_TABLE} ON ${LISTS_TABLE}.id = ${TODOS_TABLE}.list_id
      GROUP BY
        ${LISTS_TABLE}.id;
    `
  })
  .differentialWatch();
``` 

The `listsQuery` is smart, it:

- Automatically reprocesses itself if the PowerSync schema has been updated with `updateSchema`
- Automatically closes itself when the PowerSync client has been closed.
- Allows for the query parameters to be updated after instantiation.
- Allows shared listening to state changes.

``` typescript
// The registerListener method can be used multiple times to listen for updates.
// The returned dispose function can be used to unsubscribe from the updates.
const disposeSubscriber = listsQuery.registerListener({
  onData: (data) => {
    // This callback will be called whenever the data changes.
    // The data is the result of the executor.
    console.log('Data updated:', data);
  },
  onDiff: (diff) => {
    // This callback will be called whenever the data changes.
    console.log('Data updated:', diff.added, diff.updated);
  },
  onStateChange: (state) => {
    // This callback will be called whenever the state changes.
    // The state contains metadata about the query, such as isFetching, isLoading, etc.
    console.log(
      'State changed:', 
      state.error, 
      state.isFetching, 
      state.isLoading, 
      state.data
    );
  },
  onError: (error) => {
    // This callback will be called if the query fails.
    console.error('Query error:', error);
  }
});
```
`WatchedQuery` instances retain the latest state in memory. Sharing `WatchedQuery` instances can be used to introduce caching and reduce the number of duplicate DB queries between components. 

The incremental logic is customisable. Diff based queries can specify custom logic for performing diffs on the relevant data set. By default a `JSON.stringify` approach is used. Different data sets might have more optimal implementations. 

``` typescript
const watch = powersync
  .query({
    sql: /* sql */ `
      SELECT
        *
      FROM
        assets
    `,
    mapper: (raw) => {
      return {
        id: raw.id as string,
        make: raw.make as string
      };
    }
  })
  .differentialWatch({
    rowComparator: {
      keyBy: (item) => item.id,
      compareBy: (item) => JSON.stringify(item)
    }
  });
```

Updates to query parameters can be performed in a single place, affecting all subscribers.

``` typescript
watch.updateSettings({
      query: new GetAllQuery({ sql: `SELECT * FROM assets OFFSET ? LIMIT 100`, parameters: [newOffset] })
    });
```

## Reactivity

The existing `watch` method and Reactivity packages have been updated to use incremental queries with differentiation defined as an opt-in feature (defaults to no changes).

``` typescript
powersync.watch(
  'select * from assets',
  [],
  {
    onResult: () => {
      // This callback will be called whenever the data changes.
      console.log('Data updated');
    }
  },
  {
    comparator: {
      checkEquality: (current, previous) => {
        // This comparator will only report updates if the data changes.
        return JSON.stringify(current) === JSON.stringify(previous);
      }
    }
  }
);
```

```tsx
/// React hooks
const { data, isLoading, isFetching } = useQuery(`SELECT * FROM cats WHERE breed = 'tabby'`, [], {
      rowComparator: {
            keyBy: (item) => item.id,
            compareBy: (item) => JSON.stringify(item)
      }
  })
```

New hooks have also been added to use shared `WatchedQuery` instances.

``` tsx
// Managing the WatchedQuery externally can extend its lifecycle and allow in-memory caching between components.
const assetsQuery = powersync.query({
/// ....
}).watch();

/// In the component
export const MyComponent = () => {
   // In React one could import the `assetsQuery` or create a context provider for various queries
   const { data } = useWatchedQuerySubscription(assetsQuery)
}
```

The Vue and React hooks packages have been updated internally to remove duplicate implementations of common watched query logic. Historically these packages relied on manually implementing watched queries based off the `onChange` API in order to cater for exposing additional state and custom query executors. The new `WatchedQuery` APIs now support all the hook packages' requirements - this effectively reduces the heavy lifting in reactivity packages.

The `React Supabase Todolist` demo has been updated with some best practices for reducing re-renders using comparison based incrementally watched queries. 

## Differential Queries

These watched queries report the changes between result sets. A `WatchedQueryDifferential` can be accessed by registering a listener on the query.

``` typescript
export interface WatchedQueryDifferential<RowType> {
  readonly added: ReadonlyArray<Readonly<RowType>>;
  /**
   * The entire current result set.
   * Array item object references are preserved between updates if the item is unchanged.
   *
   * e.g. In the query
   * ```sql
   *  SELECT name, make FROM assets ORDER BY make ASC;
   * ```
   *
   * If a previous result set contains an item (A) `{name: 'pc', make: 'Cool PC'}` and
   * an update has been made which adds another item (B) to the result set (the item A is unchanged) - then
   * the updated result set will be contain the same object reference, to item A, as the previous result set.
   * This is regardless of the item A's position in the updated result set.
   */
  readonly all: ReadonlyArray<Readonly<RowType>>;
  readonly removed: ReadonlyArray<Readonly<RowType>>;
  readonly updated: ReadonlyArray<WatchedQueryRowDifferential<Readonly<RowType>>>;
  readonly unchanged: ReadonlyArray<Readonly<RowType>>;
}

// Listening
watchedQuery.registerListener({
    onDiff: (diff) => console.log(diff)
})
```

A common use case for this is processing newly created items as they are added. The `YJS React Supabase Text Collab` demo has been updated to take advantage of this feature. Document updates are watched via a differential incremental query. New updates are passed to YJS for consolidation as they are synced.

# Early Access

This can be tested by applying the following package versions to your project.

```json
{
  "@powersync/common": "0.0.0-dev-20250722092404",
  "@powersync/node": "0.0.0-dev-20250722092404",
  "@powersync/op-sqlite": "0.0.0-dev-20250722092404",
  "@powersync/react": "0.0.0-dev-20250722092404",
  "@powersync/react-native": "0.0.0-dev-20250722092404",
  "@powersync/tanstack-react-query": "0.0.0-dev-20250722092404",
  "@powersync/vue": "0.0.0-dev-20250722092404",
  "@powersync/web": "0.0.0-dev-20250722092404"
}
```

### DIFF
--- a/packages/vue/README.md
+++ b/packages/vue/README.md
@@ -105,7 +105,7 @@ const addList = () => {
 
 By default, the `useQuery` composable emits a new `data` reference whenever any change occurs in the query's dependent tables. With incremental queries, updates are only emitted when relevant changes are detected, and the internal `data` array preserves object references for unchanged rows between emissions. This helps prevent unnecessary re-renders in your components.
 
-````Vue
+```Vue
 // TodoListDisplayQuery.vue
 <script setup>
 import { usePowerSync, useQuery } from '@powersync/vue';
@@ -129,6 +129,7 @@ const { data: lists, isLoading, isFetching, error} = useQuery('SELECT * FROM lis
         <li v-for="l in lists" :key="l.id">{{ l.name }}</li>
     </ul>
 </template>
+```
 
 ### Query Subscriptions
 
@@ -171,7 +172,6 @@ const { data: lists } = useWatchedQuerySubscription(getListsQuery());
 </template>
 ```
 
-
 ### Static Query
 
 The `useQuery` composable can be configured to only execute initially and not every time changes to dependent tables are detected. The query can be manually re-executed by using the returned `refresh` function.
@@ -193,7 +193,7 @@ const { data: list, refresh } = useQuery('SELECT id, name FROM lists', [], {
   <button @click="refresh">Refresh list</button>
 </template>
 
-````
+```
 
 ### TypeScript Support
 


packages/react/src/hooks/useStatus.ts
```diff
@@ -14,4 +14,4 @@ import { usePowerSyncStatus } from './usePowerSyncStatus';
  *   </div>
  * };
  */
-export const useStatus = usePowerSyncStatus;
+export const useStatus = () => usePowerSyncStatus();
```
stevensJourney, line 17: This fixes an issue where the docs state the `useStatus` is deprecated due to `usePowerSyncStatus` being deprecated.

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -870,6 +889,44 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
...
+   * This is useful for only getting updates when the result set changes, or viewing the change in the result set over time.
+   * @returns A {@link WatchedQueryBuilder} for the specified watch mode.
+   *
+   * For a comparison based watch, use {@link IncrementalWatchMode.COMPARISON}.
+   * See {@link ComparisonWatchedQueryBuilder} for more details.
+   * @example
+   * ```javascript
+   * const watchedQuery = powerSync
+   *  .incrementalWatch({ mode: IncrementalWatchMode.COMPARISON })
+   *  .build({
```
stevensJourney: The thought process behind this pattern is to:

- Cater for the different  `options` required by various incremental query implementations. There are subtle differences between the `options` required for `comparison` and `differential` modes.  Specifying the `mode` first allows the following `build` step's `options` to only contain relevant items and makes Type checking while creating a `WatchedQuery` easier to satisfy. 
- Reduce the number of `watch` based methods on `AbstractPowerSyncDatabase`. We already have `watch`, `watchWithCallback` and `watchWithAsyncGenerator`. Adding another 2 (and in future potentially 3 or more) methods seems excessive.

An alternative might be to expose these under a static group e.g. 

``` js
var w = db.watch.comparison({...options})
var y = db.watch.differential({...options})
// or
var y = db.incrementalWatch.differential({...options})
// etc
```

Chriztiaan: This is nice.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -222,6 +224,9 @@ export abstract class AbstractStreamingSyncImplementation
   protected _lastSyncedAt: Date | null;
   protected options: AbstractStreamingSyncImplementationOptions;
   protected abortController: AbortController | null;
+  // In rare cases, mostly for tests, uploads can be triggered without being properly connected.
+  // This allows ensuring that all upload processes can be aborted.
+  protected uploadAbortController: AbortController | null;
```
stevensJourney, line 238: The changes here are more of a redundancy safety check. I don't believe this is yet required in real world use-cases, but it does become relevant for certain unit tests like `should trigger uploads from last connected clients`. Adding this just for a unit test is usually not good enough justification, but having an extra layer of disposing async background operations seems like a good thing to have in general.

packages/web/tests/multiple_instances.test.ts
```diff
@@ -188,20 +189,25 @@ describe('Multiple Instances', { sequential: true }, () => {
...
       remote: new WebRemote(connector1),
-      uploadCrud: async () => {
-        triggerUpload1();
-        connector1.uploadData(db);
-      },
       db: db.database as WebDBAdapter,
       identifier,
-      retryDelayMs: 100,
+      // The large delay here allows us to test between connection retries
+      retryDelayMs: 90_000,
```
stevensJourney, line 197: The test here uses a few hacks in order to mock multiple clients connected to a connected sync service. Mocking the shared sync worker is not as trivial as the standard sync implementation. 

This test broke due to the latest changes. Adding protection to the `closed` status of `DBAdapter` queries broke uploads entirely. This is because the forcing of `connected: true` allowed an upload loop to start in a manner which was not stopped after disposing the sync client (it should never have been possible to start this loop in the first case though). This loop would continue retries even after the sync client was disposed, but using a closed `DBAdapter`.

The updated logic here will:
- Create a connection to the shared sync manager and attempt connection.
    - The connection attempt will fail since we don't have an actual PowerSync service to connect to. But we set the retry delay to a large number - effectively pausing the connection loop. 
    - We then manually force the sync service into a connected state.
- Uploads can now be performed. We can verify the respective clients are used for uploads.

packages/react/src/hooks/watched/useWatchedQuerySubscription.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  React.useEffect(() => {
+    const dispose = query.registerListener({
+      onStateChange: (state) => {
+        setOutputState({ ...state });
+      }
+    });
+
+    return () => {
+      dispose();
```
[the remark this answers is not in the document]
stevensJourney: This hook only manages a subscription to an existing WatchedQuery, it does not manage the entire instance's lifecycle.

packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
```diff
@@ -175,20 +185,49 @@ export class LockedAsyncDatabaseAdapter
...
     );
   }
 
   async writeLock<T>(fn: (tx: LockContext) => Promise<T>, options?: DBLockOptions | undefined): Promise<T> {
     await this.waitForInitialized();
-    return this.acquireLock(async () =>
-      fn(this.generateDBHelpers({ execute: this._execute, executeRaw: this._executeRaw }))
+    return this.acquireLock(
+      async () => fn(this.generateDBHelpers({ execute: this._execute, executeRaw: this._executeRaw })),
+      {
```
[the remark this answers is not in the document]
stevensJourney: I don't think this comment is correct. The `pendingAbortControllers` are managed in the `aquireLock` method - where pending controllers are removed before calling the `callback`.

packages/common/src/client/watched/processors/ComparisonWatchedQueryBuilder.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/**
+ * Default implementation of the {@link WatchedQueryComparator} for watched queries.
+ * It uses JSON stringification to compare the entire result set.
+ * Array based results should use {@link ArrayComparator} for more efficient item comparison.
+ */
+export const DEFAULT_WATCHED_QUERY_COMPARATOR: WatchedQueryComparator<any> = {
+  checkEquality: (a, b) => JSON.stringify(a) === JSON.stringify(b)
+};
```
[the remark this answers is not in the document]
stevensJourney: The watched query implementation is compatible with custom query executors. These executors might not return an Array - therefore the default comparator is not the `ArrayComparator`.

packages/web/tests/multiple_instances.test.ts
```diff
@@ -132,11 +132,12 @@ describe('Multiple Instances', { sequential: true }, () => {
         await connector1.uploadData(db);
       },
       identifier,
+      retryDelayMs: 90_000, // Large delay to allow for testing
       db: db.database as WebDBAdapter
     };
 
     const stream1 = new SharedWebStreamingSyncImplementation(syncOptions1);
-
+    await stream1.connect();
```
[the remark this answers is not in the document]
stevensJourney: These are in two different tests :)

packages/react/src/hooks/watched/watch-utils.ts
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export const constructCompatibleQuery = <RowType>(
+  query: string | CompilableQuery<RowType>,
+  parameters: any[] = [],
+  options: AdditionalOptions
+) => {
+  const powerSync = usePowerSync();
+
+  const parsedQuery = React.useMemo<WatchCompatibleQuery<RowType[]>>(() => {
```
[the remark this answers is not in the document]
stevensJourney: Fair point, the default executor does use PowerSync. I've added that to the dependency array.

demos/react-supabase-todolist/src/components/providers/SystemProvider.tsx
```diff
@@ -17,10 +25,65 @@ export const db = new PowerSyncDatabase({
...
+      comparator: new ArrayComparator({
+        compareBy: (item) => JSON.stringify(item)
+      }),
+      watch: {
+        // This provides instant caching of the query results.
+        // SQLite calls are asynchronous - therefore on page refresh the placeholder data will be used until the query is resolved.
+        // This uses localStorage to synchronously display a cached version while loading.
+        // Note that the TodoListsWidget is wraped by a GuardBySync component, which will prevent rendering until the query is resolved.
+        // Disable GuardBySync to see the placeholder data in action.
+        placeholderData: JSON.parse(localStorage.getItem('listscache') ?? '[]') as EnhancedListRecord[],
```
Chriztiaan: This is nice nice nice.

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -888,38 +945,42 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
...
-        );
-      } catch (error) {
-        onError?.(error);
+    const { comparator = FalsyComparator } = options ?? {};
+    // This watch method which provides static SQL is currently only compatible with the comparison mode.
+    // Uses shared incremental watch logic under the hood, but maintains the same external API as the old watch method.
+    const watchedQuery = this.incrementalWatch({ mode: IncrementalWatchMode.COMPARISON }).build<QueryResult | null>({
+      comparator,
+      watch: {
+        query: {
```
Chriztiaan, line 966: This formal query definition works, but a shorthand alternative makes for the query option that takes in the SQL/Params and always does an executeRead/getAll by default/implicitly would be nice.

Chriztiaan (approved):
Happy with the changeset, while massive. Well documented and structured.

packages/common/src/client/AbstractPowerSyncDatabase.ts
stevensJourney: I do understand the sentiment behind this concern. I also thought about this and introduced the `GetAllQuery` helper for this reason. 

It is currently possible to supply a `GetAllQuery` as the `query` option e.g.

```javascript
const watch = powersync
      .incrementalWatch({
        mode: IncrementalWatchMode.COMPARISON
      })
      .build({
        watch: {
          query: new GetAllQuery({
            sql: 'SELECT * FROM assets',
            parameters: []
          }),
          placeholderData: []
        }
      });
```

Which is slightly more typing that just supplying an object of the form `{sql: string, parameters?: any[]}`, but IMO provides a better experience than the alternatives like:

Having `query` be an union type of `{sql: string, parameters?: any[]} | WatchCompatibleQuery`.

 This gives bad autocomplete (IMO) if a dev was trying to create this option (it gives all possible options as a first suggestion).
image

Alternatively, a Discriminated Union type would also not be directly compatible with our current `CompatibleQuery` interface.

packages/react/src/hooks/watched/useWatchedQuerySubscription.ts
```diff
@@ -0,0 +1,36 @@
+import { WatchedQuery, WatchedQueryState } from '@powersync/common';
+import React from 'react';
+
+/**
+ * A hook to access and subscribe to the results of an existing {@link WatchedQuery} instance.
+ * @example
+ * export const ContentComponent = () => {
+ * const { data: lists }  = useWatchedQuerySuspenseSubscription(listsQuery);
```
rkistner: The name is wrong in this example.

packages/react/src/hooks/suspense/useSingleSuspenseQuery.ts
```diff
@@ -0,0 +1,87 @@
...
+          const compiledQuery = parsedQuery.compile();
+          const result = await parsedQuery.execute({
+            sql: compiledQuery.sql,
+            parameters: [...compiledQuery.parameters],
+            db: powerSync
+          });
+          if (signal.aborted) {
+            return; // Abort if the signal is already aborted
+          }
+          console.log('done with query refresh');
```
rkistner: Should remove these log statements (including the one above)

rkistner (commented):
It's a big PR, so just giving some high-level comments to start with:

### Internals & Hooks

I really like the refactoring to move more of the logic into the internals - it's nice that the hooks are now fairly thin wrappers around the internal methods.

### incrementalWatch

The `db.incrementalWatch` API feels a little difficult to follow with the current structure. It feels like the API changes completely based on which mode you select, and should perhaps rather be two separate APIs?

Another option is to invert the API to start with the query part first, for example something like this:

```js
db.createQuery(sql, params).differentialWatch(...)
```

This can avoid the explosion of new methods on the database itself, instead having the different methods on a new Query-type class. You could perhaps also have different methods for creating this from different sources, e.g. to create from a Kysely or Drizzle query.

### Incremental comparison queries

I see the main benefit of these to avoid re-renders, to improve rendering performance. It would be good to see some benchmarks to validate that this is worth the additional complexity, that we're not just moving overhead from the rendering part to the query part.

I also wonder if we should unify comparitors and differentiators, always using the differentiator API (identify + compareBy instead of just compare). This way, you can do the re-rendering on an even more granular level. So if you have 1 row out of 1000 results that changed, you can keep the other 999 row objects from the previous results, avoiding further re-renders on those rows. Once again, we'd need some form of benchmarks to check that this is actually worth it.

stevensJourney:
Update:

I've updated the implementation to guard the new watch methods behind the new `query` and `customQuery` methods on `AbstractPowerSyncDatabase`, following Ralf's recommendation. The PR description now includes usage examples.

Comparison based queries are now accessed via `db.query().watch()`. This watched query method does not perform a full diff on result sets; instead, it can exit early as soon as a change is detected. Use this method when you don't need to know the specifics of what changed and don't need to maintain previous object references in the result set. For cases where a full diff and object reference preservation are required, use `db.query(...).differentialWatch()`.

The `DifferentialWatchedQuery` has been improved to better maintain previous array object references, further reducing unnecessary renders during incremental updates.

The `WatchedQuery` implementation now explicitly defines `WatchedQueryState` as `readonly`, making all state members and `data` items immutable. The Vue hooks are still in version `0`; their APIs have been updated to also convey state as readonly. The React hooks package is already at a stable `v1` release, so changing the API to report readonly state would be a breaking change. Backwards compatibility is maintained by conditionally applying readonly qualifiers if a `differentiator` parameter is supplied to the hook. Existing user code should not be affected. 

An automated React profiling benchmark suite has been added in `packages/react/tests/profile.test.tsx` to measure the impact of differential queries on render performance. This test validates how using React memoization together with differential queries can significantly reduce render times. According to the results, incremental updates to the list widget render 60–80% faster compared to standard query methods. For details on the testing methodology, refer to the test file. 
TLDR: The test starts with an initial amount of list items being rendered. We then incrementally add new items while measuring the render time of the entire list widget. Differential queries coupled with React memoization allow the incremental renders to only render the newly added item widgets, while the standard query methods re-render the entire widget. The graph below illustrates the render time for a list widget with various query and memoization methods.

image

Chriztiaan (approved):

packages/react/src/hooks/suspense/useSingleSuspenseQuery.ts
stevensJourney: 🙏 Removed this

packages/react/src/hooks/watched/useWatchedQuerySubscription.ts
stevensJourney: 👍 Fixed

demos/yjs-react-supabase-text-collab/supabase/functions/merge-document-updates/index.ts
```diff
@@ -55,7 +55,8 @@ Deno.serve(async (req) => {
     // insert the new merged update as new single update for the document
     const supabaseInsert = await supabase.from('document_updates').insert({
       document_id: document_id,
-      update_data: Uint8ArrayToHex(docState)
+      update_data: Uint8ArrayToHex(docState),
+      editor_id: 'merged_update'
```
rkistner: editor_id is a uuid in the sql, so this will fail.

But storing the data as hex also looks weird - not sure if this function works at all?

demos/yjs-react-supabase-text-collab/src/library/powersync/PowerSyncYjsProvider.ts
```diff
@@ -34,57 +36,70 @@ export class PowerSyncYjsProvider extends ObservableV2<PowerSyncYjsEvents> {
...
 
     doc.on('updateV2', this._storeUpdate);
     doc.on('destroy', this.destroy);
   }
 
   private async _storeUpdate(update: Uint8Array, origin: any) {
-    if (origin === this) {
-      // update originated from the database / PowerSync - ignore
-      return;
-    }
```
rkistner: This shouldn't be removed - it is used to avoid writing synced updates back to the db again.

demos/yjs-react-supabase-text-collab/src/library/powersync/PowerSyncYjsProvider.ts
```diff
@@ -34,57 +36,70 @@ export class PowerSyncYjsProvider extends ObservableV2<PowerSyncYjsEvents> {
...
     let synced = false;
 
-    const watchLoop = async () => {
-      for await (const results of updates) {
-        if (this.abortController.signal.aborted) {
-          break;
+    updateQuery.registerListener({
+      onStateChange: async () => {
+        for (const added of updateQuery.state.diff.added) {
+          Y.applyUpdateV2(doc, b64ToUint8Array(added.update_b64));
```
rkistner: Need `this` as the origin (see _storeUpdate below). Could alternatively potentially use a different value, e.g. a Symbol indicating the source is the db.

demos/yjs-react-supabase-text-collab/src/library/powersync/PowerSyncYjsProvider.ts
```diff
@@ -24,8 +25,9 @@ export interface PowerSyncYjsEvents {
  * @param documentId
  */
 export class PowerSyncYjsProvider extends ObservableV2<PowerSyncYjsEvents> {
-  private seenDocUpdates = new Set<string>();
   private abortController = new AbortController();
+  // This ID is updated on every new instance of the provider.
+  private id = uuidv4();
```
rkistner: I feel like the previous approach of using the Set was simpler, since it didn't require an additional field in the database to filter the updates. We could still use the Set to only filter out the local updates, or alternatively apply those updates again (should be a no-op to apply the same update again).

demos/yjs-react-supabase-text-collab/src/library/powersync/PowerSyncYjsProvider.ts
```diff
@@ -34,57 +36,70 @@ export class PowerSyncYjsProvider extends ObservableV2<PowerSyncYjsEvents> {
...
 
     let synced = false;
 
-    const watchLoop = async () => {
-      for await (const results of updates) {
-        if (this.abortController.signal.aborted) {
-          break;
+    updateQuery.registerListener({
+      onStateChange: async () => {
+        for (const added of updateQuery.state.diff.added) {
```
rkistner: When I tested this with only one tab open, I often get into a state where the initial "diff" is sent with every update, e.g. "all: 20 rows, added: 20 rows". Haven't debugged further to figure out what is going wrong.

demos/yjs-react-supabase-text-collab/sync-rules.yaml
```diff
@@ -1,10 +1,12 @@
...
-    data:
-      - SELECT * FROM documents
-  updates:
     # Allow remote changes to be synchronized even while there are local changes
     priority: 0
+    parameters: SELECT (request.parameters() ->> 'document_id') as document_id
     data:
-      - SELECT id, document_id, base64(update_data) as update_b64 FROM document_updates
+      - SELECT * FROM documents WHERE id = bucket.document_id
+      - SELECT * FROM documents WHERE id = bucket.document_id
```
rkistner: This is duplicated.

packages/common/src/client/watched/processors/DifferentialQueryProcessor.ts
```diff
@@ -0,0 +1,314 @@
...
+}
+
+type MutableDifferentialWatchedQueryState<RowType> = MutableWatchedQueryState<RowType[]> & {
+  data: RowType[];
+  diff: WatchedQueryDifferential<RowType>;
+};
+
+export interface DifferentialWatchedQuery<RowType>
+  extends WatchedQuery<ReadonlyArray<Readonly<RowType>>, DifferentialWatchedQuerySettings<RowType>> {
+  readonly state: DifferentialWatchedQueryState<RowType>;
```
rkistner: I haven't confirmed with testing, but it feels like accessing the differential state on the query could easily lead to race conditions where updates may be missed. Specifically, there could potentially be a case where the state is processed twice before it is processed in the app, leading to the diff from the first update to be missing.

With the normal async listeners this may not be an issue if the query waits for all listeners to return before processing the next one, but React hooks for example may be called at any later time.

So I'm wondering if we could send the diff in the callback, rather than the state on the query?

React hooks may have bigger issues though, since we have no control over when/how many times they are called. I'm not sure if it is actually feasible to expose the diff over a react hook's state at all.

demos/yjs-react-supabase-text-collab/supabase/functions/merge-document-updates/index.ts
stevensJourney: Good catch for the `uuid` issue.. I've removed the `editor_id` column now, so that should be resolved. I tested the function locally and it seems to work. 
image

demos/yjs-react-supabase-text-collab/src/library/powersync/PowerSyncYjsProvider.ts
stevensJourney: Thanks! I thought filtering the `document_updates` by `editor_id` would avoid that, but I was wrong. I've re-added these.

demos/yjs-react-supabase-text-collab/src/library/powersync/PowerSyncYjsProvider.ts
stevensJourney: Re-added this 🙏

demos/yjs-react-supabase-text-collab/src/library/powersync/PowerSyncYjsProvider.ts
stevensJourney: This was due to using the logic of the `onStateChange` listener. The `diff` on the `state` was relative to the last time the data changed, but the `onStateChange` hook fires for any state change e.g. `isFetching` alternating. This is a good argument for the comment below. I've removed the `diff` from the state now. This query has also been updated.

packages/common/src/client/watched/processors/DifferentialQueryProcessor.ts
stevensJourney: FWIW the listeners here are fired from within the `onChange` callback which uses a `ControlledExecutor` to ensure sequential processing.

This is a fair point though as noted in the YJS demo comment. I've removed the `diff` from the state. The diff can now be accessed by registering an `onDiff` listener.

demos/yjs-react-supabase-text-collab/src/library/powersync/PowerSyncYjsProvider.ts
stevensJourney: I assumed the main reason for having the Set was to avoid re-applying those updates. Given it's a no-op, I've removed this. With the differential queries we will be doing less no-ops, but we will currently still re-apply local edits.

.changeset/plenty-rice-protect.md
```diff
@@ -0,0 +1,6 @@
+---
+'@powersync/react': minor
```
benitav, line 2: missing an entry for the new useWatchedQuerySubscription hook

stevensJourney: I've added entries for this, including examples, in the individual packages' changesets

stevensJourney:
Updates: 
- I've added the `useWatchedQuerySubscription` composable to the `Vue` helpers package. 
- Based off feedback, I've also renamed the `differentiator` option for instantiating differentially watched queries to `comparator`

Chriztiaan (dismissed):

Chriztiaan (approved):

demos/yjs-react-supabase-text-collab/src/library/powersync/PowerSyncYjsProvider.ts
```diff
@@ -33,40 +32,60 @@ export class PowerSyncYjsProvider extends ObservableV2<PowerSyncYjsEvents> {
...
+    const origin = this;
+    updateQuery.registerListener({
+      onDiff: async (diff) => {
+        for (const added of diff.added) {
+          /**
+           * Local document updates get stored to the database and synced.
+           *
+           * These updates here originate from syncing remote updates.
+           * Applying these updates to YJS should not result in the `storeUpdate`
+           * handler creating a new `document_update` record. We mark the origin
```
benitav: Incomplete sentence - "We mark the origin..."?

PR #618, opened 2025-06-02
feat: Bump op-sqlite peer dependency to 13.x.x || 14.x.x for @powersync/op-sqlite

Chriztiaan:
Minor release,  will follow up with a major/stable release soon.

This does a bump to the 13/14 release of the op-sqlite peer dependency. Initially we wanted to communicate which major version we would support by indicated the version in either the major or minor version of this package. We've reverted that idea as it doesn't give enough room for major breaking changes on our side.

Noteworthy changes for 11 > 14 bump include:

1. SQLite updated to 3.49.1
2. Updated SQLCipher to 4.8.0, the opsqlite release docs mentions this could be a breaking change when going from 3>4, and linked an upgrade guide. But I think there was confusion with the openssl versioning, because since [October of last year](https://github.com/OP-Engineering/op-sqlite/issues/161) the project has been on 4.6.1.
3. Monorepo config resolution, you may need to move your `op-sqlite` config from your application's `package.json` to the monorepo root `package.json` depending on where your package manager tool hoists modules (see [1](https://op-engineering.github.io/op-sqlite/docs/installation) and [2](https://github.com/OP-Engineering/op-sqlite/issues/264)).

### DIFF
--- a/.changeset/rotten-pugs-beam.md
+++ b/.changeset/rotten-pugs-beam.md
@@ -2,8 +2,6 @@
 '@powersync/op-sqlite': minor
 ---
 
-Initial stable version release.
-
 Updated op-sqlite upstream peer dependency from 11.x.x to support ^13.x.x and ^14.x.x,
 
 Noteworthy changes from version 11 to version 14 include:


rkistner:
Note the SQLCipher changes mentioned here: https://github.com/OP-Engineering/op-sqlite/releases/tag/13.0.0
That may be a breaking change for SQLCipher users.

Chriztiaan:
> Note the SQLCipher changes mentioned here: https://github.com/OP-Engineering/op-sqlite/releases/tag/13.0.0 That may be a breaking change for SQLCipher users.

I have mentioned relevant links and warnings, echoing the information provided by the op-sqlite release notes.

rkistner (dismissed):

PR #633, opened 2025-06-17
Update core version to version 0.4.0

simolus3:
This updates dependencies to include version `0.4.0` of the core extension:

- `wa-sqlite`
- `react-native-quick-sqlite`
- The version downloaded by the node sdk
- The version linked by our op-sqlite helper package


simolus3:
(worth noting that the web tests seem to have become much more flaky with this - I'm not sure what's up with that. There were some failures in the web sync test before, but I had to restart the run quite a few time to get it to pass. I couldn't reproduce this locally, but it's definitely something we should keep an eye out for)

PR #636, opened 2025-06-18
Fix new sync client with RNQS

simolus3:
For reasons I don't fully understand, the array buffers we get from the RSocket package can't be passed to RNQS directly (it doesn't recognize them as array buffers and binds parameters to null). So, we copy once more.

I've also used this to benchmark syncing 1M rows once more (using RNQS with the demo todolist app and again with the barebones op-sqlite app). The sync is with an app launched using `pnpm ios` on an simulator and a sync service + Postgres on localhost. I've reinstalled the app before each run. Note that the checkpoint applies extremely slowely on the todolist app because there are FTS5 triggers.

| Implementation | Time downloading | Time until data availability |
|-----------------|----------------------------------------|----------------------------------------|
| Baseline (RNQS + JS) | 234.2s | 312.5s |
| Rust (RNQS + Rust) | 61.6s (-73% 😎 ) | 139s (still -55%) |
| Rust and OP-sqlite | 47s (-80% from baseline, -23% from RNQS)  | (not measured, different schema) |

I thought fine-tuning the backpressure handling with RSocket might speed things up even further (and there's still a bit of a "stop and go" behavior visible from sync progress), but increasing the local buffer or raising the watermark did not improve things. Anyway, the results are very promising already.

### DIFF
--- a/packages/react-native/src/sync/bucket/ReactNativeBucketStorageAdapter.ts
+++ b/packages/react-native/src/sync/bucket/ReactNativeBucketStorageAdapter.ts
@@ -1,0 +1,13 @@
+import { PowerSyncControlCommand, SqliteBucketStorage } from '@powersync/common';
+
+export class ReactNativeBucketStorageAdapter extends SqliteBucketStorage {
+  control(op: PowerSyncControlCommand, payload: string | ArrayBuffer | null): Promise<string> {
+    if (payload != null && typeof payload != 'string') {
+      // For some reason, we need to copy array buffers for RNQS to recognize them. We're doing that here because we
+      // don't want to pay the cost of a copy on platforms where it's not necessary.
+      payload = new Uint8Array(payload).buffer;
+    }
+
+    return super.control(op, payload);
+  }
+}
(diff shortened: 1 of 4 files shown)


packages/react-native/src/sync/bucket/ReactNativeBucketStorageAdapter.ts
```diff
@@ -0,0 +1,13 @@
+import { PowerSyncControlCommand, SqliteBucketStorage } from '@powersync/common';
+
+export class ReactNativeBucketStorageAdapter extends SqliteBucketStorage {
+  control(op: PowerSyncControlCommand, payload: string | ArrayBuffer | null): Promise<string> {
+    if (payload != null && typeof payload != 'string') {
+      // For some reason, we need to copy array buffers for RNQS to recognize them. We're doing that here because we
+      // don't want to pay the cost of a copy on platforms where it's not necessary.
+      payload = new Uint8Array(payload).buffer;
```
stevensJourney, line 8: This might be due to the `isArrayBuffer` check [here](https://github.com/powersync-ja/react-native-quick-sqlite/blob/c9d21d38a1f5cdccbe8be2074f46db65b1b86f14/cpp/JSIHelper.cpp#L119) failing - perhaps due to the SDK bundling a polyfil of `Buffer`. I think this is okay for now though.

stevensJourney (approved):
The performance improvements are amazing!

PR #638, opened 2025-06-19
fix: construct `open` options with defined properties only

### LINKED ISSUE #635: @powersync/op-sqlite usage with App Groups
# Issue
The upstream op-sqlite package [supports usage](https://op-engineering.github.io/op-sqlite/docs/configuration#:~:text=App%20groups%20(iOS%20only)**) with [iOS App Groups](https://developer.apple.com/documentation/xcode/configuring-app-groups). Configuring the plist entry should have been enough to use @powersync/op-sqlite with App Groups too, however this package [always provides a dbLocation param](https://github.com/powersync-ja/powersync-js/blob/main/packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts#L123-L139) which overrides the location value that would have been computed by the op-sqlite package. 

This package provides the default dbLocation param because as the [comment](https://github.com/powersync-ja/powersync-js/blob/7ad0c1cc90d8a7d4625bc7b184558c887435d3db/packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts#L124) correctly notes passing undefined/null causes the open function to fail. This happens because the open fn [checks for property existence](https://github.com/powersync-ja/powersync-js/blob/main/packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts#L123-L139) instead of the value being undefined/null and so the intended usage is to not add the location property to the options object to begin with. That'd allow the default dbLocation determination logic of the upstream package to kick in. 

# Solution
Not pass a location property in the [options](https://github.com/powersync-ja/powersync-js/blob/main/packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts#L134-L137) object in the `open()` call unless a user provided override is set. Have this working locally, happy to submit a PR

a11rew:
# Overview
Closes #635

This PR resolves an issue where `@powersync/op-sqlite` would consistently override the default database location behaviour of the upstream `op-sqlite` package. This broke features like iOS App Groups, which depend on the upstream package reading externally defined values (in the Info.plist) to determine the correct database path.

# Changes
To fix this, the openDatabase method has been updated to conditionally add the location property to the open() options only when a custom dbLocation is provided. When it is not provided, the property is omitted, allowing op-sqlite's default location logic to run as intended.

Removed the getDbLocation helper method as well because its logic was a duplication of the [default platform-specific path resolution](https://op-engineering.github.io/op-sqlite/docs/configuration#absolute-paths) already handled by the upstream `op-sqlite` package and is no longer needed when `@powersync/op-sqlite` does not handle determining the default database path


stevensJourney:
Thanks for this! The code looks good to me. We'll do some smoke testing on our side next week before merging.

Chriztiaan:
Tested upgrades for both iOS and Android. Same path is used before and after the upgrade for both platforms (synced data to old version of app, disabled connect call, updated library, verified data is still being queries offline).

PR #642, opened 2025-06-22
Feat/tanstack use queries

whygee-dev:


packages/tanstack-react-query/src/hooks/useQueries.ts
```diff
@@ -0,0 +1,234 @@
...
+          throw e;
+        }
+      };
+      return {
+        ...q.rest,
+        queryFn: q.query ? queryFn : q.rest.queryFn,
+        queryKey: q.rest.queryKey
+      };
+    });
+  }, [stringifiedQueriesDeps, errorsArr]);
```
Chriztiaan: @whygee-dev with the tests you wrote for `useQueries` in hand, it's easy to see that this code works. My concern is that it's a more complicated duplicate of our `useQuery` implementation. This is fine if we just care about getting the tests to pass, but it becomes harder if we want to iterate on our watch/onChange logic in the future.

Would it be possible to reduce some of the code in this hook by depending/extracting some common logic from our `useQuery` Tanstack hook?

whygee-dev: sure, this is actually something I thought of but wasn't sure it was worth it, but the upcoming changes do justify it 👌

Chriztiaan:
This looks better, I'd like to retest it after we merge the incremental watch PR.

packages/tanstack-react-query/tests/useQueries.test.tsx
```diff
@@ -0,0 +1,409 @@
...
+          combine: (results) => results.map((r) => r.data).flat()
+        }),
+      { wrapper }
+    );
+
+    await waitFor(() => {
+      expect(result.current).toContain('SELECT 1');
+      expect(result.current).toContain(1);
+      expect(result.current).toContain('SELECT 2');
+      expect(result.current).toContain(2);
```
Chriztiaan: I have been looking at the docs here: https://tanstack.com/query/latest/docs/framework/react/reference/useQueries#combine

Based on that I don't really understand why we expect/want the query or parameters to be in the result at all. I would suggest also mocking a result in here so that we can see how combine works with actual data.

packages/tanstack-react-query/tests/useQueries.test.tsx
```diff
@@ -0,0 +1,409 @@
...
+            { queryKey: ['q2'], query: 'SELECT 2' }
+          ]
+        }),
+      { wrapper }
+    );
+
+    await waitFor(() => {
+      const results = result.current as any[];
+      expect(results[0].data[0].test).toEqual('custom');
+      expect(results[1].data[0]).toEqual('SELECT 2');
```
Chriztiaan: Echoing the previous comment, I logged what we have for the original `useQuery` implementation. This seems inconsistent, am I missing something?
image

packages/tanstack-react-query/tests/useQueries.test.tsx
whygee-dev: it's because of the mock   

```
getAll: vi.fn((sql, params) => Promise.resolve([sql, ...(params || [])]))
```

I'll improve it so it reflects better real results

Chriztiaan: Thanks for the cleanup. It helps to have unit tests that illustrate usage similar to what devs might actually do.

Chriztiaan:
Last bit of housekeeping @whygee-dev , could you add a changeset entry? `npm run changeset`/`pnpm changeset`, and just add a nice patch entry describing what you did for the package :)

Might also need to update with main. And then we can get this in!

whygee-dev:
> Last bit of housekeeping @whygee-dev , could you add a changeset entry? `npm run changeset`/`pnpm changeset`, and just add a nice patch entry describing what you did for the package :)
> 
> Might also need to update with main. And then we can get this in!

done

Chriztiaan (approved):

Chriztiaan:
Hey @whygee-dev ! Really sorry for the long delay on this, appreciate your patience. 

I've just pushed some updates (fixed imports and added explicit types + jsdoc examples). Would you mind giving it one last look when you get a chance?

Thanks again for the contribution!

whygee-dev:
@Chriztiaan 

> Added explicit types for useQueries() options. Added jsdoc examples.

LGTM ! Thank you

Chriztiaan:
Thanks for checking @whygee-dev,
I was doing a last check by incorporating your work in a demo. Do you know why `error` isn't being set here?
 ```
  const x = useQueries({
    queries: [
      { queryKey: ['todoLists'], query: 'SELECT * FROM lists' },
      { queryKey: ['todos'], query: 'sele' }
    ]
  });
 ```
image

whygee-dev:
Hmm I'm not able to reproduce while incorporating useQueries into the existing example-nextjs demo

```ts
  const [customersList, customers] = useQueries({
    queries: [
      { queryKey: ['customersList'], query: 'SELECT * FROM customers' },
      { queryKey: ['customers'], query: 'sele' }
    ]
  });

  console.log("customersList error", customersList.error);
  console.log("customers error", customers.error);
 ```
 
 
image

**Comment by Chriztiaan** [comment]
So @whygee-dev it seems like there is something strange happening.
On the react-supabase-demo I am seeing this:
It shows null, and after about 5 seconds the error shows for one moment and then disappears. I think the error is accidentally being cleared.
image

**Comment by whygee-dev** [comment]
> So @whygee-dev it seems like there is something strange happening. On the react-supabase-demo I am seeing this: It shows null, and after about 5 seconds the error shows for one moment and then disappears. I think the error is accidentally being cleared. image

Can you push somewhere the reproduction please ? I'll take a look ASAP

**Comment by Chriztiaan** [comment]
> Can you push somewhere the reproduction please ? I'll take a look ASAP
Sure @whygee-dev, please the following. You might just need to populate the .env file if you want to use it straight from there.

Repo: https://github.com/Chriztiaan/tanstack-powersync-reproduction
Diff: https://github.com/Chriztiaan/tanstack-powersync-reproduction/commit/64091c303fd2e771e63cb6605a7860a6f96555f6

**Comment by whygee-dev** [comment]
> > Can you push somewhere the reproduction please ? I'll take a look ASAP
> > Sure @whygee-dev, please the following. You might just need to populate the .env file if you want to use it straight from there.
> 
> Repo: https://github.com/Chriztiaan/tanstack-powersync-reproduction Diff: [Chriztiaan/tanstack-powersync-reproduction@64091c3](https://github.com/Chriztiaan/tanstack-powersync-reproduction/commit/64091c303fd2e771e63cb6605a7860a6f96555f6)

So after investigating, it seems like it the error is unset and soon after is set again. This seems expected behavior by tanstack query retry mechanism

From: https://tanstack.com/query/latest/docs/framework/react/guides/query-retries

> When a useQuery query fails (the query function throws an error), TanStack Query will automatically retry the query if that > query's request has not reached the max number of consecutive retries (defaults to 3) or a function is provided to 
> determine if a retry is allowed.

So what happens is when tanstack query retries, the error is unset during the loading.

I can confirm this by disabling the retry 

```ts
  const [customersList, customers] = useQueries({
    queries: [
      { queryKey: ['lists'], query: 'SELECT * FROM lists' },
      { queryKey: ['todos'], query: 'sele', retry: false }
    ]
  });
```

The error is then never unset after it has been set

stevensJourney (approved):

Chriztiaan:
Thanks again for this contribution @whygee-dev! 🚀

PR #643, opened 2025-06-24
Fix `connect()` resolving early with Rust client

simolus3:
With the JavaScript sync client, the promise returned by `connect()` is supposed to return after a connection has been established and the first sync line received (this is different to all our other SDKs, which return as soon as a connection task has been set up and started).

The way this behavior is implemented is that `connect()` registers a `statusUpdated` handler that completes once the `connected` status is changed. This works with the JavaScript client because it only changes fields that actually need to change. With Rust however, we receive a full `SyncStatus` snapshot instead (so we call `updateSyncStatus` with a complete status that will always have `connected` set - even to `false` initially).

The workaround feels a bit hacky, but since something like `statusUpdated` doesn't really work with the Rust client at all, the best approach I could come up with is to exclude `connected` from the options passed to `updateSyncStatus` in the initial connection setup status.

This fixes a bug [reported on Discord](https://discord.com/channels/1138230179878154300/1385731015804326020/1386736722905272552). However, I could not report longer connection times with the new Rust client after adding the `connect()` call with the fix.

### DIFF
--- a/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
+++ b/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
@@ -945,11 +945,16 @@ The next upload iteration will be delayed.`);
         }
 
         const info = instruction.UpdateSyncStatus.status;
+        const lastStatus = syncImplementation.syncStatus;
         const coreCompleteSync = info.priority_status.find((s) => s.priority == FULL_SYNC_PRIORITY);
         const completeSync = coreCompleteSync != null ? coreStatusToJs(coreCompleteSync) : null;
 
         syncImplementation.updateSyncStatus({
-          connected: info.connected,
+          // The first update to the connected field should only happen when it has really changed - there's a
+          // statusUpdated listener in connect() that will only make the promise complete once connected has changed.
+          // We only want to apply that workaround while we're connecting because it's only relevant in the initial
+          // connect call. Afterwards, we want to forward the sync status unchanged.
+          connected: lastStatus.connected == info.connected && info.connecting ? undefined : info.connected,
           connecting: info.connecting,
           dataFlow: {
             downloading: info.downloading != null,
--- a/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
+++ b/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
@@ -410,23 +410,15 @@ The next upload iteration will be delayed.`);
     this.abortController = controller;
     this.streamingSyncPromise = this.streamingSync(this.abortController.signal, options);
 
-    // Return a promise that resolves when the connection status is updated
+    // Return a promise that resolves when the connection status is updated to indicate that we're connected.
     return new Promise<void>((resolve) => {
       const disposer = this.registerListener({
-        statusUpdated: (update) => {
-          // This is triggered as soon as a connection is read from
-          if (typeof update.connected == 'undefined') {
-            // only concern with connection updates
-            return;
-          }
-
-          if (update.connected == false) {
-            /**
-             * This function does not reject if initial connect attempt failed.
-             * Connected can be false if the connection attempt was aborted or if the initial connection
-             * attempt failed.
-             */
+        statusChanged(status) {
+          if (status.dataFlowStatus.downloadError != null) {
             this.logger.warn('Initial connect attempt did not successfully connect to server');
+          } else if (status.connecting) {
+            // Still connecting.
+            return;
           }
 
           disposer();
@@ -945,16 +937,11 @@ The next upload iteration will be delayed.`);
         }
 
         const info = instruction.UpdateSyncStatus.status;
-        const lastStatus = syncImplementation.syncStatus;
         const coreCompleteSync = info.priority_status.find((s) => s.priority == FULL_SYNC_PRIORITY);
         const completeSync = coreCompleteSync != null ? coreStatusToJs(coreCompleteSync) : null;
 
         syncImplementation.updateSyncStatus({
-          // The first update to the connected field should only happen when it has really changed - there's a
-          // statusUpdated listener in connect() that will only make the promise complete once connected has changed.
-          // We only want to apply that workaround while we're connecting because it's only relevant in the initial
-          // connect call. Afterwards, we want to forward the sync status unchanged.
-          connected: lastStatus.connected == info.connected && info.connecting ? undefined : info.connected,
+          connected: info.connected,
           connecting: info.connecting,
           dataFlow: {
             downloading: info.downloading != null,
--- a/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
+++ b/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
@@ -342,17 +342,18 @@ export abstract class AbstractStreamingSyncImplementation
         let checkedCrudItem: CrudEntry | undefined;
 
         while (true) {
-          this.updateSyncStatus({
-            dataFlow: {
-              uploading: true
-            }
-          });
           try {
             /**
              * This is the first item in the FIFO CRUD queue.
              */
             const nextCrudItem = await this.options.adapter.nextCrudItem();
             if (nextCrudItem) {
+              this.updateSyncStatus({
+                dataFlow: {
+                  uploading: true
+                }
+              });
+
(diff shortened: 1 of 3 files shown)


packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -928,11 +928,16 @@ The next upload iteration will be delayed.`);
...
 
         const info = instruction.UpdateSyncStatus.status;
+        const lastStatus = syncImplementation.syncStatus;
         const coreCompleteSync = info.priority_status.find((s) => s.priority == FULL_SYNC_PRIORITY);
         const completeSync = coreCompleteSync != null ? coreStatusToJs(coreCompleteSync) : null;
 
         syncImplementation.updateSyncStatus({
-          connected: info.connected,
+          // The first update to the connected field should only happen when it has really changed - there's a
+          // statusUpdated listener in connect() that will only make the promise complete once connected has changed.
```
stevensJourney: The comment here reads a little differently to what my understanding of the implementation is.

The promise in the `connect` method is completed whenever an update to the `connected` member of the sync status is attempted (a change is not required) - it just resolves after the `connected` status is either true or false. Which was typically the first update after the connection attempt either succeeds or fails. 

It might be possible to update the listener in the connect method to resolve if `connected` is true or if `connecting` is true and a download error is present.

rkistner:
I did some testing to see the exact status sequence during the initial connection attempt on the various versions. For these tests, I added an artificial delay of 1500ms before sending the first checkpoint message on the service (after connecting). The first number in the logs here indicate time since calling `connect()`. Testing using a modified `react-supabase-todolist` demo that doesn't actually connect to supabase.

### main branch, JS implementation:
```
0 'Calling connect()'
167 'Status' {connecting: false, connected: false, downloading: false, uploading: false}
[PowerSyncDatabase] Attempting to connect to PowerSync instance
229 'Status' {connecting: true, connected: false, downloading: false, uploading: false}
[PowerSyncStream] requesting lock for  streaming-sync-sync-example.db
[PowerSyncStream] Streaming sync iteration started
[PowerSyncStream] Requesting stream from server
[PowerSyncStream] calling the last port client provider for credentials
[PowerSyncStream] Stream established. Processing events
1898 'Status' {connecting: false, connected: true, downloading: false, uploading: false}
1901 'connect() resolved. connected?' true
1902 'Status' {connecting: false, connected: true, downloading: false, uploading: true}
1902 'Status' {connecting: false, connected: true, downloading: true, uploading: true}
[PowerSyncStream] validateChecksums priority, checkpoint, result item undefined {last_op_id: '107100665', buckets: Array(3)} {result: '{"valid":true,"failed_buckets":[]}'}
1904 'Status' {connecting: false, connected: true, downloading: true, uploading: false}
[PowerSyncStream] validated checkpoint {last_op_id: '107100665', buckets: Array(3)}
1915 'Status' {connecting: false, connected: true, downloading: false, uploading: false}
2899 'Status' {connecting: false, connected: true, downloading: false, uploading: true}
2901 'Status' {connecting: false, connected: true, downloading: false, uploading: false}
```

Comments:
1. First status update after calling `connect()` has `connecting: false`, which feels wrong.
2. It feels weird to only have `connected: true` once the first line was received, rather than when the connection was established.
3. We only get `downloading: true` a short while after the first checkpoint line was received. I feel it would be better to have `downloading: true` together with the first `connected: true` status, since we're effectively checking for new changes at that point.
4. Around a second after the connection is established, there is a spike of `uploading: false -> true -> false`, even though there are no changes to upload. This is consistent for me.

### main branch, RUST implementation:
```
0 'Calling connect()'
185 'Status' {connecting: false, connected: false, downloading: false, uploading: false}
[PowerSyncDatabase] Attempting to connect to PowerSync instance
268 'Status' {connecting: true, connected: false, downloading: false, uploading: false}
[PowerSyncStream] requesting lock for  streaming-sync-sync-example.db
[PowerSyncStream] Initial connect attempt did not successfully connect to server
l[PowerSyncStream] calling the last port client provider for credentials
390 'connect() resolved. connected?' false
1900 'Status' {connecting: false, connected: true, downloading: true, uploading: false}
[PowerSyncStream] Validated and applied checkpoint
1908 'Status' {connecting: false, connected: true, downloading: false, uploading: false}
```

Comments:
1. This shows the core issue fixed by this PR: `connect()` resolving before we have `connected: true`.
2. This does have `connected: true, downloading: true` as the first status update after connecting, which is nice.
3. No uploading status being toggled after connecting.

### this branch, RUST implementation:
```
0 'Calling connect()'
171 'Status' {connecting: false, connected: false, downloading: false, uploading: false}
[PowerSyncDatabase] Attempting to connect to PowerSync instance
244 'Status' {connecting: true, connected: false, downloading: false, uploading: false}
[PowerSyncStream] requesting lock for  streaming-sync-sync-example.db
[PowerSyncStream] calling the last port client provider for credentials
1883 'Status' {connecting: false, connected: true, downloading: true, uploading: false}
1887 'connect() resolved. connected?' true
1887 'Status' {connecting: false, connected: true, downloading: true, uploading: true}
[PowerSyncStream] Validated and applied checkpoint
1890 'Status' {connecting: false, connected: true, downloading: false, uploading: true}
1892 'Status' {connecting: false, connected: true, downloading: false, uploading: false}
2883 'Status' {connecting: false, connected: true, downloading: false, uploading: true}
2886 'Status' {connecting: false, connected: true, downloading: false, uploading: false}
```

Comments:
1. This shows the issue is fixed.
2. This does also have `connected: true, downloading: true` as the first status update after connecting.
3. Uploading status is changed to true just after the connection is established and again a second later. I don't know why this is different from the main branch, but I consistently get this behavior here and not on the main branch.

### Overall comments

1. This fix is a clear improvement, so I'm happy for this to be merged and released as-is. With this change the rust implementation's status reporting appears to be on-par or better than the JS implementation.
2. I also don't see any actual performance difference here in connection times between the rust and JS implementations.

However, I think we should improve the status reporting further:
1. The first status update should have `connecting: true`.
2. We should have `connected: true` and resolve the `connect()` Promise as soon as the connection is established (or perhaps the rsocket stream is opened?), not only when we receive the first sync line, as there can be a significant delay before we receive the checkpoint in some cases.
3. Together with the above, we should have `downloading: true` for the first status update where we have `connected: true` (already the case with the rust implementation, but not the JS one).
4. We should not have `uploading: true` when there isn't data to upload.
5. There appears to be a significant delay (over 200ms) before getting `connecting: true` on all versions here. I'm not sure where that delay comes from - some of it may be overhead from the demo app or using a debug build.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
simolus3: > It might be possible to update the listener in the connect method to resolve if `connected` is true or if `connecting` is true and a download error is present.

I think I prefer that approach 👍 Done!

simolus3:
> The first status update should have `connecting: true`.

At least in the node test (`connect() waits for connection`), that appears to be the case.

> We should have `connected: true` and resolve the `connect()` Promise as soon as the connection is established (or perhaps the rsocket stream is opened?)

Fixed for Rust now.

> Together with the above, we should have `downloading: true` for the first status update where we have `connected: true` 

I don't agree with this one in combination with the one above. How are we downloading rows before the first `checkpoint` line? The Rust client sets `downloading: true` from the checkpoint until a `checkpoint_complete` message which sounds like the correct behavior to me. Which behavior would you expect, `downloading: true` initially until the first checkpoint is completed and then again for subsequent ones?

> We should not have `uploading: true` when there isn't data to upload.

Fixed!

stevensJourney (dismissed):
From my side this looks good to me.

rkistner (dismissed):
> I don't agree with this one in combination with the one above. How are we downloading rows before the first checkpoint line? The Rust client sets downloading: true from the checkpoint until a checkpoint_complete message which sounds like the correct behavior to me. Which behavior would you expect, downloading: true initially until the first checkpoint is completed and then again for subsequent ones?

For this one, the general issue is that if you get `{connecting: false, connected: true, downloading: false}`, there is no nice way to know if it is "we are expecting a checkpoint soon which may contain data to download" (which is the case on initial connection) versus "connection is idle" (already received a checkpoint). This could potentially be solved by adding another status field. But either way this is not a regression, so happy to revisit later.

stevensJourney (approved):

PR #644, opened 2025-06-25
feat: Common-js output for @powersync/common

Chriztiaan:
Added a CommonJS output for the @powersync/common package.


Chriztiaan:
Verified via a dev release that this works on end user's side.

PR #647, opened 2025-06-30
[POC] SQL.js

stevensJourney:
# Overview

Certain environments have trouble loading PowerSync due to it requiring a SQLite implementation which supports loading our Rust core extension. One of these environments is Expo Go which is a prebuilt app container for Expo apps.

This POC uses a pure JS implementation of SQLite via [SQL.js](https://sql.js.org/#/). SQLite and our Rust core extension are compiled to JavaScript via Emscripen. This pure JS payload can be used in environments such as Expo Go.

SQL.js already has a pure JS build, this POC locally links PowerSync into that build process. This POC applies Git patches to the SQL.js repo and inits a build. 

A small `DBAdapter` is included for SQL.js. This POC currently uses this adapter in the React Native Supabase Todolist Demo. This demo can now be started in Expo GO.

## Outstanding work

- SQL.js does support SQLite update hooks, but these don't seem to trigger. I've added extra temporary utilities to attempt to debug, but have not been successful. 
- This currently builds SQL.js in the monorepo. This requires toolchains which we should not expect all users of the monorepo to have present. We could fallback to prebuilt assets in such cases.

# Demo

Note that watched queries do not currently work due to the update hook issue.

https://github.com/user-attachments/assets/13c1226a-54ff-448c-ac85-33c7c351f437


demos/react-native-supabase-todolist/package.json
```diff
@@ -14,6 +14,7 @@
     "@powersync/attachments": "workspace:*",
     "@powersync/common": "workspace:*",
     "@powersync/react": "workspace:*",
+    "@powersync/dev-adapter": "workspace:*",
```
stevensJourney: I've put this driver in it's own package for now. 

SQL.js is relatively large. About 2Mb. Alternatively we could publish this under a different export e.g. `@powersync/common/driver`

packages/dev-adapter/rollup.config.mjs
```diff
@@ -0,0 +1,40 @@
...
+      file: 'dist/bundle.mjs',
+      format: 'esm',
+      sourcemap: sourceMap
+    },
+    plugins: [
+      nodeResolve({ preferBuiltins: false, browser: true }),
+      commonjs({}),
+      alias({
+        entries: [
+          { find: 'fs', replacement: path.resolve(__dirname, 'vendored/empty.js') },
```
stevensJourney: The default Emscripten output contains code like `require("fs")`. This seems to be unreachable, but Metro complains when it detects it. This replaces the requires to return empty modules.

packages/dev-adapter/src/SQLJSAdapter.ts
```diff
@@ -0,0 +1,292 @@
...
+
+export class SQLJSOpenFactory implements SQLOpenFactory {
+  constructor(protected options: SQLJSOpenOptions) {}
+
+  openDB(): DBAdapter {
+    return new SQLJSDBAdapter(this.options);
+  }
+}
+
+(globalThis as any).onSqliteUpdate = (
```
stevensJourney: The standard update hook function did not trigger for me. I've implemented this as a workaround.

packages/sqljs/patches/powersync.patch
```diff
@@ -0,0 +1,235 @@
...
+     var sqlite3_errmsg = cwrap("sqlite3_errmsg", "string", ["number"]);
+     var sqlite3_column_count = cwrap(
+@@ -1155,6 +1160,13 @@ Module["onRuntimeInitialized"] = function onRuntimeInitialized() {
+         return sqlite3_changes(this.db);
+     };
+ 
++    /** TODO
++    @return {number} the last insert id
++    */
++    Database.prototype["lastInsertId"] = function lastInsertId() {
```
stevensJourney: This was not exposed by SQL.js, but we need it in order to validate the response from checkpoint validations.

packages/dev/rollup.config.mjs
```diff
@@ -0,0 +1,40 @@
...
+      sourcemap: sourceMap
+    },
+    plugins: [
+      nodeResolve({ preferBuiltins: false, browser: true }),
+      commonjs({}),
+      alias({
+        entries: [
+          { find: 'fs', replacement: path.resolve(__dirname, 'vendored/empty.js') },
+          { find: 'path', replacement: path.resolve(__dirname, 'vendored/empty.js') },
+          { find: 'crypto', replacement: path.resolve(__dirname, 'vendored/empty.js') }
```
simolus3: I think it would be nice to add a comment here explaining why we're doing that - I've seen the PR comment, maybe just copy that?
```suggestion
          // The default Emscripten output contains code like `require("fs")`. This seems
          // to be unreachable, but Metro complains when it detects it.
          { find: 'fs', replacement: path.resolve(__dirname, 'vendored/empty.js') },
          { find: 'path', replacement: path.resolve(__dirname, 'vendored/empty.js') },
          { find: 'crypto', replacement: path.resolve(__dirname, 'vendored/empty.js') }
```

packages/dev/src/SQLJSAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  protected async init(): Promise<SQLJs.Database> {
+    const SQL = await SQLJs({
+      locateFile: (filename: any) => `../dist/${filename}`,
+      print: (text) => {
+        console.log('[stdout]', text);
+      },
+      printErr: (text) => {
+        console.error('[stderr]', text);
```
simolus3: Should these two redirect to e.g. `logger.info` and `logger.warning` perhaps?

packages/dev/src/SQLJSAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+      }
+    });
+    const existing = await this.options.persister.readFile();
+    const db = new SQL.Database(existing);
+    this.dbP = db['db'];
+    // debugger;
+    // (db as any).updateHook(function (operation, database, table, rowId) {
+    //   console.log(`Update Hook: ${operation} on ${table}`);
+    //   this.tableUpdateCache.add(table);
+    // });
```
simolus3: Looks like this can be removed now that we have another workaround?

packages/dev/src/SQLJSAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+      return {
+        rowsAffected: totalRowsAffected
+      };
+    } catch (error) {
+      throw error;
+    }
+  }
+
+  readLock<T>(fn: (tx: LockContext) => Promise<T>, options?: DBLockOptions): Promise<T> {
+    return this.writeLock(fn, options);
```
simolus3: Maybe add a small comment here explaining that we're not using RW locks here because we can't implement connection pools ontop of SQL.js.

packages/dev/src/SQLJSAdapter.ts
```diff
@@ -0,0 +1,331 @@
...
+        totalRowsAffected += db.getRowsModified();
+      }
+
+      stmt.free();
+
+      return {
+        rowsAffected: totalRowsAffected
+      };
+    } catch (error) {
+      throw error;
```
simolus3: I think we may need to free the statement in a `finally` block instead, otherwise we'd be leaking it if `stmt.run()` throws? Also I'm not sure what the purpose of this catch block is.

packages/dev/README.md
```diff
@@ -0,0 +1,85 @@
+# PowerSync Dev
+
+A development package for PowerSync which uses [SQL.js](https://sql.js.org/#/)
```
simolus3: Maybe there could be some motivation here on why this package is necessary, e.g. "which uses SQL.js to avoid native dependencies." or something a little longer?

simolus3 (commented):
This looks like a really helpful addition for Expo Go users 👍 

I've left some minor comments. One other thing that stands out is that we may want to make it clearer in the readme (or perhaps even by logging something when the database is opened) that this is intended for development only. In particular, I feel like we should warn about this being much slower than the "real" SDK, so that we don't get reports about general PowerSync performance coming from users only trying the SQL.js adapter because it's the easiest one to use.

simolus3 (dismissed):
Looks good to me 👍

packages/dev/src/SQLJSAdapter.ts
```diff
@@ -0,0 +1,334 @@
...
+            columnNames = statement.getColumnNames();
+          }
+          rawResults.push(statement.get());
+        }
+
+        const rows = rawResults.map((row) => {
+          return Object.fromEntries(row.map((value, index) => [columnNames![index], value]));
+        });
+        return {
+          insertId: (db as any).lastInsertId(), // does not seem to be available in SQL.js??
```
stevensJourney: The comment here is a bit outdated. The `lastInsertId` method is not present on the original version of `SQL.js`, but it's been added to our fork. IIRC the types for SQL.js come from the `@types/sql.js` package, which is why the `as any` is used here. We should probably update this comment.

packages/dev/README.md
```diff
@@ -0,0 +1,89 @@
+# PowerSync Dev
+
+A development package for PowerSync which uses [SQL.js](https://sql.js.org/#/) to provide a pure JavaScript SQLite implementation.
+This eliminates the need for native dependencies and enables seamless development with Expo Go and other JavaScript-only environments.
+
+This package is specifically designed to streamline the development workflow and will be much slower than DB adapters that use native dependencies.
```
rkistner: I think we should also explicitly mention that the entire database file is re-written to persistent storage every time you do a write. In addition to the performance overhead of that, we also don't provide any of SQLite's consistency guarantees, and you may end up with missing data or a corrupted database file if the app is killed while writing the database file.

packages/dev/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@powersync/dev",
+  "version": "0.0.0",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org/",
+    "access": "public"
+  },
+  "description": "Development helpers for JourneyApps PowerSync",
```
rkistner: I wonder if we should use a more specific name and description here? This makes it sound like general development tooling, while it feels like this should primarily be used when you don't have another option, i.e. Expo Go.

Chriztiaan: I agree with the sentiment, I couldn't think of a better name that makes it clear that is meant for development and isn't being pinned to only being useful for Expo/React-native. It's a universal pure JS db adapter intended for dev use. Very open to suggestions.

rkistner: Maybe something like `@powersync/dev-memory-db` or `@powersync/dev-sql-js-adapter`? Also open to other suggestions

Chriztiaan: I am starting to lean toward `adapter-sql-js`, that would make it easy to distinguish adapter packages in the future.

benitav: with `@journeyapps/react-native-quick-sqlite` and `@powersync/op-sqlite` I'd say `@powersync/sql-js`  or `@powersync/dev-sql-js` follows current conventions best.

Chriztiaan: Since the underlying fork is already named `@powersync/sql-js`, `@powersync/adapter-sql-js` feels best.

.changeset/orange-spies-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/dev': patch
```
rkistner: This should also be updated

packages/adapter-sql-js/src/SQLJSAdapter.ts
```diff
@@ -0,0 +1,335 @@
...
+          // Ignore updates from other databases.
+          return;
+        }
+        this.tableUpdateCache.add(table);
+      }
+    });
+
+    this.writeScheduler = new ControlledExecutor(async (db: SQLJs.Database) => {
+      await this.options.persister.writeFile(db.export());
+    });
```
rkistner, line 100: Looking at the benchmarks, this is probably the cause for slowdown for the in-memory config: This is doing an export of the entire database, right before discarding it. Using a filesystem probably gives better performance because it effectively throttles the exports while waiting for the write, while "in-memory" could trigger an export for every write.

I'd recommend setting persister to null and bypassing this step for the default/in-memory config, instead of creating a no-op persister.

Chriztiaan: Will fix in a follow up PR.

PR #648, opened 2025-06-30
WebSocket keepalive changes

rkistner:
## Background

RSocket handles keepalive as follows:
1. Client sends a keepalive message once every `keepAlive` ms (20s).
2. Server responds with a keepalive message.
3. Client checks that it gets one keepalive response every `lifetime` ms (30s).

That gives a 10s grace period between sending and receiving the message.

The issue here comes in when there is either:
1. A slow network connection, that buffers a lot of data on the TCP level.
2. CPU overhead with parsing the BSON data (observed on React-Native), that could have the same buffering effect.

Either of the above could block the keepalive response for more than that 10s, resulting in `Error: Closed. Original cause [Error: No keep-alive acks for 30000 millis].`. 

One workaround was added in #494, reducing the number of messages that are requested by the client at a time. However, this can decrease throughput.

## Keepalive Changes

This implements an additional workaround that:
1. Checks that we get one message of any type every 30s.
2. Increases the keepalive `lifetime` to 90s.

This applies per-message, not per-byte or per-frame. This means we can still get a timeout error if a single message takes more than 10s to send over the connection. Our messages are generally limited to 1MB in size (unless a single operation is larger than that), so it should cover most cases.

## BSON parsing changes

This now delays parsing of BSON data until the line is processed, rather than when it is received. This helps since:
1. The serialized buffer should use less memory than the parsed data, resulting in slightly less memory usage when we buffer a lot of data.
2. This makes the websocket handling code more responsive, making it less likely that the keepalive messages are significantly delayed.

The Rust implementation takes this further to move the BSON parsing into the SQLite extension, which helps even more.

This refactors the DataStream handling a bit to support this. This also fixes some type issues we had, such as mixing up Uint8Array and ArrayBuffer.

## Future options

1. For NodeJS and Chrome, we can consider using [WebSocketStream](https://developer.mozilla.org/en-US/docs/Web/API/WebSocketStream) instead of WebSocket, which has built-in backpressure support, which may mitigate this issue somewhat. This may require significant changes to RSocket though.
2. We could implement for "dynamic" backpressure, rather than always requesting 10 messages at a time. However, implementing an effective strategy here could be quite difficult.

### NodeJS
Did regression testing on NodeJS using all 4 combinations of `connectionMethod` and `clientImplementation` - no issues found.

### React Native

On React-Native, tested using websockets only, for both `clientImplementation` methods.For the JavaScript implementation, overall sync performance appears to be slightly better now, but I didn't do proper benchmarking.

Before, data would typically be processed batches of 10x deserialize, then 10x saveSyncData, alternating between them. Now, one message is deserialized at a time.

For op-sqlite, I used react-native-barebones-opsqlite. I had trouble with the fetch implementation - both write checkpoints and http streams get `Value is undefined, expected a String` (see https://github.com/facebook/react-native/issues/27741#issuecomment-2672485433). But websockets do work with both the JS and Rust implementations.

### Web

Tested using the diagnostics app, using all 4 combinations of `connectionMethod` and `clientImplementation` - no issues found.

## TODO

 * [x] Fix tests hanging


Chriztiaan (approved):

packages/react-native/src/sync/bucket/ReactNativeBucketStorageAdapter.ts
```diff
@@ -1,13 +1,24 @@
...
 export class ReactNativeBucketStorageAdapter extends SqliteBucketStorage {
-  control(op: PowerSyncControlCommand, payload: string | ArrayBuffer | null): Promise<string> {
-    if (payload != null && typeof payload != 'string') {
-      // For some reason, we need to copy array buffers for RNQS to recognize them. We're doing that here because we
-      // don't want to pay the cost of a copy on platforms where it's not necessary.
-      payload = new Uint8Array(payload).buffer;
+  control(op: PowerSyncControlCommand, payload: string | Uint8Array | ArrayBuffer | null): Promise<string> {
+    if (payload instanceof Uint8Array) {
+      // RNQS doesn't accept Uint8Array arguments - convert to ArrayBuffer first.
+      payload = uint8ArrayToArrayBuffer(payload);
```
rkistner, line 7: This refactoring doesn't actually make a big difference, since in practice we always hit the "copy" branch. The important part here is just properly keeping track of Uint8Array vs ArrayBuffer.

packages/common/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -279,9 +285,9 @@ export abstract class AbstractRemote {
    */
   async socketStreamRaw<T>(
     options: SocketSyncStreamOptions,
-    map: (buffer: Buffer) => T,
+    map: (buffer: Uint8Array) => T,
```
rkistner, line 288: Note that Buffer extends Uint8Array. Keeping as Uint8Array for general compatibility here, and only covert to ArrayBuffer at the last moment for RNQS. Other implementations can keep it as Uint8Array.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -847,10 +847,7 @@ The next upload iteration will be delayed.`);
     // Pending sync lines received from the service, as well as local events that trigger a powersync_control
     // invocation (local events include refreshed tokens and completed uploads).
     // This is a single data stream so that we can handle all control calls from a single place.
-    let controlInvocations: DataStream<{
-      command: PowerSyncControlCommand;
-      payload?: ArrayBuffer | string;
```
rkistner, line 853: This bit was wrong - the payload was never an ArrayBuffer, it was a Buffer/Uint8Array or string.

packages/common/src/utils/DataStream.ts
```diff
@@ -151,56 +168,31 @@ export class DataStream<Data extends any = any> extends BaseObserver<DataStreamL
...
-        stream.enqueueData(callback(data));
-      },
-      closed: () => {
-        stream.close();
-        l?.();
-      }
-    });
-
-    return stream;
-  }
```
rkistner, line 179: This did not properly account for backpressure, and doesn't seem to be used, so I removed this rather.

packages/common/src/utils/DataStream.ts
```diff
@@ -137,11 +142,23 @@ export class DataStream<Data extends any = any> extends BaseObserver<DataStreamL
...
+  protected processQueue() {
     if (this.processingPromise) {
       return;
     }
 
+    let promise = this._processQueue().finally(() => {
+      this.processingPromise = null;
+    });
+    this.processingPromise = promise;
+    return promise;
```
rkistner: @stevensJourney Could you please check if the refactoring here makes sense? The previous implementation appeared to have a potential race condition where you could have multiple concurrent `_processQueue()` invocations

packages/common/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -304,12 +310,26 @@ export abstract class AbstractRemote {
...
     const url = this.options.socketUrlTransformer(request.url);
     const connector = new RSocketConnector({
       transport: new WebsocketClientTransport({
         url,
         wsCreator: (url) => {
-          return this.createSocket(url);
+          const socket = this.createSocket(url);
+          socket.addEventListener('message', (event) => {
+            resetTimeout();
+          });
```
rkistner, line 331: This is the core of the new keepalive implementation - we reset the timer when we receive _any_ message, rather than only on keepalive messages.

packages/common/src/utils/DataStream.ts
stevensJourney: Looks like this is outdated, I went through the latest/updated DataStream implementation and could not think of any potential issues.

PR #649, opened 2025-06-30
Fix `lastSyncedAt` values with Rust client

simolus3:
When the Rust client exports timestamps for the sync status, they're encoded as UTC seconds. `new Date` expects milliseconds though, and so the reported time used to be close to 1970 instead of the actual time.

### DIFF
--- a/.changeset/tall-dolphins-sleep.md
+++ b/.changeset/tall-dolphins-sleep.md
@@ -1,0 +1,5 @@
+---
+'@powersync/common': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/tall-dolphins-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/common': patch
```
stevensJourney, line 2: It helps to add the SDK packages themselves to these kinds of changesets. That way the change message is also present on the individual packages' changelogs.

stevensJourney (approved):

PR #650, opened 2025-06-30
Rust: Trigger CRUD upload on connect

simolus3:
This triggers a CRUD upload when receiving the first sync line to ensure that local writes made while offline have a chance to get uploaded.

This also refactors `DataStream` to allow adding data while it's blocked on `lowWater` - this is used by the Rust sync client because we might have to wait for lines from upstream and also have a upload completion notification that needs to be added to the stream.


packages/common/src/client/ConnectionManager.ts
```diff
@@ -194,7 +194,6 @@ export class ConnectionManager extends BaseObserver<ConnectionManagerListener> {
 
     this.logger.debug('Attempting to connect to PowerSync instance');
     await this.syncStreamImplementation?.connect(appliedOptions!);
-    this.syncStreamImplementation?.triggerCrudUpload();
```
simolus3, line 197: I think this isn't necessary because we're also triggering this in the streaming sync implementation itself (both for `legacyStreamingSyncIteration` and now also for the new one).

Implementing this in the sync client directly is safer because this one only triggers once, whereas we want to trigger the upload every time a sync iteration begins to recover local writes during bad connectivity.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -993,6 +999,7 @@ The next upload iteration will be delayed.`);
       );
 
       this.notifyCompletedUploads = () => {
+        console.log('notify completed');
```
stevensJourney: We should remove this or switch to using a `logger` instance.

packages/common/src/utils/DataStream.ts
```diff
@@ -188,10 +194,13 @@ export class DataStream<Data extends any = any> extends BaseObserver<DataStreamL
     }
 
     if (this.dataQueue.length <= this.lowWatermark) {
-      await this.iterateAsyncErrored(async (l) => l.lowWater?.());
-    }
+      const dataAdded = new Promise((resolve) => {
```
stevensJourney: I think setting updating this to `new Promise<void>(...)` should allow a**void**ing 😄 The `_:null` param of `notifyDataAdded`

PR #653, opened 2025-07-02
Mention metro workaround for op-sqlite

simolus3:
This links the workaround we recommend as a metro config when using the React Native SDK without Expo in the OP-sqlite adapter readme as well. The reason is that we've seen users get weird errors when adopting OP-sqlite without Expo in their existing projects, and the metro configuration we suggest seems to fix them.

### DIFF
--- a/packages/powersync-op-sqlite/README.md
+++ b/packages/powersync-op-sqlite/README.md
@@ -16,6 +16,10 @@ Follow the installation instructions for the [React Native SDK](https://github.c
 npx expo install @powersync/op-sqlite
 ```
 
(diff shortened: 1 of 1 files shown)


stevensJourney (approved):
Thanks for the addition. Note that without a changeset and version bump this README addition won't yet reflect on npmjs.

simolus3:
> Note that without a changeset and version bump this README addition won't yet reflect on npmjs.

Thanks for catching this. I think it may be alright since we can at least point to the readme in our repository. And it's not urgent, so I assume we can just release it with whatever the next update may be.

PR #654, opened 2025-07-06
Raw tables

simolus3:
_This is blocked on rolling 0.4.1 of the core extension into the JS SDKs - but apart from that it works already._ Still, it may be worth holding off a review until I've updated the extension.

This adds support for "raw tables", which are managed by the user instead of being created as JSON views by PowerSync. Benefits include better performance for complex queries as well as support for more complex table definitions (including local table and column constraints).

To enable raw tables, one would currently first create a regular schema and then call `withRawTables` to add them (not sure if we should have a secondary parameter on the `Schema` constructor for this - perhaps after stabilizing this feature?). Raw tables consist of:

1. A name, identifying the table name used by the sync service (so that the client knows to forward those operations to a raw table).
2. A custom upsert statement to run for `PUT` ops.
3. A custom delete statement to run for `REMOVE` ops.

The statements are defined as SQL with prepared statement parameters as well as a value source (which can currently either be the id column or a named column extracted from JSON). The syntax for that is not particularly nice at the moment (it reflects the internal JSON structure used by the core extension). There are multiple options to improve that here, but since the feature is explicitly marked as experimental we can improve on API changes once we better understand how this feature is used in practice.


packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -469,7 +469,10 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
    * Connects to stream of events from the PowerSync instance.
    */
   async connect(connector: PowerSyncBackendConnector, options?: PowerSyncConnectionOptions) {
-    return this.connectionManager.connect(connector, options);
+    const resolvedOptions = options ?? {};
```
stevensJourney: It looks like the `serializedSchema` option is also exposed on the `options`. Is there a use-case where devs would want to supply this in the `connect` call?
image

packages/common/src/db/schema/RawTable.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * A parameter to use as part of {@link PendingStatement}.
+ *
+ * For delete statements, only the `"Id"` value is supported - the sync client will replace it with the id of the row to
+ * be synced.
+ *
+ * For insert and replace operations, the values of columns in the table are available as parameters through
+ * `{Column: 'name'}`.
+ */
+export type PendingStatementParameter = 'Id' | { Column: string };
```
stevensJourney, line 25: Just for interest sake. Why does the `Column` member start with a capital `C`?

.changeset/bright-snakes-clean.md
```diff
@@ -0,0 +1,37 @@
...
+'@powersync/react-native': minor
+---
+
+Add experimental support for raw tables, giving you full control over the table structure to sync into.
+While PowerSync manages tables as JSON views by default, raw tables have to be created by the application
+developer. Also, the upsert and delete statements for raw tables needs to be specified in the app schema:
+
+```JavaScript
+const customSchema = new Schema({});
+customSchema.withRawTables({
```
stevensJourney, line 14: I don't have a strong opinion on this. Just wondering if it's feasible to have the RawTable definition be part of the `Table` instantiation?

I know in JS it might be a little more complicated since we have legacy Table V1 syntax. But maybe having the entire table definition be in one place could be nice. 

Maybe something like:

```js
const todos = new Table(
  {
    list_id: column.text,
    created_at: column.text,
    completed_at: column.text,
    description: column.text,
    created_by: column.text,
    completed_by: column.text,
    completed: column.integer
  },
  {
    indexes: { list: ['list_id'] },
    rawTable: {
      put: {
        sql: 'INSERT OR REPLACE INTO lists (id, name) VALUES (?, ?)',
        // put statements can use `Id` and extracted columns to bind parameters.
        params: ['Id', { Column: 'name' }]
      },
      delete: {
        sql: 'DELETE FROM lists WHERE id = ?',
        // delete statements can only use the id (but a CTE querying existing rows by id could
        // be used as a workaround).
        params: ['Id']
      }
    }
  }
);
```

OR:

```javascript
const todos = new Table(
  {
    list_id: column.text,
    created_at: column.text,
    completed_at: column.text,
    description: column.text,
    created_by: column.text,
    completed_by: column.text,
    completed: column.integer
  },
  {
    indexes: { list: ['list_id'] }
  }
).withRawTables({
  put: {
    sql: 'INSERT OR REPLACE INTO lists (id, name) VALUES (?, ?)',
    // put statements can use `Id` and extracted columns to bind parameters.
    params: ['Id', { Column: 'name' }]
  },
  delete: {
    sql: 'DELETE FROM lists WHERE id = ?',
    // delete statements can only use the id (but a CTE querying existing rows by id could
    // be used as a workaround).
    params: ['Id']
  }
});
```

rkistner: The main issue here is that the normal table definition syntax will not be used at all here (including columns, indexes, and other settings such as insertOnly), which could make things confusing. The columns could still be used for e.g. Kysely types, but the indexes and other options would be ignored completely.

packages/common/src/db/schema/RawTable.ts
simolus3: It's mostly an artifact coming from the fact that we're mapping these JSONs into [Rust enums](https://github.com/powersync-ja/powersync-sqlite-core/blob/8a1efcc8021ca6c4c2828590051c0716c62929cd/crates/core/src/schema/table_info.rs#L253-L265), where using uppercase names for variants and structs is kind of the default.

.changeset/bright-snakes-clean.md
simolus3: I don't think the raw table should be associated with the existing `Table` class, if that's what you mean. The idea of a raw table is that the PowerSync SDK doesn't know much about the table except its name and statements for upserts/deletes. So having to specify columns feels wrong to me - they wouldn't be used at all.

packages/common/src/client/AbstractPowerSyncDatabase.ts
simolus3: I've introduced a hidden internal type so that `serializedSchema` is not visible on the public `connect()` arguments.

.changeset/bright-snakes-clean.md
stevensJourney: Sounds good to me. Thanks for the details.

PR #655, opened 2025-07-06
Export AsyncDatabaseConnection types in `@powersync/web`

LucDeCaf:
The entire change is adding one more file's types to `packages/web/src/index.ts`.

Long story short, I need access to these types in my codebase :)


Chriztiaan:
Hey @LucDeCaf, that file contains internal members which we might have not exposed intentionally so that we might be able to drop it in the future without being a breaking change - in which case I might suggest making your own copies of the types on your side. Any thoughts @stevensJourney?

stevensJourney:
Indeed these are currently internals.

I do see that devs might want access to these internals in order to build custom database workers - e.g. if custom SQLite functions are desired.

I quite like having the freedom to change these interfaces without making breaking changes. 

If users really have a pressing need for this - one potential workaround might be to mark these exports as `@internal` and/or maybe also expose them separately from the main package exports in `index.ts` with a [subpath export](https://nodejs.org/api/packages.html#subpath-exports). Users of those APIs would need to understand the risk of breaking changes.

LucDeCaf:
@Chriztiaan @stevensJourney I think it would be useful to have access to these functions - right now I just modify the powersync/web folder in node_modules, but that doesn't feel like a very robust solution. My reasoning for why it might make sense to expose these APIs is that the port returned by the public facing shareConnection API proxies into OpenAsyncDatabaseConnection.

Example: this snippet uses only public APIs except for those in AsyncDatabaseConnection:

```ts
const powersync = new PowerSyncDatabase({ /* ... */ });

// Obtain a port for use elsewhere (eg. in a custom worker)
const db = powersync.database; // Will be of type WebDBAdapter unless using SSR
const { port } = await db.shareConnection();
const resolvedOptions = db.getConfiguration();

// my_worker.ts
const remote = Comlink.wrap<OpenAsyncDatabaseConnection>(port);
const connection: AsyncDatabaseConnection = await remote(resolvedOptions);
```

It's worth noting that WebDBAdapter isn't public either, but the method is still publicly available via `PowerSyncDatabase.database`.

If we don't want to expose the API that's fine too, like I said it would just be nice 👍

Chriztiaan:
Hey @LucDeCaf two things from my side:
1. Could you add a a changeset entry for this? Can be `minor`.
2. Could you mark all the members from the newly export file with `@internal`?

Chriztiaan:
Thanks for the suggestion and contribution :)

PR #659, opened 2025-07-10
Improve logging

rkistner:
1. Propagate logger instance from PowerSyncDatabase to the StreamingSyncImplementation and Remote implementations in the various SDKs. This means we now actually use the `logger` passed in the constructor, at the cost of not having a more specific prefix like `[PowerSyncStream]`.
2. Include database name in the default logger, i.e. `PowerSyncDatabase[test.db]`. This helps debugging issues when multiple databases are used.
3. Add some logging for write checkpoints / upload complete.
4. Cleanup for some other log messages.

### DIFF
--- a/.changeset/fresh-laws-join.md
+++ b/.changeset/fresh-laws-join.md
@@ -5,4 +5,4 @@
 '@powersync/web': minor
 ---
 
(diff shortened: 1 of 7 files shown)


stevensJourney (approved):
Nice! LGTM

PR #661, opened 2025-07-11
Dont block on crud uploads

simolus3:
This ports the Rust approach of listening to CRUD upload completions to the JS client as well.

Instead of polling on a promise that resolves on a CRUD upload, we register a listener on completed uploads as a callback. This callback then schedules another round of the main loop handling sync lines, triggering another attempt to apply the checkpoint. This mirrors the "merged streams" approach we have in Dart (and also in the Rust client).

To support this, I had to remove the non-raw methods from `AbstractRemote` - we need custom line mapping to inject the parsed upload complete notifications.

### DIFF
--- a/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
+++ b/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
@@ -687,7 +687,12 @@ The next upload iteration will be delayed.`);
 
       if ('crud_upload_completed' in line) {
         if (validatedCheckpoint != null) {
-          await this.applyCheckpoint(validatedCheckpoint);
+          const { applied, endIteration } = await this.applyCheckpoint(validatedCheckpoint);
+          if (applied) {
+            appliedCheckpoint = validatedCheckpoint;
+          } else if (endIteration) {
+            break;
+          }
         }
 
         continue;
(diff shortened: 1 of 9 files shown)


packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -802,25 +842,7 @@ The next upload iteration will be delayed.`);
...
-            }
-          });
-        } else if (validatedCheckpoint === targetCheckpoint) {
-          const result = await this.applyCheckpoint(targetCheckpoint!, signal);
-          if (result.endIteration) {
-            return;
-          } else if (result.applied) {
-            appliedCheckpoint = targetCheckpoint;
-          }
-        }
```
simolus3, line 823: At this point I'm pretty convinced that this was supposed to do the same thing - apply pending validated checkpoints on crud uploads. It just never did that, because unlike on Dart where we had a merged streams approach possibly calling this, this would only trigger on unknown sync lines before.

I assume this is a leftover from an initial Dart port, looking at how the if/else chain is structured today, I don't see how this could get called.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -639,25 +641,63 @@ The next upload iteration will be delayed.`);
...
       if (!line) {
         // The stream has closed while waiting
         return;
       }
 
+      if ('crud_upload_completed' in line) {
+        if (validatedCheckpoint != null) {
+          const { applied, endIteration } = await this.applyCheckpoint(validatedCheckpoint);
+          if (applied) {
+            appliedCheckpoint = validatedCheckpoint;
```
rkistner: Not important here, but we could remove `appliedCheckpoint` - it does not appear to be read anywhere.

rkistner (approved):

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
simolus3: I've removed that variable now 👍

PR #662, opened 2025-07-11
Remove Old Mutex

stevensJourney:
# Overview

The very first version of the JS SDK used a shared mutex instance for read and write locks. Future versions moved the locking to the `DBAdapter` interface.

The `AbstractPowerSyncDatabase` still contained the shared mutex code which was used in the common `readLock` and `writeLock` implementations. 

Using the same mutex for `readLock` and `writeLock` breaks concurrency. The React native SDK had an override for the `readLock` and `writeLock` method which avoided this issue. Web only has a single connection, so concurrency wasn't an issue there. The Node.js SDK however would rely on the common implementation and would therefore have some concurrency issues. Luckily the individual `get`, `getAll` and `getOptional` methods go through the `DBAdapter` and not the `readLock` API. Currency was only partially broken for the Node.js SDK.

This PR removes the old mutex entirely.

### DIFF
--- a/.changeset/short-countries-swim.md
+++ b/.changeset/short-countries-swim.md
@@ -1,0 +1,6 @@
+---
+'@powersync/common': patch
+'@powersync/node': patch
+---
+
+Fixes an issue where the `readLock` and `writeLock` methods of `AbstractPowerSyncDatabase` shared a common wrapped mutex. This primarily affected the Node.js SDK. The concurrency of individual `get`, `getAll`, and `getOptional` executions was not impacted and functioned correctly prior to this fix.
--- a/packages/common/src/client/AbstractPowerSyncDatabase.ts
+++ b/packages/common/src/client/AbstractPowerSyncDatabase.ts
@@ -671,8 +671,7 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
    * @returns The query result as an object with structured key-value pairs
    */
   async execute(sql: string, parameters?: any[]) {
-    await this.waitForReady();
-    return this.database.execute(sql, parameters);
+    return this.writeLock((tx) => tx.execute(sql, parameters));
   }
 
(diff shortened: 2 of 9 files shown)


rkistner:
What was affected by this?

stevensJourney:
The only affect I can think of is in the Node.js SDK: calls to `readLock` and `writeLock` would be wrapped in a shared mutex. Calling `readLock` in a `writeLock` and vice-versa would hang. Calls would also be executed sequentially.

packages/web/vitest.config.ts
```diff
@@ -31,7 +31,7 @@ const config: UserConfigExport = {
     // Don't optimise these packages as they contain web workers and WASM files.
     // https://github.com/vitejs/vite/issues/11672#issuecomment-1415820673
     exclude: ['@journeyapps/wa-sqlite', '@powersync/web'],
-    include: ['bson', 'comlink']
+    include: ['bson', 'comlink', 'async-mutex']
```
stevensJourney, line 34: Web tests were [failing](https://github.com/powersync-ja/powersync-js/actions/runs/16220517298/job/45801362038) a lot in CI (could not reproduce locally). 

This log was present
```
packages/web test: 1:22:31 PM [vite] ✨ new dependencies optimized: async-mutex
packages/web test: 1:22:31 PM [vite] ✨ optimized dependencies changed. reloading
packages/web test: 
packages/web test: [vitest] Vite unexpectedly reloaded a test. This may cause tests to fail, lead to flaky behaviour or duplicated test runs.
packages/web test: For a stable experience, please add mentioned dependencies to your config's `optimizeDeps.include` field manually.
```

Adding the above might have made the web tests more stable. I verified that the React Supabase Todolist (Vite) app still works without any changes. I don't see the above warnings when building that Vite app.

rkistner (approved):

PR #663, opened 2025-07-11
Experimental: Trigger Based Table Diff Tracking

stevensJourney:
# Overview

We currently expose watched query functionality which allows watching changes of an arbitrary SQLite query's result set. This PR adds methods which can be used to track and process row level changes to **individual** SQLite tables. 

Temporary SQLite triggers and tables are used to populate a set of table change operations. Records for table row `INSERT`, UPDATE` and `DELETE` operations are created, and persisted in a temporary SQLite table, as changes occur. 

## Diff Differences

This API provides a similar result to the diff from `WatchedQuery`s created with `differentialWatch()`, but there are a few key differences to consider.

- `WatchedQuery`s can track the diff of arbitrary query results, while triggers are limited to row level changes of a single table.
- The trigger based diffs record the changes to temporary tables as they are made. This adds some overhead to the WRITE operations. `WatchedQuery`s run off read-connections which have the ability to run concurrently in some platforms.
- Trigger based diffs store the diff as rows in SQLite, allowing SQLite queries to use the diff as an input. `WatchedQuery` diffs expose the diff in JS as a JS objects/arrays.
- `WatchedQuery`s query the SQLite DB on any change to the query's dependent tables, the changes are filtered after querying SQLite. Trigger based diffs can filter/skip storing diff records inside the SQLite trigger, which prevents emissions on a lower level. 
- To avoid race conditions, the trigger based diff tracking requires a `writeLock` during processing (we only allow a single concurrent `writeLock`). 

This PR currently adds 2 new SQLite trigger based methods.

## Creating a Diff Trigger

`createDiffTrigger` Configures temporary SQLite triggers which populate a temporary SQLite table with operations as they occur. 

```javascript
const dispose = await database.triggers.createDiffTrigger({
  source: 'lists',
  destination: tempTable,
  columns: ['name'],
  when: { 
     [DiffTriggerOperation.INSERT]: `TRUE`
     [DiffTriggerOperation.UPDATE]: `TRUE`
  },
});

// logic performs operations on the lists table

// The changes can be fetched in a writeLock

await database.writeLock(async (tx) => {
  const changes = await tx.getAll<TriggerDiffRecord>(/* sql */ `
    SELECT
      *
    FROM
      ${tempTable}
  `);

  // Process changes

  // Clear the temp table after processing
  await tx.execute(/* sql */ ` DELETE FROM ${tempTable}; `);
});
```

## Processing Change Operations

A more convenient `trackTableDiff` method uses `createDiffTrigger` to get table changes, but adds additional logic to consume the changes as they occur. 

The example below uses triggers to process `todo` records as they are inserted/created. 

```javascript
const dispose = await database.triggers.trackTableDiff({
  source: 'todos',
  columns: ['list_id'],
  when: { [DiffTriggerOperation.INSERT]: `json_extract(NEW.data, '$.list_id') = '${firstList.id}'` },
  onChange: async (context) => {
    // Fetches the todo records that were inserted during this diff
    const newTodos = await context.getAll<Database['todos']>(/* sql */ `
      SELECT
        todos.*
      FROM
        DIFF
        JOIN todos ON DIFF.id = todos.id
    `);

    // Process newly added TODOs
  }
});
```

## Docs

For more info see the unit tests and API doc comments.


packages/common/src/client/triggers/TriggerManager.ts
```diff
@@ -0,0 +1,359 @@
...
+   * some basic sanitization, extra external sanitization is recommended.
+   *
+   * @example
+   * {
+   *  'INSERT': whenClause`json_extract(NEW.data, '$.list_id') = ${sanitizeUUID(list.id)}`
+   *  'UPDATE': whenClause`NEW.id = 'abcd' AND json_extract(NEW.data, '$.status') = 'active'`
+   *  'DELETE': whenClause`json_extract(OLD.data, '$.list_id') = 'abcd'`
+   * }
+   */
+  when?: Partial<Record<DiffTriggerOperation, string>>;
```
simolus3: This is really nice 👍

packages/common/src/client/triggers/TriggerManagerImpl.ts
```diff
@@ -0,0 +1,309 @@
...
+            DiffTriggerOperation.DELETE
+          ]} BEGIN
+          INSERT INTO
+            ${destination} (id, operation, timestamp, value)
+          VALUES
+            (
+              OLD.id,
+              'DELETE',
+              strftime ('%Y-%m-%dT%H:%M:%fZ', 'now'),
+              OLD.data
```
simolus3: Should this also apply `jsonFragment`?

simolus3 (commented):
I think it's great to have this functionality! I have some general API notes:

- I see how `trackTableDiff` neatly builds ontop of `createDiffTrigger`, but I'm not sure if the latter needs to be a public API. It sounds to me like it would be easy to misuse by e.g. forgtting to clean up the pending changes after processing them. I can't think of scenarios where the additional flexibility is required, do you have some?
- I'm not that familiar with the JS libraries, but I think we should check if it's possible to use Drizzle / Kysely with this API. I don't think we need dedicated APIs in our packages for those, but it could be worth checking whether we can e.g. build the `WHEN` clause with Kysely APIs. Similarly, can we run the inner query in `trackTableDiff` with those query builders or do they need access to some connection context that would be hard to pass down?

packages/common/src/client/triggers/TriggerManagerImpl.ts
stevensJourney: Good catch. I've fixed this now.

stevensJourney:
> I see how `trackTableDiff` neatly builds ontop of `createDiffTrigger`, but I'm not sure if the latter needs to be a public API. It sounds to me like it would be easy to misuse by e.g. forgtting to clean up the pending changes after processing them. I can't think of scenarios where the additional flexibility is required, do you have some?

That's a fair point. I currently view the `trackTableDiff` method as an automatic "watch" which allows users to process the diffs reactively as they occur. While I can imagine there might be a use-case where the diff should _not_ be processed in "real-time". One example might be a view/form which initiates diff tracking on open, allows for changes to be performed, and only processes the changes when a user-action (like pressing a button) is performed. One could buffer the changes from `trackTableDiff` in memory, but you could also avoid obtaining a writeLock during that period by handling it manually.

> I'm not that familiar with the JS libraries, but I think we should check if it's possible to use Drizzle / Kysely with this API. I don't think we need dedicated APIs in our packages for those, but it could be worth checking whether we can e.g. build the `WHEN` clause with Kysely APIs. Similarly, can we run the inner query in `trackTableDiff` with those query builders or do they need access to some connection context that would be hard to pass down?

I took a brief look-in to this. I think building the `WHEN` clauses with the JS "ORMs" could be tricky due to the JSON structure of the `data` column which the triggers are based off. The ORMS generally are typed according to the views we expose. Currently, the ORM packages we expose "wrap" the PS JS client. It seems like we could expose additional utilities in those ORMs which expose the schema of the `diff` tables - this could help for querying against tables tracked with `createDiffTrigger`. It doesn't seem like it would be easy to directly use a ORM query inside the `onChange` callback of `trackTableDiff`, but I _think_ we might be able to wrap this method inside ORM versions of this API which could be compatible with their query builders. Overall I get the feeling we could release ORM support for triggers as a later release.

packages/node/tests/trigger.test.ts
```diff
@@ -0,0 +1,597 @@
...
+        const result = await tx.execute(
+          /* sql */ `
+            INSERT INTO
+              todos (id, content)
+            VALUES
+              (uuid (), ?) RETURNING id
+          `,
+          [content]
+        );
+        return result.rows?._array?.[0].id;
```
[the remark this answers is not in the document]
stevensJourney: I think the chaining is fine here (test).

packages/node/tests/trigger.test.ts
```diff
@@ -0,0 +1,597 @@
...
+        const result = await tx.execute(
+          /* sql */ `
+            INSERT INTO
+              todos (id, content, columnA, columnB)
+            VALUES
+              (uuid (), ?, ?, ?) RETURNING id
+          `,
+          [content, columnA, columnB]
+        );
+        return result.rows?._array?.[0];
```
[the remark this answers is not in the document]
stevensJourney: I think the chaining is fine here (test).

packages/common/src/client/triggers/TriggerManagerImpl.ts
```diff
@@ -0,0 +1,320 @@
...
+              strftime ('%Y-%m-%dT%H:%M:%fZ', 'now'),
+              ${jsonFragment('OLD')}
+            );
+
+          END;
+        `);
+      }
+    };
+
+    try {
```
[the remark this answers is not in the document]
stevensJourney: Added an AggregateError here

packages/common/src/client/triggers/TriggerManagerImpl.ts
```diff
@@ -0,0 +1,320 @@
...
+        when,
+        hooks
+      });
+
+      return async () => {
+        abortOnChange();
+        await removeTrigger();
+      };
+    } catch (error) {
+      abortOnChange();
```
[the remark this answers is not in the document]
stevensJourney: Added an AggregateError here

packages/common/src/client/triggers/TriggerManager.ts
```diff
@@ -0,0 +1,385 @@
...
+
+  /**
+   * Optional condition to filter when the triggers should fire.
+   * This corresponds to a SQLite [WHEN](https://sqlite.org/lang_createtrigger.html) clause in the trigger body.
+   * This is useful for only triggering on specific conditions.
+   * For example, you can use it to only trigger on certain values in the NEW row.
+   * Note that for PowerSync the row data is stored in a JSON column named `data`.
+   * The row id is available in the `id` column.
+   *
+   * NB! The WHEN clauses here added added directly to the SQLite trigger creation SQL.
```
benitav:
```suggestion
   * NB! The WHEN clauses here are added directly to the SQLite trigger creation SQL.
```

packages/common/src/client/triggers/TriggerManager.ts
```diff
@@ -0,0 +1,385 @@
...
+  hooks?: TriggerCreationHooks;
+}
+
+/**
+ * @experimental
+ * Options for {@link TriggerManager#createDiffTrigger}.
+ */
+export interface CreateDiffTriggerOptions extends BaseCreateDiffTriggerOptions {
+  /**
+   * Destination table to track changes to.
```
benitav:
```suggestion
   * Destination table to send changes to.
```

packages/common/src/client/triggers/TriggerManagerImpl.ts
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    const replicatedColumns = columns ?? sourceDefinition.columns.map((col) => col.name);
+
+    const internalSource = sourceDefinition.internalName;
+
+    const invalidWhenOperations =
+      when && Object.keys(when).filter((operation) => operations.includes(operation as DiffTriggerOperation) == false);
+    if (invalidWhenOperations?.length) {
+      throw new Error(
```
simolus3: I wonder if the API could be a little simpler if we merged `when` and `operations` into a single option, making this check unecessary by enforcing it on a type level, e.g. with something like

```TypeScript
trackTableDiff({
  // ...
  operations: {
    DiffTriggerOperation.INSERT: true, // no when condition
    DiffTriggerOperation.UPDATE: sql`...`,
  }
})
```

But on the other hand, the `true` vs `string` thing here also stands out a bit, so I'm not sure if that actually simplifies things. So treat this as highly optional idea :)

packages/common/src/client/triggers/sanitizeSQL.ts
```diff
@@ -0,0 +1,66 @@
...
+export function sanitizeSQL(strings: TemplateStringsArray, ...values: any[]): string {
+  let result = '';
+  strings.forEach((str, i) => {
+    result += str;
+    if (i < values.length) {
+      // For SQL, escape single quotes in string values
+      const value = values[i];
+      if (typeof value === 'string') {
+        result += sanitizeString(value);
+      } else if (value === null || value === undefined) {
```
simolus3: Looks like we could do `== null` here?

packages/common/src/client/triggers/TriggerManagerImpl.ts
```diff
@@ -0,0 +1,328 @@
...
+        const deleteTriggerId = `ps_temp_trigger_delete_${id}`;
+        triggerIds.push(deleteTriggerId);
+
+        // Create delete trigger for basic JSON
+        await tx.execute(/* sql */ `
+          CREATE TEMP TRIGGER ${deleteTriggerId} AFTER DELETE ON ${internalSource} ${whenConditions[
+            DiffTriggerOperation.DELETE
+          ]} BEGIN
+          INSERT INTO
+            ${destination} (id, operation, timestamp, value)
```
simolus3, line 182: ~Would it make sense to insert into `previous_value` here instead, considering that the new value is essentially undefined after a delete? To me that would feel more consistent with how SQLite handles `OLD` and `NEW`.~ Nevermind, I see this is hard to integrate with `withExtractedDiff`.

packages/common/src/client/triggers/sanitizeSQL.ts
```diff
@@ -0,0 +1,66 @@
...
+ * // Result: "New.id = 'O''Reilly'"
+ * ```
+ *
+ * Avoid manually quoting placeholders:
+ * ```typescript
+ * // Incorrect:
+ * sanitizeSQL`New.id = '${myID}'` // Produces double quotes: New.id = ''O''Reilly''
+ * ```
+ */
+export function sanitizeSQL(strings: TemplateStringsArray, ...values: any[]): string {
```
simolus3, line 45: I think a few independent unit tests asserting exact outputs of this function would be helpful. I see we're using it in `trigger.test.ts`, we could also assert that objects and null values are bound correctly.

packages/common/src/client/triggers/TriggerManagerImpl.ts
stevensJourney: Jip, that was the reason for this :)

packages/common/src/client/triggers/sanitizeSQL.ts
stevensJourney: Jip! Updated this now

packages/common/src/client/triggers/sanitizeSQL.ts
stevensJourney: That's a fair and good idea. I've added some of these tests to the `common` tests now.

packages/common/src/client/triggers/TriggerManagerImpl.ts
stevensJourney: I also mulled this over when writing the initial API. 

Initially I thought there might be some cases where one would want to track all the changes for one operation and only filter changes for another - but this is probably a vary rare use-case - and just specifying `TRUE` for the WHEN clause is a reasonable requirement. 

The comment here is enough to sway me other to a slightly different middle ground. Where we can remove the `operations` option and make `when` required.  Where the WHEN clause can be updated to an explicit `'TRUE'`  if all changes need to be tracked.

```javascript
    await database.triggers.trackTableDiff({
      source: 'lists',
      when: {
        [DiffTriggerOperation.UPDATE]: sanitizeSQL`NEW.id = ${sanitizeUUID(list.id)}`,
        [DiffTriggerOperation.DELETE]: 'TRUE'
      },
```

I've updated the implementation to support the above.

PR #665, opened 2025-07-14
Clear pendingValidatedCheckpoint on new checkpoints

rkistner:
Fix regression from #661. What would happen:
1. "Could not apply checkpoint due to local data. We will retry applying the checkpoint after that upload is completed." -> sets `pendingValidatedCheckpoint` to checkpoint A.
2. Receive a `data`  and `checkpoint_complete` message -> validates checkpoint B.
3. Crud upload complete -> queues a `crud_upload_completed` message.
4. Attempts `this.applyCheckpoint(pendingValidatedCheckpoint)` with checkpoint A. Checkpoint A is no longer valid, so this results in "Checksums failed", and re-downloading the data.

Was not part of a stable release, so no changeset required.


simolus3 (approved):
The implementation changes look good to me, but we'll need someone to check the node logging / testing changes I've added.

rkistner (commented):
Test looks good to me

PR #666, opened 2025-07-15
Improve docs for raw tables

simolus3:
This:

1. Removes a concrete example for raw tables from the changelog - the feature is more complicated than that and requires e.g. custom triggers to detect local writes. So it's better to link to the [upcoming docs page](https://github.com/powersync-ja/powersync-docs/pull/198).
2. Explicitly mentions that the Rust client is required for raw tables.
3. Issues a warning log if the JS client is used with raw tables.

### DIFF
--- a/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
+++ b/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
@@ -623,10 +623,8 @@ The next upload iteration will be delayed.`);
   }
 
   private async legacyStreamingSyncIteration(signal: AbortSignal, resolvedOptions: RequiredPowerSyncConnectionOptions) {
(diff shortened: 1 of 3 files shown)


packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -623,6 +623,12 @@ The next upload iteration will be delayed.`);
   }
 
   private async legacyStreamingSyncIteration(signal: AbortSignal, resolvedOptions: RequiredPowerSyncConnectionOptions) {
+    if (resolvedOptions.serializedSchema) {
```
stevensJourney: I think we could use `resolvedOptions.serizlizedSchema?.raw_tables != null` here instead of the double `if`.

stevensJourney (approved):

PR #668, opened 2025-07-19
Fix crash in React Native: add optional chaining to stream?.close()

felixgourdeau:
This PR fixes a fatal crash (TypeError: Cannot read property 'close' of undefined') occurring in React Native environments.

The issue appears when the app is in the background and a periodic background process triggers. Just before the crash, the last log message is:
`[PowerSyncRemote] No data received on WebSocket in 30000ms, closing connection.`

The root cause is that the stream object can be undefined at this point. Adding optional chaining to the stream.close() call prevents the crash. However, this may mask a deeper issue—ideally, we should investigate why stream is sometimes undefined here.

### DIFF
--- a/packages/common/src/client/sync/stream/AbstractRemote.ts
+++ b/packages/common/src/client/sync/stream/AbstractRemote.ts
@@ -310,7 +310,7 @@ export abstract class AbstractRemote {
       clearTimeout(keepAliveTimeout);
       keepAliveTimeout = setTimeout(() => {
         this.logger.error(`No data received on WebSocket in ${SOCKET_TIMEOUT_MS}ms, closing connection.`);
-        stream.close();
+        stream?.close();
       }, SOCKET_TIMEOUT_MS);
     };
     resetTimeout();


packages/common/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -310,7 +310,7 @@ export abstract class AbstractRemote {
       clearTimeout(keepAliveTimeout);
       keepAliveTimeout = setTimeout(() => {
         this.logger.error(`No data received on WebSocket in ${SOCKET_TIMEOUT_MS}ms, closing connection.`);
-        stream.close();
+        stream?.close();
       }, SOCKET_TIMEOUT_MS);
     };
     resetTimeout();
```
stevensJourney: If I understand the issue correctly, I think the `stream` variable has not been assigned by the time the first timeout happens - perhaps due to the initial connection taking longer than the timeout.

I think we might be able to avoid this by removing the `resetTimeout` call here which starts the timeout process before `connector.connect()` has been called.

felixgourdeau: That would make sense. I've updated the PR

stevensJourney: Thanks for the work here! After digging deeper into this issue, I realized that we should handle a timeout during the initial `connect` call better. I've implemented that timeout functionality here https://github.com/powersync-ja/powersync-js/pull/671
If that PR gets merged we can close this one. Thanks again for for bringing this issue to our attention and for all your efforts.

stevensJourney:
@felixgourdeau Thanks again for your efforts. The work here has been adopted in https://github.com/powersync-ja/powersync-js/pull/671. We'll release the fix soon.

PR #681, opened 2025-07-31
fix(adapter-sql-js): Updated sql-js upstream version to support rust sync implementation

Chriztiaan:
Small dependency bump for adapter-sql-js package.
See upstream change https://github.com/powersync-ja/powersync-sql-js/pull/5

### DIFF
--- a/.changeset/fair-buttons-deliver.md
+++ b/.changeset/fair-buttons-deliver.md
@@ -1,0 +1,5 @@
+---
+'@powersync/adapter-sql-js': patch
+---
+
+Updated sql-js upstream version to support rust sync implementation.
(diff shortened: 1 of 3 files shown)


rkistner:
So if I understand correctly, the issue is that we're passing an ArrayBuffer, while upstream sql-js only supports Uint8Array?

This probably comes up because we're explicitly converting from Uint8Array to ArrayBuffer [here](https://github.com/powersync-ja/powersync-js/blob/47a230291e5313c64fa51713757b64c48293ceed/packages/react-native/src/sync/bucket/ReactNativeBucketStorageAdapter.ts#L5-L8), and now we're effectively converting back to Uint8Array. I wonder if we should modify that check, to keep the `Uint8Array` as-is for anything other than RNQS? It likely also affects our op-sqlite implementation.

PR #682, opened 2025-08-01
[PoC] Direct op-sqlite connections

rkistner:
This exposes an additional `OPSqliteOpenFactory.openDirectConnection()` method, to get direct access to an op-sqlite DB.

This is similar to opening another op-sqlite DB directly, but:
1. Uses the same sqlite options directly from the `OPSqliteOpenFactory`.
2. Auto-loads the powersync extension.
3. Propagates any updates in that connection to the PowerSyncDatabase, to trigger watch queries and crud uploads.

We do this rather than exposing the existing DB, since that would bypass any existing transaction locking mechanisms, leading to likely race conditions when the DB is used directly.

One specific advantage of this is allowing synchronous queries to be executed, to make it easier to migrate existing applications currently built on synchronous queries. I would not recommend this for other use cases.

Even with this approach, synchronous writes should be wrapped with `PowerSyncDatabase.writeLock()`, otherwise it could fail with `SQLITE_BUSY` if another transaction is busy. We cannot really use `BUSY_TIMEOUT` for this, since the synchronous query would block the JS thread, preventing the other lock from being released.

Synchronous reads do not have the same restriction, and can run concurrently with any other transaction on the PowerSyncDatabase.

## Other changes

1. This makes the number of read connections configurable.
2. This changes write transactions to use `BEGIN EXCLUSIVE` instead of `BEGIN`: Same as in the web SDK. This reduces the number of lock steps, and avoids triggering SQLITE_BUSY errors later in the transaction.

## TODOs

 * [ ] This currently builds on many private APIs of `OPSQLiteDBAdapter` - this can probably be done better.
 * [ ] Needs testing - I haven't actually tested this at all yet.

### DIFF
--- a/packages/powersync-op-sqlite/src/db/OPSQLiteConnection.ts
+++ b/packages/powersync-op-sqlite/src/db/OPSQLiteConnection.ts
@@ -58,6 +58,10 @@ export class OPSQLiteConnection extends BaseObserver<DBAdapterListener> {
     });
   }
 
+  hasUpdates(): boolean {
+    return this.updateBuffer.length > 0;
+  }
+
   flushUpdates() {
     if (!this.updateBuffer.length) {
       return;
--- a/packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts
+++ b/packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts
@@ -1,4 +1,4 @@
-import { ANDROID_DATABASE_PATH, getDylibPath, IOS_LIBRARY_PATH, open, type DB } from '@op-engineering/op-sqlite';
+import { getDylibPath, open, type DB } from '@op-engineering/op-sqlite';
 import { BaseObserver, DBAdapter, DBAdapterListener, DBLockOptions, QueryResult, Transaction } from '@powersync/common';
 import Lock from 'async-lock';
 import { Platform } from 'react-native';
@@ -158,6 +158,10 @@ export class OPSQLiteDBAdapter extends BaseObserver<DBAdapterListener> implement
 
   async readLock<T>(fn: (tx: OPSQLiteConnection) => Promise<T>, options?: DBLockOptions): Promise<T> {
     await this.initialized;
+    if (this.readConnections?.length == 0) {
+      // When opened with no read connections, use the write connection for reads.
+      return this.writeLock(fn, options);
+    }
     return new Promise(async (resolve, reject) => {
       const execute = async () => {
         // Find an available connection that is not busy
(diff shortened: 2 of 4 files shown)


a11rew:
+1 on this being added. Appreciate it!

My use-case for this would be replaying write notifications to the database that occurred in another connection so the OPSqliteConnection the UI and watched queries listen on gets updated. Access to the connection would enable calling `addTableUpdate` on the UI connection directly

rkistner:
@a11rew Could you expand on your specific use case? Specifically, where would you like to do (1) writes, (2) reads and (3) watched queries, between the `PowerSyncDatabase`, op-sqlite connection, or another connection?

We probably won't go ahead with this PR in the current form, since it doesn't solve any actual use case I'm aware of: synchronous writes on the op-sqlite connection have issues regarding write locks (see description above), and synchronous reads can be done directly using op-sqlite, without the changes here (see https://github.com/powersync-community/powersync-react-native-synchronous-operations).

a11rew:
@rkistner use case here is reflecting changes made to the underlying database outside the Powersync connection in the UI. i.e. if a change is made to the sqlite db in a connection outside of the one Powersync establishes, watched queries should still update. 

In my specific case:
1) Writes are done in native iOS code to the same SQLite DB Powersync creates, this allows the app to respond to [App Intents](https://developer.apple.com/documentation/appintents) without the JS runtime being available
2) Reads are done in the JS runtime when the app is foregrounded. They're done through Powersync.
3) The app uses watched queries to display data in some views. Since [sqlite3_update_hook](https://sqlite.org/c3ref/update_hook.html) and subsequently op-sqlite's update hook functionality work on a per-connection basis, the connection Powersync is listening on doesn't receive update notifications. 

What we're currently doing to work around this is registering a sqlite3_update_hook on that other sqlite connection and then calling [tablesUpdated](https://github.com/powersync-ja/powersync-js/blob/03831dc3099838518fdbb146243838f8204145eb/packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts#L86-L88) on all the listeners registered on the op-sqlite connection the app is using for UI. This also means we've had do things like re-implement the buffering and flushing [OPSQLiteConnection](https://github.com/powersync-ja/powersync-js/blob/03831dc3099838518fdbb146243838f8204145eb/packages/powersync-op-sqlite/src/db/OPSQLiteConnection.ts#L54) does. Direct access to the/a connection instance would make the ergonomics of working with the connection better.

rkistner:
Thanks, that makes sense. So you're making database updates in native iOS code, and need those to trigger update notifications / watched queries in the PowerSyncDatabase.

I think this PR as-is isn't quite suitable for that, since opening another op-sqlite connection wouldn't help you. But adding a way to trigger those update notifications directly on the PowerSyncDatabase instance would help.

PR #686, opened 2025-08-04
Improve Exports

stevensJourney:
# Overview

The PowerSync Node.js SDK currently provides both ESM and CommonJS exports for compatibility with different environments. However, some related packages such as the Drizzle and Kysely drivers export ESM JavaScript without explicitly declaring the package as an ES module.

While this isn't an issue for bundled applications, it causes problems when the packages are imported directly into a Node.js application. In such cases, Node.js assumes the packages use CommonJS and fails to load them due to the presence of ESM syntax like `import` and `export`.

For additional context. See https://github.com/microsoft/TypeScript/issues/60663

This PR updates the affected packages to explicitly export both ESM and CommonJS (now added via Rollup). As with other packages in this repo - modules are compiled with TypeScript by targeting ES. Rollup is used to compile a CommonJS export which is declared in the `package.json` file. 

A small test suite has also been added to verify that the packages load correctly in both environments.

A test for checking package exports via the [Are The Types Wrong CLI](https://www.npmjs.com/package/@arethetypeswrong/cli) has been added for relevant packages. Most packages have been updated with tests. Testing focus has been applied to packages which might be used in a Node.js environment. React native specific packages are not modified in this PR.

## Package Changes Summary

- `@powersync/attachments`: Added a CJS export which is generated via Rollup.
- `@powersync/common`: Added CJS Type declarations.
- `@powersync/drizzle-driver`:  Added a CJS export which is generated via Rollup.
- `@powersync/kysely-driver`:  Added a CJS export which is generated via Rollup. 
- `@powersync/node`: Added CJS Type declarations
- `@powersync/react`: Declared the module as a `module`. We currently only provide ES exports for this package.
- `@powersync/vue`: Declared the module as a `module`. We currently only provide ES exports for this package.
- `@powersync/tanstack-react-query`: Declared the module as a `module`. We currently only provide ES exports for this package.


packages/node/tsconfig.json
```diff
@@ -13,7 +13,8 @@
     "preserveConstEnums": true,
     "esModuleInterop": true,
     "skipLibCheck": false,
-    "strictNullChecks": true
+    "strictNullChecks": true,
+    "tsBuildInfoFile": "./lib/tsconfig.tsbuildinfo"
```
rkistner: Is this just to make the `pnpm clean` command remove it automatically? It seems like this would include the file in the built package, which we probably don't want?

stevensJourney: Yeah, it makes it a little easier to clean the project (including manually). I have added an entry to the `npmignore` file to avoid it being included in the published package. I still need to test that with a dev package release though.

packages/drizzle-driver/tests/sqlite/watch.test.ts
```diff
@@ -158,9 +158,9 @@ describe('Watch Tests', () => {
       await db
         .insert(assets)
         .values({
-          id: sql`uuid()`,
+          id: sql`uuid ()`,
```
rkistner, line 161: Why do these get the additional spaces now? It is valid syntax, but a little unusual

stevensJourney: The prettier formatter seems to apply these changes

packages/node/tsconfig.json
rkistner: Isn't this just increasing the amount of config then, since it's now added in two places (tsconfig.json and npmignore), instead of only in the clean command?

packages/drizzle-driver/tests/sqlite/watch.test.ts
rkistner: Ok, it's a known issue for "user-defined functions": https://github.com/sql-formatter-org/sql-formatter/issues/444

packages/node/tsconfig.json
stevensJourney: I was hoping that the npmignore entry could be specified once in the monorepo root. But, after testing I noticed the buildinfo file was still being included in the published files. I've removed this change.

rkistner (approved):
This is great - the arethetypeswrong tool and docs gives by far the best explanation I've seen of the issues and what needs to be done to fix them.

Chriztiaan (approved):

Issue #687, opened 2025-08-07
releaseHold is not a function (it is undefined)


mtford90:
Seeing this error on hot reload in react native. 

it seems to be happening here: 

https://github.com/powersync-ja/powersync-js/blob/7f2c53d573e9f7137db6245db799102233092c44/packages/react/src/hooks/suspense/useWatchedQuerySuspenseSubscription.ts#L47-L51

my current workaround is to update the node_module, adding an if statement, but perhaps that will be introducing some kind of memory leak

```
if (!query.state.isLoading && releaseHold) {
      releaseHold();
}
```

## versions

```
"@powersync/kysely-driver": "1.2.0",
"@powersync/react-native": "1.23.1",
"react-native": "0.79.5",
"expo": "53.0.10"
```

## screenshots

<img width="1790" height="3372" alt="Image" src="https://github.com/user-attachments/assets/d4ec7520-bffc-4f0d-bdd7-7611a1e7d0a3" />

PR #688, opened 2025-08-07
React 19 Fixes

stevensJourney:
# Overview

Closes https://github.com/powersync-ja/powersync-js/issues/687

I have only been able to reproduce the issue above when using React 19 in Strict mode. 

It seems like the flow is as follows.

- Initially `useWatchedQuerySuspenseSubscription` is called. We create a temporary hold and suspend the component during loading.
- After loading is complete React calls `useWatchedQuerySuspenseSubscription` which calls `useTemporaryHold` - since the query is not loading, we don't need to create a temporary hold. We return a no-op `releaseHold` method (we don't assign it to the `releaseTemporaryHold` Ref since it's a no-op).
-  A third invocation of `useWatchedQuerySuspenseSubscription` is eventually triggered, we get the value Ref value of `releaseTemporaryHold` (which has not been assigned). We then enter the useEffect in `useWatchedQuerySuspenseSubscription` and attempt to release using the undefined method.

Generally the current implementation is unnecessarily flawed. We cannot feasibly remove the temporary hold from `useWatchedQuerySuspenseSubscription`. We have to rely on the polling to remove this.

There are primarily 2 scenarios

**If the WatchedQuery is loading on the first render:** 

During the initial render (while the query is loading). We will detect `isLoading == true` and throw the pending promise which triggers suspense. Throwing this promise will not trigger the `useEffect` in this "render".

Once the promise resolves React's suspense will reinvoke `useWatchedQuerySuspenseSubscription` (without any previous state). Calling `useTemporaryHold` will not create a temporary hold since the WatchedQuery is no longer in the loading state. The `useEffect` will run, but releasing would be a no-op.

**If the supplied WatchedQuery was already loaded on the first render:**

We will not suspend or create any temporary hold. There is nothing to release in the `useWatchedQuerySuspenseSubscription::useEffect`.

I'm not entirely sure why React 19 Strict mode triggered the third invocation of `useWatchedQuerySuspenseSubscription` where the `useEffect` was triggered, but removing this unnecessary check avoids this issue entirely. 

The `should suspend on initial load` test ensures that temporary holds are removed correctly. 

I manually verified that the unit tests pass when using React 19. This required manually extracting the unit tests from the Monorepo. We should eventually update React in this monorepo.

### DIFF
--- a/packages/react/src/hooks/suspense/suspense-utils.ts
+++ b/packages/react/src/hooks/suspense/suspense-utils.ts
@@ -1,33 +1,28 @@
 import { WatchedQuery } from '@powersync/common';
 import React from 'react';
 
-const NO_OP_RELEASE = () => {};
-
 /**
  * The store will dispose this query if it has no subscribers attached to it.
  * The suspense promise adds a subscriber to the query, but the promise could resolve
  * before this component is committed. The promise will release it's listener once the query is no longer loading.
  * This temporary hold is used to ensure that the query is not disposed in the interim.
  * Creates a subscription for state change which creates a temporary hold on the query
- * @returns a function to release the hold
  */
 export const useTemporaryHold = (watchedQuery?: WatchedQuery<unknown>) => {
   // Defaults to a no-op. If the provided WatchedQuery is not loading, we don't need a
   // temporary hold.
-  const releaseTemporaryHold = React.useRef<() => void>(NO_OP_RELEASE);
+  const releaseTemporaryHold = React.useRef<() => void | undefined>(undefined);
   const addedHoldTo = React.useRef<WatchedQuery<unknown> | undefined>(undefined);
 
   if (addedHoldTo.current !== watchedQuery) {
     // The query changed, we no longer need the previous hold if present
     releaseTemporaryHold.current?.();
-    releaseTemporaryHold.current = NO_OP_RELEASE;
+    releaseTemporaryHold.current = undefined;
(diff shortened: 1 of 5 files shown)


packages/react/src/hooks/suspense/suspense-utils.ts
```diff
@@ -7,23 +7,23 @@ import React from 'react';
...
  * This temporary hold is used to ensure that the query is not disposed in the interim.
  * Creates a subscription for state change which creates a temporary hold on the query
- * @returns a function to release the hold
  */
 export const useTemporaryHold = (watchedQuery?: WatchedQuery<unknown>) => {
-  const releaseTemporaryHold = React.useRef<(() => void) | undefined>(undefined);
+  const releaseTemporaryHold = React.useRef<() => void | undefined>(undefined);
   const addedHoldTo = React.useRef<WatchedQuery<unknown> | undefined>(undefined);
 
   if (addedHoldTo.current !== watchedQuery) {
```
stevensJourney, line 15: One might be able to argue this check is not necessary either. We generally create the temporary hold before suspending, so we can't really keep track of it if I understand correctly.

stevensJourney: I'll leave this as an optimization/cleanup for later. We do need to avoid creating a temporary hold during normal operations, so we need some check - but we probably can make this method thinner in future.

PR #690, opened 2025-08-08
Node: Support using node:sqlite driver

simolus3:
This adds work-in-progress support for using `node:sqlite3` as a driver in addition to `better-sqlite3`. Because there are no update hooks in that driver, this requires https://github.com/powersync-ja/powersync-sqlite-core/pull/120 to be released first.

Most tests seem to work, but there are some failures I still need to investigate. We also need to figure out how to import `node:sqlite` properly, we can't use a regular import because `node:sqlite` is not available on older Node versions we still want to support. A dynamic import seems to work for ESM, but rollup doesn't properly translate it to a `require` call.

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -69,6 +69,7 @@
   "devDependencies": {
     "@powersync/drizzle-driver": "workspace:*",
     "@types/async-lock": "^1.4.0",
+    "@types/node": "^24.2.0",
     "drizzle-orm": "^0.35.2",
     "rollup": "4.14.3",
     "typescript": "^5.5.3",
--- a/packages/node/src/db/AsyncDatabase.ts
+++ b/packages/node/src/db/AsyncDatabase.ts
@@ -1,4 +1,5 @@
 import { QueryResult } from '@powersync/common';
+import { NodeDatabaseImplementation } from './options.js';
 
 export type ProxiedQueryResult = Omit<QueryResult, 'rows'> & {
   rows?: {
@@ -7,8 +8,14 @@ export type ProxiedQueryResult = Omit<QueryResult, 'rows'> & {
   };
 };
 
+export interface AsyncDatabaseOpenOptions {
+  path: string;
+  isWriter: boolean;
+  implementation: NodeDatabaseImplementation;
+}
+
 export interface AsyncDatabaseOpener {
-  open(path: string, isWriter: boolean): Promise<AsyncDatabase>;
+  open(options: AsyncDatabaseOpenOptions): Promise<AsyncDatabase>;
 }
 
 export interface AsyncDatabase {
--- a/packages/node/src/db/BetterSQLite3DBAdapter.ts
+++ b/packages/node/src/db/BetterSQLite3DBAdapter.ts
@@ -120,7 +120,7 @@ export class BetterSQLite3DBAdapter extends BaseObserver<DBAdapterListener> impl
       const database = (await comlink.open({
         path: dbFilePath,
         isWriter,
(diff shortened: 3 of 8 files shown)


rkistner:
This looks quite promising!

If this works well, we could perhaps switch to node:sqlite by default when we do a v1.0 release, moving better-sqlite3 support to a separate package or import (removing it from the default dependencies).

I'm also wondering: Could we use the same approach for update notifications for better-sqlite3? Would that allow us to switch back to upstream better-sqlite3, or are there other blockers currently?

simolus3:
> Could we use the same approach for update notifications for better-sqlite3

Yes! I've refactored the connection setup to share that between `node:sqlite` and `better-sqlite3`.

> Would that allow us to switch back to upstream better-sqlite3, or are there other blockers currently?

That seems to have been the only blocker, I've migrated the package back to `better-sqlite3`.

I've also expanded the worker API to allow custom `better-sqlite3` forks, the demo now supports encryption thanks to this.

> we could perhaps switch to node:sqlite by default when we do a v1.0 release

To prepare for that, I've made `better-sqlite3` an optional peer dependency. That is a breaking change, because:

- users now need to import that package.
- we only import `better-sqlite3` with a dynamic `import()` / `require()` call when it's actually used (to avoid having it as a dependency when users want to use `node:sqlite`). This breaks some bundler setups, but I've described the issue in the readme and updated the examples.

Still, I think it's worth it because we need this for encryption / to later support `node:sqlite`.

.changeset/wet-cooks-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/node': minor
+---
+
+Use upstream better-sqlite3 dependency instead of the PowerSync fork.
```
stevensJourney, line 5: It seems like this could be viewed as a breaking change - which is fine, we should just communicate it well. Maybe we could add a little more information here: in order to help existing users upgrade this package. 
I'm not sure what is best, we could just mention to run:
```bash
npm uninstall @powersync/better-sqlite3
npm install better-sqlite3
```

packages/node/README.md
```diff
@@ -15,12 +15,12 @@ See a summary of features [here](https://docs.powersync.com/client-sdk-reference
...
 
-# Installation
+## Installation
 
-## Install Package
+### Install Package
 
 ```bash
-npm install @powersync/node
+npm install @powersync/node better-sqlite3
```
stevensJourney, line 23: We should also update this in the docs https://docs.powersync.com/client-sdk-references/node#quickstart

stevensJourney (approved):
This looks amazing. I'm very happy with the changes. I left some minor comments for the changelog entry and README - however, the implementation looks perfect to me.

packages/node/README.md
simolus3: Good point, docs PR is here: https://github.com/powersync-ja/powersync-docs/pull/273

.changeset/wet-cooks-dress.md
simolus3: > It seems like this could be viewed as a breaking change

If I'm not misremembering semver on npm again, it is also marked as such. Since we're going from `0.11.x` to `0.12.x` with this, users depending on `^0.11.0` would not get upgraded automatically.

I agree that we shoud mention the commands needed for an upgrade though, I've added them to this changeset entry.

stevensJourney: That's correct regarding the version numbering. IMO mentioning it in the changeset wording makes it even clearer - version zero is often a tricky rule to remember. The new wording looks good to me 😄

PR #691, opened 2025-08-08
[test] Capacitor Investigation Additions

stevensJourney:
# Overview

The purpose of this PR is mainly to attempt to reproduce and debug the issue from https://discord.com/channels/1138230179878154300/1401770441911177276

Additional logging has been added to:

- Log any uncaught exceptions in the DB worker to a logger in the main JS context.
- Log VFS debug logs - this is in the hopes that this might expose any additional details if the issue is reproducible. I'm not sure yet if we'd want to expose this as part of the API.

Our capacitor demo is rather simple. We don't have any external logging or error reporting services. I've configured a custom `Logger` to store all logs and make them presentable in the application. This is somewhat of a diagnostics view in the application. The purpose of this is mainly to make debugging possible, but it could be a nice general addition (perhaps without persisting the logs).

I've added a local only table and widgets to modify the data to the Capacitor demo. This is mainly since the Capacitor demo does not have a full fledged upload connector. For testing it would be nice to perform SQLite operations. Again, this might be a nice general addition to the demo. 

Generally this also improves the layout and responsiveness of the capacitor demo.

Screenshots for those who are interested:

image
image

### DIFF
--- a/.changeset/neat-crabs-lie.md
+++ b/.changeset/neat-crabs-lie.md
@@ -1,0 +1,5 @@
+---
+'@powersync/web': patch
+---
+
+Testing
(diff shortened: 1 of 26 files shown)


stevensJourney:
Closing this since the Capacitor SDK has been released

PR #693, opened 2025-08-14
Add `getCrudTransactions()` returning an async generator

simolus3:
Sometimes, users want a batch of data to upload while also ensuring that that batch respects transaction boundaries. At the moment, we don't really have APIs for that, because:

- `getCrudBatch()` returns a batch of items with a flexible size limit, but doesn't respect transaction boundaries (so it's possible to accidentally upload half of a transaction).
- `getNextCrudTransaction()` returns a complete transaction, but nothing more. It will also always return the oldest transaction, so you can't call `getNextCrudTransaction()` multiple times to get more than one.

This adds `getCrudTransactions()`, an async iterable that starts with the oldest transaction and then keeps on yielding more transactions until the iterator is closed. This gives users a tool to compose batches with flexible sizes while still ensuring everything is grouped by transactions.

### DIFF
--- a/packages/common/src/client/AbstractPowerSyncDatabase.ts
+++ b/packages/common/src/client/AbstractPowerSyncDatabase.ts
@@ -632,35 +632,72 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
    * @returns A transaction of CRUD operations to upload, or null if there are none
    */
   async getNextCrudTransaction(): Promise<CrudTransaction | null> {
-    return await this.readTransaction(async (tx) => {
-      const first = await tx.getOptional<CrudEntryJSON>(
-        `SELECT id, tx_id, data FROM ${PSInternalTable.CRUD} ORDER BY id ASC LIMIT 1`
-      );
+    for await (const transaction of this.getCrudTransactions()) {
+      return transaction;
+    }
 
-      if (!first) {
-        return null;
+    return null;
+  }
+
+  /**
+   * Returns an async iterator of completed transactions with local writes against the database.
+   *
+   * This is typically used from the {@link PowerSyncBackendConnector.uploadData} callback. Each entry emitted by the
+   * returned flow is a full transaction containing all local writes made while that transaction was active.
+   *
+   * Unlike {@link getNextCrudTransaction}, which always returns the oldest transaction that hasn't been
+   * {@link CrudTransaction.complete}d yet, this flow can be used to collect multiple transactions. Calling
+   * {@link CrudTransaction.complete} will mark _all_ transactions emitted by the flow until that point as completed.
+   *
+   * This can be used to upload multiple transactions in a single batch, e.g with:
+   *
+   * ```TypeScript
+   * let lastTransaction: CrudTransaction | null = null;
+   * let batch: CrudEntry[] = [];
+   *
+   * for await (const transaction of database.getCrudTransactions()) {
+   *   batch.push(...transaction.crud);
+   *   lastTransaction = transaction;
+   *
+   *   if (batch.length > 10) {
+   *     break;
+   *    }
+   * }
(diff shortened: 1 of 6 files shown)


packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -632,35 +632,72 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
...
+   * This is typically used from the {@link PowerSyncBackendConnector.uploadData} callback. Each entry emitted by the
+   * returned flow is a full transaction containing all local writes made while that transaction was active.
+   *
+   * Unlike {@link getNextCrudTransaction}, which always returns the oldest transaction that hasn't been
+   * {@link CrudTransaction.complete}d yet, this flow can be used to collect multiple transactions. Calling
+   * {@link CrudTransaction.complete} will mark _all_ transactions emitted by the flow until that point as completed.
+   *
+   * This can be used to upload multiple transactions in a single batch, e.g with:
+   *
+   * ```TypeScript
```
stevensJourney: Nit: If I remember correctly, we try and target JavaScript, as the lowest common denominator, for example snipplets in docs.

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -632,35 +632,72 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
    * @returns A transaction of CRUD operations to upload, or null if there are none
    */
   async getNextCrudTransaction(): Promise<CrudTransaction | null> {
-    return await this.readTransaction(async (tx) => {
-      const first = await tx.getOptional<CrudEntryJSON>(
-        `SELECT id, tx_id, data FROM ${PSInternalTable.CRUD} ORDER BY id ASC LIMIT 1`
-      );
+    for await (const transaction of this.getCrudTransactions()) {
```
stevensJourney: We need to be cautious about making this change. React Native used to (and might still) have issues with Async Iterator support.

This update will cause the current `getNextCrudTransaction` method to rely on Async iterators - which might break some users' apps if they have not installed the Async iterator [polyfils](https://github.com/powersync-ja/powersync-js/tree/main/packages/react-native#install-polyfills).

I haven't checked in a while if recent versions of RN support Async Iterators, but even if they do this change might affect users on older versions of RN/Expo.

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -632,35 +632,72 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
...
+   *   batch.push(...transaction.crud);
+   *   lastTransaction = transaction;
+   *
+   *   if (batch.length > 10) {
+   *     break;
+   *    }
+   * }
+   * ```
+   *
+   * If there is no local data to upload, the async iterator complete without emitting any items.
```
stevensJourney, line 665: Related to React Native and Async Iterators: We might want to update the docs comments for the RN version of this method to mention Async Iterator usage.

packages/common/src/client/AbstractPowerSyncDatabase.ts
simolus3: I have desugared the `for await` here by hand to avoid this issue.

packages/common/src/client/AbstractPowerSyncDatabase.ts
simolus3: I've added a sentence to the docs here, and I've also refactored the implementation to not use async generators as a language feature anymore.

Removing the babel plugin from the supabase demo seems to work with local changes calling `getNextCrudTransaction()`.

stevensJourney (approved):

PR #697, opened 2025-08-20
Avoid calling hooks conditionally

simolus3:
In `useQuery`, we used to invoke the inner `useWatchedQuery` or `useSingleQuery` hooks depending on options passed to the `useQuery` hook.

This violates the [rules of hooks](https://react.dev/reference/rules/rules-of-hooks), because they must be invoked unconditionally. Mixing them like this can trip up the internal state of `useEffect` hooks and others, causing all kinds of issues that are hard to debug.

So to avoid this, we add an `active: boolean` parameter to the internal hooks to conditionally disable them. Then, we can just call both hooks all the time since the inactive one won't actually do anything. I've added a test that fails with the old implementation but works with the new one.

### DIFF
--- a/packages/react/src/hooks/watched/useWatchedQuery.ts
+++ b/packages/react/src/hooks/watched/useWatchedQuery.ts
@@ -1,5 +1,5 @@
 import React from 'react';
-import { useWatchedQuerySubscription } from './useWatchedQuerySubscription.js';
+import { useNullableWatchedQuerySubscription } from './useWatchedQuerySubscription.js';
 import { DifferentialHookOptions, QueryResult, ReadonlyQueryResult } from './watch-types.js';
 import { InternalHookOptions } from './watch-utils.js';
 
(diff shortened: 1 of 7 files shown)


packages/react/src/hooks/watched/useWatchedQuery.ts
```diff
@@ -14,9 +14,13 @@ import { InternalHookOptions } from './watch-utils.js';
 export const useWatchedQuery = <RowType = unknown>(
   options: InternalHookOptions<RowType[]> & { options: DifferentialHookOptions<RowType> }
 ): QueryResult<RowType> | ReadonlyQueryResult<RowType> => {
-  const { query, powerSync, queryChanged, options: hookOptions } = options;
+  const { query, powerSync, queryChanged, options: hookOptions, active } = options;
+
+  function createWatchedQuery() {
```
simolus3, line 19: Note: I had to remove the `useCallback` here because the inner function would otherwise use `active` and `hookOptions` from the time the hooks was called initially (which I think was also an issue before if you would update `powersync` and `hookOptions` at once - the `useEffect` below would run, but the function would still use old values).

There isn't really a reason to cache this function anyway since it's not used as a dependency of another hook. So evaluating it is cheap.

PR #700, opened 2025-08-28
Fix: Rust Client Disconnect Race

stevensJourney:
# Overview

When disconnecting and connecting frequently, it is currently possible for the Rust sync clients to get locked in a frozen state if the `StreamingSyncClient` is disconnected/aborted before it has connected.

This can happen if:
- The client is triggered for a connect
- A disconnect is requested shortly after
- The `StreamingSyncImplementation`'s `AbortController` is marked as aborted, but we nest the `AbortController` [here](https://github.com/powersync-ja/powersync-js/blob/2f6153c95c9b24e5dca6213c3654e55eb8577cff/packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts#L905). 
- For WebSocket requests, we usually check if the provided `AbortSignal` has been aborted before making the request (and we throw), but due to the wrapped controller used in the Rust implementation we don't detect that the wrapped controller has been aborted. The wrapped controller will also never be aborted since the upstream controller won't emit the `aborted` event again (since it was aborted before)
- This causes the Rust client to remain connected. The `ConnectionManager`'s `disconnectingPromise` never resolves, which causes a shared sync implementation to be frozen.

This checks the upstream AbortSignal before wrapping it. The abort is propagated by throwing a `AbortOperation`. 

This could also have happened with the HTTP connection method. A check has been added there also.

### DIFF
--- a/.changeset/honest-penguins-worry.md
+++ b/.changeset/honest-penguins-worry.md
@@ -1,0 +1,5 @@
+---
+'@powersync/common': patch
+---
+
+Fix issue where Rust sync implementation might not disconnect in some circumstances.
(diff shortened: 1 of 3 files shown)


simolus3 (approved):
Nice catch, thanks!

PR #703, opened 2025-09-02
fix: useQuery() not reacting to query/param changes in StrictMode

Chriztiaan:
The `useWatchedQuery`'s useEffect responsible for updating the watchedQuery's settings on query change never triggered in [React.StrictMode](https://react.dev/reference/react/StrictMode). 

Basically the `queryChanged` prop changes from `false` => `true` => `false` before the useEffect would pick up the change. So that "query change" event was consumed without doing anything, I have introduced a ref approach to protect against this from happening. 

For good measure I have opted to change the `useQuery` test suite to run the tests in both normal and Strict mode, if we like this pattern we can apply it to the other React package's tests.

### DIFF
--- a/packages/react/src/hooks/watched/useWatchedQuery.ts
+++ b/packages/react/src/hooks/watched/useWatchedQuery.ts
@@ -16,6 +16,8 @@ export const useWatchedQuery = <RowType = unknown>(
 ): QueryResult<RowType> | ReadonlyQueryResult<RowType> => {
   const { query, powerSync, queryChanged, options: hookOptions, active } = options;
 
+  // This ref is used to protect against cases where `queryChanged` changes multiple times too quickly to be
+  // picked up by the useEffect below. This typically happens when React.StrictMode is enabled.
   const queryChangeRef = React.useRef(false);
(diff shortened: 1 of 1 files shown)


packages/react/src/hooks/watched/useWatchedQuery.ts
```diff
@@ -16,6 +16,11 @@ export const useWatchedQuery = <RowType = unknown>(
 ): QueryResult<RowType> | ReadonlyQueryResult<RowType> => {
   const { query, powerSync, queryChanged, options: hookOptions, active } = options;
 
+  const queryChangeRef = React.useRef(false);
```
simolus3, line 21: Perhaps add a comment here pointing out that we're introducing this to avoid reacting to multiple query changes concurrently, and mention that this is relevant for strict mode causing `queryChanged` to re-run the hook multiple times?

simolus3 (approved):
The implementation and test changes look good to me 👍

simolus3 (approved):

stevensJourney (approved):
Looks good to me :)

PR #704, opened 2025-09-02
Diagnostics app: Support Rust client

simolus3:
Since we'll likely want to show information about sync streams in the diagnostics app, we need that tool to support the Rust client.
This requires some hacks because the diagnostic app displays information not exposed from the SDKs, such as:

- Involved buckets.
- Progress information (okay, that is exposed - but not on a per-bucket granularity).
- Metadata and payload data sizes.

For the JS client, we've used a custom bucket storage implementation that intercepts some operations to update metadata. For the Rust client, we can't use the exact same approach because the internal bucket storage is written in Rust and not patchable from JS. However, we can intercept the `powersync_control` calls to infer the sync lines, and then use that to derive progress information from there.

The Rust client is enabled by default, but this can be changed in the configuration:

grafik

After connecting, the pages show whether the Rust or JS client is used ('ve added a new field to `SyncStatus` for this):

grafik


stevensJourney (approved):
Happy with the changes here from my side :)

PR #707, opened 2025-09-04
Sync streams

simolus3:
This ports sync streams (Dart PR: https://github.com/powersync-ja/powersync.dart/pull/317) to the JavaScript SDK.

Subscriptions are requested by calling `db.syncStream('stream_name', params).subscribe()`. The `subscribe()` call returns a subscription handle which can be used to `unsubscribe()` the stream. As an additional measure to ensure streams don't leak, we also add a finalization registry on streams to call `unsubscribe()` implicitly.

When a stream is subscribed to while offline, we don't really have to do much in the SDK. Most of the logic is implemented in the core extension, which will remember the subscription (so that, if we're connecting later but within the TTL, the stream is included even if the subscription no longer exists).
Because all streams (even those that have never been resolved - relevant for introspection) are included in the sync status, we need to update the sync status for new subscriptions even while offline. The new `powersync_offline_sync_status()` funciton can be used for that, it also replaces the query on `ps_sync_state` used during initialization.

To report the current status of stream subscriptions, we can simply take them from the JSON provided by the core extension.

While connected, the core extension needs to know all streams that have a subscription active. This information is provided in two ways:

1. When `connect()` is called, we provide a snapshot of all streams that are currently active.
2. When that list changes while connected, we call `updateSubscriptions()` with the new set of stream subscriptions. The core extension is informed about this change, which has two effects:
    - When subscriptions are removed, their expiry date is no longer periodically updated.
    - When a subscription is added, or if the expiry date of an inactive subscription expires, the core extension requests the sync iteration to restart. The next iteration would then request the new actual set of stream subscriptions.

Within a `PowerSyncDatabase` instance, it is possible to subscribe to the same stream / params multiple times. The subscriptions we pass to the core extension is a de-duplicated set of all active subscription instances (managed with a refcount).

With a shared sync worker, the same concept applies across different tabs as well: Each tabs sends its de-duplicated set of active subscriptions to the worker, which then applies its own de-duplication logic before notifying the core extension.
The shared worker also keeps track of which tabs are owning which subcriptions. That allows it to update the set when a tab is closing (in case some subscriptions were only active in a single tab).

TODO:

- [x] (probably best in a separate PR): Use the weblocks hack to detect closing tabs from the shared worker.
- [x] Test different subscriptions across tabs

I will add hooks for subscriptions in a follow-up PR.

### DIFF
--- a/.changeset/angry-ducks-sneeze.md
+++ b/.changeset/angry-ducks-sneeze.md
@@ -1,0 +1,7 @@
+---
+'@powersync/react-native': minor
+'@powersync/common': minor
+'@powersync/web': minor
+---
+
+Add alpha support for sync streams, allowing different sets of data to be synced dynamically.
--- a/packages/node/tests/utils.ts
+++ b/packages/node/tests/utils.ts
@@ -68,6 +68,8 @@ export async function createDatabase(
     database: {
       dbFilename: 'test.db',
       dbLocation: tmpdir,
+      // Using a single read worker (instead of multiple, the default) seems to improve the reliability of tests in GH
+      // actions. So far, we've not been able to reproduce these failures locally.
       readWorkerCount: 1
     },
     logger: defaultLogger,


rkistner:
I still need to review properly, but this part stood out to me:

> As an additional measure to ensure streams don't leak, we also add a finalization registry on streams to call unsubscribe() implicitly.

I wonder if this could break some common use cases? For example, a developer may just want to subscribe to relevant streams when loading a page:

```js
async function loadData(id) {
  const subscription = db.syncStream('my_stream', {id}).subscribe();
  await subscription.waitForFirstSync();
}
```

However, the subscription is garbage collected, which would trigger the implicit unsubscribe logic.

Of course, we don't do that, and the user uses a long-running SPA, the subscription could stay alive forever, long after the user navigated to a different page.

So it feels like overall for an SPA, the developer will need to keep track of these subscriptions somewhere. I'm just not sure what the best behavior is if they don't - both unsubscribing automatically and not doing that has potential for issues. Maybe instead just logging a warning when a subscription is garbage collected without an explicit unsubscribe? We'll also need some guides on how to handle this with some common frameworks.

simolus3:
> Maybe instead just logging a warning when a subscription is garbage collected without an explicit unsubscribe?

That sounds good to me, I've changed the finalizer to a warning.

> We'll also need some guides on how to handle this with some common frameworks.

I assume that at least in React, we'd have a `useSyncStream` instead of manual `subscribe()` and `unsubscribe()` calls.

packages/common/src/client/sync/bucket/BucketStorageAdapter.ts
```diff
@@ -49,6 +50,7 @@ export interface BucketChecksum {
    * Count of operations - informational only.
    */
   count?: number;
+  subscriptions?: any;
```
stevensJourney, line 58: Should this have a type?

packages/common/src/client/sync/stream/core-instruction.ts
```diff
@@ -53,3 +69,29 @@ export interface BucketProgress {
...
+
+export function coreStatusToJs(status: CoreSyncStatus): sync_status.SyncStatusOptions {
+  const coreCompleteSync = status.priority_status.find((s) => s.priority == FULL_SYNC_PRIORITY);
+  const completeSync = coreCompleteSync != null ? priorityToJs(coreCompleteSync) : null;
+
+  return {
+    connected: status.connected,
+    connecting: status.connecting,
+    dataFlow: {
+      downloading: status.downloading != null,
```
stevensJourney, line 91: Is it possible that `status.downloading` will ever be `false` and converted to `true` here? Adding a comment here would help make this a little more readable.

packages/web/src/db/sync/SharedWebStreamingSyncImplementation.ts
```diff
@@ -179,6 +190,19 @@ export class SharedWebStreamingSyncImplementation extends WebStreamingSyncImplem
      * This performs bi-directional method calling.
      */
     Comlink.expose(this.clientProvider, this.messagePort);
+
+    // Request a random lock until this client is disposed. The name of the lock is sent to the shared worker, which
+    // will also attempt to acquire it. Since the lock is returned when the tab is closed, this allows the share worker
+    // to free resources associated with this tab.
```
stevensJourney, line 196: This is very nice

stevensJourney (approved):
I'm quite happy with the changes from my side. I could not spot any concerns or potential issues. This looks good to me :)

packages/common/src/client/sync/bucket/BucketStorageAdapter.ts
simolus3: It's only used for tests (so that we can pass instances of `BucketChecksum` to the Rust client), so ignoring the type is easier. I've added a comment explaining the reason.

packages/common/src/client/sync/stream/core-instruction.ts
simolus3: I've added a comment, `status.downloading` has a `DownloadProgress | null` (decoded from the core extension) so it can't be `false`.

packages/common/src/client/sync/stream/core-instruction.ts
```diff
@@ -53,3 +69,31 @@ export interface BucketProgress {
 export interface FetchCredentials {
   did_expire: boolean;
 }
+
+function priorityToJs(status: SyncPriorityStatus): sync_status.SyncPriorityStatus {
+  return {
+    priority: status.priority,
+    hasSynced: status.has_synced ?? undefined,
+    lastSyncedAt: status?.last_synced_at != null ? new Date(status!.last_synced_at! * 1000) : undefined
```
stevensJourney: Nit: I think we can remove some of the operators here?

```suggestion
    lastSyncedAt: status.last_synced_at != null ? new Date(status.last_synced_at! * 1000) : undefined
```

.changeset/angry-ducks-sneeze.md
```diff
@@ -0,0 +1,7 @@
+---
+'@powersync/react-native': minor
+'@powersync/common': minor
+'@powersync/web': minor
```
stevensJourney, line 4: Not technically required, but would be nice to include `node` here also

simolus3: Will do that on the hooks PR

PR #708, opened 2025-09-05
[POC] Capacitor Community SQLite

stevensJourney:
# Overview

Theoretically, one should be able to use the PowerSync Web SDK in Capacitor Webviews. Recent support items have shown that while this works in practice, the experience is not very stable in some circumstances.

This uses https://github.com/capacitor-community/sqlite as the SQLite driver for PowerSync - with the main aim to target iOS and Android platforms (only iOS has been tested so far).

The Capacitor Community SQLite package currently poses some challenges for PowerSync's use-case:
- Extension loading is not supported. There are APIs for loading an extension, but attempting to use those on iOS results in 
 `Unhandled Promise Rejection: Error: "CapacitorSQLite.loadExtension()" is not implemented on ios`
- Update hooks don't seem to be implemented
- The methods for SQL execution are split between execute and query operations. The PowerSync JS clients currently allow running queries which return results from the `execute` methods.

We can work around most of these limitations.

**Extension Loading**
This PR adds a Capacitor Plugin which includes the PowerSync Rust core into a project. We can then register a static SQLite auto extension. We use the low level C APIs for this. This currently works for iOS. Android needs to be investigated.

**Update Hooks**
A future version of the PowerSync core could support internal table update reporting mechagnisms.

**Query execution differences**
Perhaps this could be well documented for Capacitor

# TODOS

- [ ] The Example Capacitor currently uses this plugin by using a `CapacitorSQLiteAdapter` for the `DBAdapter` provided to the PowerSync Web SDK. While this does work, it seems messy. We should probably introduce a dedicated Capacitor PowerSyncDatabase constructor.
- [ ] Android support
- [ ] Testing
- [ ] Tests


simolus3:
>  A future version of the PowerSync core could support internal table update reporting mechagnisms.

The future is now! With version 0.4.5 of the core extension, you can:

1. When the write connection is initialized, run `SELECT powersync_update_hooks('install')` ([Kotlin for reference](https://github.com/powersync-ja/powersync-kotlin/blob/aa4e7bc10f8dc2c39a7823be8bfcb371f5dd6353/core/src/commonMain/kotlin/com/powersync/db/driver/InternalConnectionPool.kt#L60-L63)).
2. After a transaction may have completed (you can likely put this at the end of `writeLock`), call `SELECT powersync_update_hooks('get') as r`. This will return a json array of all tables with a commited write since the last `get` invocation ([Kotlin for reference](https://github.com/powersync-ja/powersync-kotlin/blob/aa4e7bc10f8dc2c39a7823be8bfcb371f5dd6353/core/src/commonMain/kotlin/com/powersync/db/driver/InternalConnectionPool.kt#L76-L86)).

packages/capacitor/src/adapter/CapacitorSQLiteAdapter.ts
```diff
@@ -0,0 +1,245 @@
...
+    );
+
+    return {
+      rowsAffected: result.changes?.changes ?? 0,
+      insertId: result.changes?.lastId
+    };
+  }
+
+  /**
+   * We're not using separate read/write locks here because we can't implement connection pools on top of SQL.js.
```
Chriztiaan: Is this also true for the Capacitor sqlite package?

stevensJourney: It's not exactly true, but still somewhat relevant. It seems like Capacitor, as other SQLite libs, tracks the low-level SQLite connection in a map to the DB file name. This usually means that multiple connections to the same DB are not possible. However, it seems like Capacitor kind of works around this by prefixing the DB name in the map with a READ/WRITE prefix - so it seems we might be able to get one read and one write connection. I still need to investigate how the query APIs (including their transaction APIs) map to the underlaying connections.

packages/capacitor/README.md
```diff
@@ -0,0 +1,81 @@
...
+- **Electron**: Uses WASQLite via the PowerSync Web SDK.
+
+## Limitations
+
+- Encryption for native mobile platforms is not yet supported.
+- Multiple tab support is not available for native Android and iOS targets.
+
+## Examples
+
+See the [`demos/example-capacitor/`](https://github.com/journeyapps/powersync-react-native-sdk/blob/capacitor-sdk/demos/example-capacitor/README.md#L1) directory for a working example.
```
stevensJourney: TODO: Need to update this before merging

demos/example-capacitor/src/library/powersync/AppSchema.ts
```diff
@@ -1,13 +1,18 @@
...
+    created_at: column.text
+  },
+  {
+    viewName: 'customers'
+  }
+);
 
 export const AppSchema = new Schema({
-  customers
+  lists: customers
```
stevensJourney: TODO: Need to revert this. This makes testing easier for me :D

stevensJourney:
Some basic benchmark comparisons were ran on an Android Emulator + iPhone Simulator.
A basic test for executing 1000 `INSERT` statements (1000 `.execute(...) calls`) was conducted. The time take for the INSERT operations to complete was recorded.
```javascript
const now = performance.now();
for (let i = 0; i < 1_000; i++) {
  await powerSync.execute('INSERT INTO customers (id, name) VALUES (uuid(), ?)', [`Customer`]);
}
const end = performance.now();
```
| Platform | Capacitor SQLite (ms) | AccessHandle OPFS (ms) | COOP OPFS (ms) | IndexedDB (ms) |
|----------|----------------------|-----------------------|----------------|----------------|
| **iOS**     | 864                  | 895                   | 982            | 4,948          |
| **Android** | 3,129                | 3,931                 | 3,431          | 10,890         |

For general mobile comparisons, see https://www.powersync.com/blog/react-native-database-performance-comparison

packages/capacitor/src/plugin/web.ts
```diff
@@ -0,0 +1,9 @@
+import { WebPlugin } from '@capacitor/core';
+
+import type { PowerSyncPlugin, RegistrationResponse } from './PowerSyncPlugin';
+
+export class PowerSyncWeb extends WebPlugin implements PowerSyncPlugin {
+  async registerCore(): Promise<RegistrationResponse> {
+    throw new Error('This code path is not supported on web.');
```
simolus3, line 7: Should this perhaps be a noop since `@powersync/web` already works and this doesn't need to do anything? Or does running this always indicate an error in the configuration?

simolus3 (approved):
Cool to see `powersync_update_hooks` in action :)

packages/capacitor/src/plugin/web.ts
stevensJourney: > Or does running this always indicate an error in the configuration?
This is the case. This plugin should not be used on Web. The WA-SQLite implementation should be used before we reach this point (on web we should never reach here)

Chriztiaan (approved):

PR #717, opened 2025-09-15
React hooks for sync streams

simolus3:
Blocked on the parent PR: https://github.com/powersync-ja/powersync-js/pull/707.

This updates the `react` package to expose hooks for sync streams:

- The `useSyncStream()` hook has been added. It takes name and parameters of the stream, along with optional subscription options. When called, it:
  - creates a stream subscription on the requested stream
  - returns the current subscription state of that stream subscription
  - unsubscribes when the component is unmounted
- The `useQuery()` hook has been updated to optionally take an array of streams as well.
  - Similarly to `useSyncStream()`, it subscribes to those streams while the component is active.
  - Additionally, it delays the queries until the subscriptions have been synced at least once. This is a neat shortcut to display a loading state while data is downloading / querying from the database and then transition to data once it's available.

I've also fixed a few minor issues with the sync stream implementation in this PR.


packages/react/src/hooks/streams.ts
```diff
@@ -0,0 +1,124 @@
...
+          subscription = sub;
+          setSubscription(sub);
+        } else {
+          sub.unsubscribe();
+        }
+      });
+
+    return () => {
+      active = false;
+      subscription?.unsubscribe();
```
stevensJourney, line 64: I havent tested this: It might be safer to locally store a reference to the Promise returned by `db.syncStream(..).subscribe(..)` and use that in the disposer function.

I'm particularly wondering about a potential case where the component:
- mounts and triggers the `subscribe()` function.
- unmounts before the promise resolves
- the `subscription` state is null, so it doesn't unsubscribe
- The subscription is not automatically disposed.

The same might also happen in React Strict mode, where the useEffect is called twice in development. The disposal there could likely happen before the promise resolved.

stevensJourney: I see the `active` boolean will actually handle this case above. This is a neat solution. Maybe add a comment about this in the disposer to potentially help future readers understand.

packages/react/src/hooks/watched/useQuery.ts
```diff
@@ -58,15 +59,20 @@ export function useQuery<RowType = any>(
...
   if (!powerSync) {
-    return { isLoading: false, isFetching: false, data: [], error: new Error('PowerSync not configured.') };
+    return {
+      ..._loadingState,
+      isLoading: false,
+      error: new Error('PowerSync not configured.')
+    };
   }
   const { parsedQuery, queryChanged } = constructCompatibleQuery(query, parameters, options);
+  const streamsHaveSynced = useAllSyncStreamsHaveSynced(powerSync, options?.streams);
```
stevensJourney, line 69: Maybe this is a bit of a meta question:

It looks like the flow here is that the `useQuery` will remain loading until the streams have been synced. Which means the query could remain loading indefinitely while offline (if the first boot is offline or it went offline before completely syncing)?

This might make the `useQuery` tricky to use for offline cases in those scenarios. E.g. if:
- This `useQuery` is watching the todos of a todolist based off a subscription. 
- If we display `isLoading=true` while syncing, users most likely will default to a loading screen during this time. 
- Local todo creations are blocked until we've synced (unless if the app does some ignoring of the isLoading state).

In some scenarios that's perfectly fine. But maybe we'd need to make this more configurable. Maybe instead of returning the `isLoading` state, we could report an additional `streamsHaveSynced` result? Or make the blocking nature optional in the provided `options`

packages/react/src/hooks/streams.ts
simolus3: I've added comments.

packages/react/src/hooks/watched/useQuery.ts
simolus3: > Maybe instead of returning the `isLoading` state, we could report an additional `streamsHaveSynced` result?

To me it feels like an `isLoading` state is quite natural for this, because we are currently loading data from the network (well, at least in an online-first mindset. If we're offline there would indeed be an infinite loading spinner whereas an error message might be more suitable for online-first apps).

> Or make the blocking nature optional in the provided `options`

I like that suggestion 👍  I've made the blocking behavior opt-in which sounds it should avoid surprises.

PR #719, opened 2025-09-17
feat: add end to end encrypted (E2EE) Todo app demo

marcus-pousette:
This demo contains a E2EE encrypted TODO app. 

This app has separated from the encryption layer so it can be repurposed for other apps that needs E2EE. 

The encryption works by creating a decrypted mirror locally that using the 

```
query.differentialWatch({...}).registerListener({ onDiff: () => {...})))
``` 

Project is using mono repo for subpackage management, but could potentially also be swapped out with a yarn workspace solution


kobiebotha:
moved to https://github.com/powersync-community/react-supabase-chat-e2ee

PR #725, opened 2025-09-22
Fix deadlock issue with OPFS and Rust client

simolus3:
With the OPFS VFS enabled, each tab uses its own dedicated worker to host the database. The shared sync worker will proxy its requests through a connection on one of these tabs.

When the tab currently hosting the database for the shared worker is closed, the worker calls `disconnect()` and is supposed to pick another tab. That works well with the JS sync implementation, but for the Rust implementation, a `disconnect()` call requires a `SELECT powersync_control('stop');` invocation. Since the tab is already gone, and Comlink is not handling this case, all database promises will just never resolve.
Worse, since the database from the tab is wrapped in an `LockedAsyncDatabaseAdapter`, the sync worker holds a global lock on the database while waiting on a dead promise. This pretty much brings everything to a halt.

As a solution, this PR makes `WorkerWrappedAsyncDatabaseConnection` aware of the fact that its remote may close. We're only using this when a shared worker requests a connection from a tab, in other cases where a tab owns a dedicated worker this can't happen.
When removing a port, the shared worker will set a flag on the worker-wrapped connection that will reject all outstanding promises and deny new requests. This can cause an exception in the shared worker, but it reliably recovers from that by reconnecting with a different tab.

To reproduce the issue, apply the following path to adopt OPFS and the Rust client in the react supabase demo:

```patch
diff --git a/demos/react-supabase-todolist/src/components/providers/SystemProvider.tsx b/demos/react-supabase-todolist/src/components/providers/SystemProvider.tsx
index 8a3f3c20..b3521c00 100644
--- a/demos/react-supabase-todolist/src/components/providers/SystemProvider.tsx
+++ b/demos/react-supabase-todolist/src/components/providers/SystemProvider.tsx
@@ -3,7 +3,15 @@ import { AppSchema, ListRecord, LISTS_TABLE, TODOS_TABLE } from '@/library/power
 import { SupabaseConnector } from '@/library/powersync/SupabaseConnector';
 import { CircularProgress } from '@mui/material';
 import { PowerSyncContext } from '@powersync/react';
-import { createBaseLogger, DifferentialWatchedQuery, LogLevel, PowerSyncDatabase } from '@powersync/web';
+import {
+  createBaseLogger,
+  DifferentialWatchedQuery,
+  LogLevel,
+  PowerSyncDatabase,
+  SyncClientImplementation,
+  WASQLiteOpenFactory,
+  WASQLiteVFS
+} from '@powersync/web';
 import React, { Suspense } from 'react';
 import { NavigationPanelContextProvider } from '../navigation/NavigationPanelContext';
 
@@ -12,9 +20,13 @@ export const useSupabase = () => React.useContext(SupabaseContext);
 
   schema: AppSchema,
-  database: {
-    dbFilename: 'example.db'
-  }
+    dbFilename: 'example.db',
+    vfs: WASQLiteVFS.OPFSCoopSyncVFS,
+    flags: {
+      enableMultiTabs: typeof SharedWorker !== 'undefined'
+    }
+  })
 });
 
 export type EnhancedListRecord = ListRecord & { total_tasks: number; completed_tasks: number };
@@ -68,7 +80,7 @@ export const SystemProvider = ({ children }: { children: React.ReactNode }) => {
     const l = connector.registerListener({
       initialized: () => {},
       sessionStarted: () => {
-        powerSync.connect(connector);
+        powerSync.connect(connector, { clientImplementation: SyncClientImplementation.RUST });
       }
     });
```

Open the demo app, login and open two additional tabs of the demo app. I'm then using a Chrome extension to reload the first two tabs. These tabs will run into a locked database and will never connect.

### DIFF
--- a/packages/web/src/db/adapters/WorkerWrappedAsyncDatabaseConnection.ts
+++ b/packages/web/src/db/adapters/WorkerWrappedAsyncDatabaseConnection.ts
@@ -60,7 +60,7 @@ export class WorkerWrappedAsyncDatabaseConnection<Config extends ResolvedWebSQLO
     this.notifyRemoteClosed!.abort();
   }
 
-  private withRemote<T>(inner: () => Promise<T>): Promise<T> {
+  private withRemote<T>(workerPromise: () => Promise<T>): Promise<T> {
     const controller = this.notifyRemoteClosed;
     if (controller) {
       return new Promise((resolve, reject) => {
@@ -72,20 +72,20 @@ export class WorkerWrappedAsyncDatabaseConnection<Config extends ResolvedWebSQLO
           reject(new Error('Remote peer closed with request in flight'));
         }
 
-        function markResolved(inner: () => void) {
+        function completePromise(action: () => void) {
           controller!.signal.removeEventListener('abort', handleAbort);
-          inner();
+          action();
         }
 
         controller.signal.addEventListener('abort', handleAbort);
 
-        inner()
-          .then((data) => markResolved(() => resolve(data)))
-          .catch((e) => markResolved(() => reject(e)));
+        workerPromise()
+          .then((data) => completePromise(() => resolve(data)))
+          .catch((e) => completePromise(() => reject(e)));
       });
     } else {
-      // Can't close, so just return the inner promise unguarded.
-      return inner();
+      // Can't close, so just return the inner worker promise unguarded.
+      return workerPromise();
     }
   }
 


packages/web/src/db/adapters/WorkerWrappedAsyncDatabaseConnection.ts
```diff
@@ -43,6 +47,48 @@ export class WorkerWrappedAsyncDatabaseConnection<Config extends ResolvedWebSQLO
...
+      return new Promise((resolve, reject) => {
+        if (controller.signal.aborted) {
+          reject(new Error('Called operation on closed remote'));
+        }
+
+        function handleAbort() {
+          reject(new Error('Remote peer closed with request in flight'));
+        }
+
+        function markResolved(inner: () => void) {
```
stevensJourney: Maybe rename the `inner` param here. It was a little confusing reading where there are two potential `inner` variables.

PR #726, opened 2025-09-24
feat: expose connection options

khawarizmus:
Exposing the connector would simplify the [registration of the diagnostics in the PowerSync Nuxt module](https://powersync-nuxt.netlify.app/guide/getting-started.html#connect-to-powersync) by removing the need to use `shareConnectorWithInspector(nuxtApp, connector)` function which registers the connector globally for Nuxt to access.

Exposing the connection options would also allow the diagnostics to directly peek at the options with having to [ask the user to hard code them](https://powersync-nuxt.netlify.app/guide/getting-started.html#setup-config) which will not work when the user dynamically changes these especially the connection parameters

### DIFF
--- a/tools/diagnostics-app/CHANGELOG.md
+++ b/tools/diagnostics-app/CHANGELOG.md
@@ -1,5 +1,11 @@
 # diagnostics-app
 
+## 0.9.12
+
+### Patch Changes
+
+- 0a6af1c: Added support for a "token" query parameter, when supplied to the root URL it will be used as the API token for the diagnostics tool instead of needing to be provided manually to the login form.
+
 ## 0.9.11
 
 ### Patch Changes
--- a/tools/diagnostics-app/package.json
+++ b/tools/diagnostics-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@powersync/diagnostics-app",
-  "version": "0.9.11",
+  "version": "0.9.12",
   "private": true,
   "scripts": {
     "dev": "vite",
--- a/tools/diagnostics-app/src/app/page.tsx
+++ b/tools/diagnostics-app/src/app/page.tsx
@@ -1,14 +1,33 @@
 import React from 'react';
 import { CircularProgress, Grid, styled } from '@mui/material';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useSearchParams } from 'react-router-dom';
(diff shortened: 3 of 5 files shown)


packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -592,6 +594,15 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
...
   }
 
+  /**
+   * Get the connection options used to connect to the PowerSync backend instance.
+   *
+   * @returns The connection options used to connect to the PowerSync backend instance.
+   */
+  getConnectionOptions(): PowerSyncConnectionOptions | null {
+    return this.connectionOptions ?? null;
+  }
```
simolus3: I don't really see the purpose of this function considering that we'd already have a public field then?

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -514,6 +515,7 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
    */
   async connect(connector: PowerSyncBackendConnector, options?: PowerSyncConnectionOptions) {
     const resolvedOptions: InternalConnectionOptions = options ?? {};
+    this.connectionOptions = resolvedOptions;
```
simolus3: I think the connection manager also stores the current option, so implementing the field as a getter referring to that may be more reliable as storing it again.

khawarizmus: Exposing the connection manager might actually be better since we can get access to the connector as well but i don't see it exposing the configuration options. I will see how i can expose both.

packages/common/src/client/AbstractPowerSyncDatabase.ts
```diff
@@ -193,6 +193,14 @@ export abstract class AbstractPowerSyncDatabase extends BaseObserver<PowerSyncDB
     return this.connectionManager.syncStreamImplementation;
   }
 
+  get connector() {
```
simolus3, line 201: Let's add a documentation comment here and on `connectionOptions` as well, perhaps something like "The current connector / resolved connection options, or `null` if `connect()` has not been called on this database yet."

PR #731, opened 2025-10-02
Rename diagnostics app to be more specific

benitav:
i.e. it's a tool for sync diagnostics


simolus3 (approved):

simolus3:
I think you should also add a changeset entry for the app so that these changes can get published.

PR #733, opened 2025-10-03
[Node.js] Bundle PowerSync Core

stevensJourney:
# Overview

We currently download the relevant binary for the PowerSync rust core as an `install` script of `@powersync/node`. Some environments, such as PNPM require explicitly approving install scripts, which can result in cases where the extension is not downloaded.

This bundles the PowerSync extension for all architectures in the published NPM package - avoiding the need for install scripts.

The extension is also downloaded in local scripts to ease local development.

### DIFF
--- a/.changeset/seven-fireants-boil.md
+++ b/.changeset/seven-fireants-boil.md
@@ -1,0 +1,5 @@
+---
+'@powersync/node': minor
+---
+
+Bundle PowerSync extension with NPM package
--- a/packages/node/src/db/SqliteWorker.ts
+++ b/packages/node/src/db/SqliteWorker.ts
@@ -1,9 +1,9 @@
-import * as path from 'node:path';
 import BetterSQLite3Database, { Database } from '@powersync/better-sqlite3';
 import * as Comlink from 'comlink';
-import { parentPort, threadId } from 'node:worker_threads';
 import OS from 'node:os';
+import * as path from 'node:path';
 import url from 'node:url';
+import { parentPort, threadId } from 'node:worker_threads';
 import { AsyncDatabase, AsyncDatabaseOpener } from './AsyncDatabase.js';
 
 class BlockingAsyncDatabase implements AsyncDatabase {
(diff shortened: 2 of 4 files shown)


packages/node/src/db/SqliteWorker.ts
```diff
@@ -23,28 +22,55 @@ export interface PowerSyncWorkerOptions {
...
+export function getPowerSyncExtensionFilename() {
+  const platform = OS.platform();
+  const arch = OS.arch();
+  let extensionFile: string;
+
+  if (platform == 'win32') {
+    if (arch == 'x64') {
+      extensionFile = 'powersync_x64.dll';
+    } else {
+      throw new Error('Windows platform only supports x64 architecture.');
```
simolus3: Can we also add an `arm64` branch here? Starting from version 0.4.6 of the core extension, we can use `powersync_aarch64.dll` for this.

Adding an empty `powersync_aarch64.dll` entry in the hashes at `download_core.js` and then re-generating hashes should be enough to ensure we have that file.

.changeset/seven-fireants-boil.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/node': minor
+---
+
+Pre-package all the PowerSync Rust extension binaries for all supported platforms in the NPM package. Install scripts are no longer required to download the PowerSync core.
```
simolus3: Given the changes we had to make to the examples (and I know some clients have the same logic on their side), should we perhaps also add a link to the necessary changes to this entry? After all, filenames have changed and so existing code could stop working.

simolus3 (commented):
This is a nice improvement over the old install scripts, I really like seeing that setup code be removed in the examples 👍

.changeset/seven-fireants-boil.md
stevensJourney: Good point! I've added a note and small diff for the required changes in the changelog entry.

packages/node/src/db/SqliteWorker.ts
stevensJourney: Nice! I've added the `arm64` and `x86` Windows builds here. 
I noticed we also had some additional Linux binaries. I added those also. I did not add support for `armv7`, it seems like detecting that platform requires additional work - which I cannot test to verify.

simolus3 (approved):
Nice 🚀

packages/node/src/db/SqliteWorker.ts
simolus3: Yeah I don't think anyone would have a good time running modern Node on `armv7` anyway, they'll have other problems :D

Issue #715, opened 2025-09-11
@powersync/attachments StorageAdapter method parameters are a bit limited


hunterm-wbc:
As a user, when I'm implementing my own `StorageAdapter`, I find the parameters defined by the interface to be a bit limiting. Specifically, the `uploadFile`, `downloadFile` and `deleteFile` methods. In order to interface with my own backend, I need more information than just the filename or uri, I specifically need some of the information from the attachment record, including the additional columns I've defined.

For my own use case, I've resorted to stuffing any information I need to process the file to the filename of the record, which works - however this seems a bit hacky.

Is there any possibility of having this interface expanded to provide pass alone the attachment record?

stevensJourney:
That's a fair point. We could, and should pass the entire attachment record to those methods. The more recent attachment helpers in other SDKs do this. We'll keep this issue open till the enhancements have been made :)

Issue #714, opened 2025-09-11
@powersync/attachments saveToQueue doesn't save additionalColumns


hunterm-wbc:
The attachments package lets us define `additionalColumns` when we define our attachment schema.

When we save an attachment in the `AbstractAttachmentQueue` using the `saveToQueue` method, passing along the any additional columns as part of the attachmentRecord, it does not save these additional columns, I think this is a bit unexpected.

A work around I think will work is to take the record returned by `saveToQueue` and perform an update, however this is a bit inefficient, and I think it would be clearer to save any additional columns provided to `saveToQueue`

stevensJourney:
My understanding of the approach taken here is that the `saveToQueue` method (and potentially other persistence methods) would need to be overridden if additional columns are used. I agree that the DX for this is not amazing. In the subsequent implementations of attachment helpers in other SDKs we've removed the ability to specify arbitrary additional columns - replacing it with a single JSON column for `metadata`.

hunterm-wbc:
Hi Steven - the metadata approach makes sense. I ended up moving the fields to my record associated to the attachment, and it works decent. If the javascript SDK ever gets the metadata, I'd probably switch to that approach. Thanks for the response!

PR #735, opened 2025-10-07
Update attachment package

joshua-journey-apps:
This a port of the Swift, Dart and Kotlin attachment's API to the JS SDK. 

This addresses the following issues #715 and #714 by adding the `meta_data` column to the table and persisting it in the attachments table.

It is currently still a work-in-progress so there are a couple things that still needs to be done:

- [x] Move attachment package to common
  - [x] Move Node adapter under Node package
  - [x] Move indexDB adapter under Web package
  - [x] Move Expo adapter under RN package
- [x] Update demos
- [x] Improve testing
- [x] Update the README
- [x] Implement sync throttling
- [x] Complete clearing archived attachments
- [x] Removing some holdovers from the previous implementation
  - [x] Supported file data format

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,13 +25,6 @@
     "test": "pnpm run -r --workspace-concurrency=1 test --run",
     "test:packages:exports": "pnpm -r --filter {./packages/**} test:exports"
   },
-  "pnpm": {
-    "onlyBuiltDependencies": [
-      "@powersync/better-sqlite3",
-      "better-sqlite3",
-      "better-sqlite3-multiple-ciphers"
-    ]
-  },
   "keywords": [],
   "type": "module",
   "author": "JOURNEYAPPS",


packages/attachments/src/AttachmentContext.ts
```diff
@@ -0,0 +1,159 @@
...
+  tableName: string;
+  logger: ILogger;
+
+  constructor(db: AbstractPowerSyncDatabase, tableName: string = 'attachments', logger: ILogger) {
+    this.db = db;
+    this.tableName = tableName;
+    this.logger = logger;
+  }
+
+  watchActiveAttachments(onUpdate: () => void): AbortController {
```
stevensJourney: The `AttachmentContext` here is typically an exclusive access context (provided by the AttachmentsService). The context provides general methods for reading and writing AttachmentRecords in a safe context.
The `watchActiveAttachments` method does fall in to the category of reading Attachment records, but it does not typically require being executed in an exclusive context. It's usually only called once when the sync starts for initialisation. Furthermore, the handler in `onUpdate` typically will request an `AttachmentContext` in order to process the active attachments, which can produce a rather interesting code flow.
For consistency with other SDKs, I'd recommend moving this to the `AttachmentService`.

`onUpdate` should probably be an async handler.

Alternatively, we could use the Differentially Watched queries here. That would have the advantage of only triggering when there are actual changes to the result set. It also could make closing/aborting and listening for changes simpler. 

```typescript
watchActiveAttachments(): DifferentialWatchedQuery<AttachmentRecord> 
```

packages/common/src/attachments/AttachmentContext.ts
```diff
@@ -0,0 +1,159 @@
+import { AbstractPowerSyncDatabase, ILogger, Transaction } from '@powersync/common';
+import { AttachmentRecord, AttachmentState, attachmentFromSql } from './Schema.js';
+
+export class AttachmentContext {
```
stevensJourney, line 14: Just a reminder to add doc comments to all public classes and interfaces

packages/attachments/src/StorageService.ts
```diff
@@ -0,0 +1,157 @@
+import { ILogger } from '@powersync/common';
+import { AttachmentContext } from './AttachmentContext.js';
+import { EncodingType, LocalStorageAdapter } from './LocalStorageAdapter.js';
+import { RemoteStorageAdapter } from './RemoteStorageAdapter.js';
+import { AttachmentRecord, AttachmentState } from './Schema.js';
+import { SyncErrorHandler } from './SyncErrorHandler.js';
+
+export class StorageService {
```
stevensJourney: For consistency, this should be called `SyncingService`

packages/common/src/attachments/SyncingService.ts
```diff
@@ -0,0 +1,157 @@
+import { ILogger } from '@powersync/common';
+import { AttachmentContext } from './AttachmentContext.js';
+import { EncodingType, LocalStorageAdapter } from './LocalStorageAdapter.js';
+import { RemoteStorageAdapter } from './RemoteStorageAdapter.js';
+import { AttachmentRecord, AttachmentState } from './Schema.js';
+import { SyncErrorHandler } from './SyncErrorHandler.js';
+
+export class StorageService {
+  context: AttachmentContext;
```
stevensJourney: The `SyncingService` is usually instantiated when the attachment queue is initialised.  The `SyncingService` then dynamically requests a lock on an `AttachmentContext` via the `AttachmentService`. This implementation should receive a reference to an `AttachmentService` here.

packages/attachments/src/StorageService.ts
```diff
@@ -0,0 +1,157 @@
...
+        };
+      }
+
+      return attachment;
+    }
+  }
+
+  async downloadAttachment(attachment: AttachmentRecord): Promise<AttachmentRecord> {
+    try {
+      const fileBlob = await this.remoteStorage.downloadFile(attachment);
```
stevensJourney: I think we should have the local storage adapter support writing the same type as returned from the remote `downloadFile` method without a conversion here.

packages/attachments/src/SyncErrorHandler.ts
```diff
@@ -0,0 +1,28 @@
+import { AttachmentRecord } from './Schema.js';
+
+/// If an operation fails and should not be retried, the attachment record is archived.
+export abstract class SyncErrorHandler {
```
stevensJourney: Rather make this an interface if all the methods are `abstract`.

packages/attachments/src/storageAdapters/IndexDBFileSystemAdapter.ts
```diff
@@ -0,0 +1,122 @@
...
+      req.onerror = () => reject(req.error);
+    });
+  }
+
+  async makeDir(path: string): Promise<void> {
+    // No-op for IndexedDB
+  }
+
+  async rmDir(path: string): Promise<void> {
+    // No-op for IndexedDB
```
stevensJourney: Should the delete not delete all the files in the directory?

packages/attachments/src/RemoteStorageAdapter.ts
```diff
@@ -0,0 +1,27 @@
...
+     */
+    uploadFile(fileData: ArrayBuffer, attachment: AttachmentRecord): Promise<void>;
+    /**
+     * Downloads a file from remote storage.
+     *
+     * @param attachment The `Attachment` describing the file to download.
+     * @returns The binary data of the downloaded file.
+     * @throws An error if the download fails or the file is not found.
+     */
+    downloadFile(attachment: AttachmentRecord): Promise<Blob>;
```
stevensJourney: I haven't tested this, but it seems like React Native might not support Blob directly. It seems like Expo has a [package](https://docs.expo.dev/versions/latest/sdk/blob/) for it, but typing might be mismatched. 
Maybe use ArrayBuffer here.

packages/attachments/src/storageAdapters/NodeFileSystemAdapter.ts
```diff
@@ -0,0 +1,79 @@
+import { promises as fs } from 'fs';
+import * as path from 'path';
+import { EncodingType, LocalStorageAdapter } from '../LocalStorageAdapter.js';
+
+export class NodeFileSystemAdapter implements LocalStorageAdapter {
+  async initialize(): Promise<void> {
+    // const dir = this.getUserStorageDirectory();
+    const dir = path.resolve('./user_data');
```
stevensJourney: Maybe allow the base directory to be specified as a constructor param and store it for use later.

packages/attachments/vitest.config.ts
```diff
@@ -1,8 +1,19 @@
 import topLevelAwait from 'vite-plugin-top-level-await';
+import wasm from 'vite-plugin-wasm';
 import { defineConfig, UserConfigExport } from 'vitest/config';
 
 const config: UserConfigExport = {
-  plugins: [topLevelAwait()],
+  worker: {
```
stevensJourney: An alternative could be to use the Node.js SDK for tests instead. That should require less config here.

packages/attachments/src/AttachmentQueue.ts
```diff
@@ -0,0 +1,250 @@
...
+  archivedCacheLimit: number;
+
+  constructor({
+    db,
+    localStorage,
+    remoteStorage,
+    watchAttachments,
+    logger,
+    tableName = ATTACHMENT_TABLE,
+    syncInterval = 30 * 1000,
```
stevensJourney: We typically indicate the millisecond unit in the variable name
```suggestion
    syncIntervalMs= 30 * 1000,
```

packages/attachments/src/AttachmentQueue.ts
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  async saveFile({
+    data,
+    fileExtension,
+    mediaType,
+    metaData,
+    id
+  }: {
+    data: ArrayBuffer | Blob | string;
```
stevensJourney: It would be nice to export this to it's own dedicated type instead of inlining.

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,250 @@
...
+    mediaType,
+    metaData,
+    id
+  }: {
+    data: ArrayBuffer | Blob | string;
+    fileExtension: string;
+    mediaType?: string;
+    metaData?: string;
+    id?: string;
+  }): Promise<AttachmentRecord> {
```
stevensJourney, line 345: The `updateHook` callback used in other SDKs is important and should be specified here.

That hooks allows the `saveFile` operation to also associate the newly created attachment record with relational data. Failing to associate the record could result in racy archiving of the newly created record.

stevensJourney (commented):
This looks like a good improvement so far. I've added some comments for some items I spotted.

packages/common/src/attachments/AttachmentContext.ts
joshuabrink: I added doc comments all of the public classes, functions and interfaces - should have total coverage now.

packages/attachments/src/AttachmentContext.ts
joshuabrink: I see, okay the separation makes sense - improved the consistency in the latest batch of commits. Also took the differential watched query approach.

packages/attachments/vitest.config.ts
khawarizmus: Although i have switched to Node. I think i might attempt to also write a test for the Browser testing our IndexDB adapter. Which makes me think. if we offer other adapters are we supposed to cover them in our tests? But i can see that our RN SDK doesn't have tests

packages/attachments/src/AttachmentQueue.ts
khawarizmus: Dropped Blob as it is Browser specific and kept ArrayBuffer and string for Base64 support.

demos/react-native-supabase-todolist/library/powersync/system.ts
```diff
@@ -50,18 +54,46 @@ export class System {
...
+      });
 
-          return { retry: true };
-        }
+      this.photoAttachmentQueue = new AttachmentQueue({
+        db: this.powersync,
+        localStorage,
+        remoteStorage,
+        watchAttachments: (onUpdate) => {
+          this.powersync.watch(
```
stevensJourney: It looks like we would need a way to dispose of the watched queries here. e.g. if the AttachmentsQueue was to be closed (or sync stopped), we'd need to stop these queries.

packages/attachments/package.json
```diff
@@ -9,6 +9,7 @@
     "type": "git",
     "url": "git+https://github.com/powersync-ja/powersync-js.git"
   },
+  "deprecated": "This package has been merged into @powersync/common. Please use @powersync/common instead.",
```
stevensJourney: Maybe we should recommend using the various SDK packages directly instead? We typically don't require users to install `@powersync/common` directly for most cases.

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,430 @@
...
+          // The attachment is present again. Need to queue it for sync.
+          // We might be able to optimize this in future
+          if (existingQueueItem.hasSynced === true) {
+            // No remote action required, we can restore the record (avoids deletion)
+            attachmentUpdates.push({
+              ...existingQueueItem,
+              state: AttachmentState.SYNCED
+            });
+          } else {
+            // The localURI should be set if the record was meant to be downloaded
```
stevensJourney:
```suggestion
            // The localURI should be set if the record was meant to be uploaded
```

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,430 @@
...
+
+  /**
+   * Verifies the integrity of all attachment records and repairs inconsistencies.
+   *
+   * This method checks each attachment record against the local filesystem and:
+   * - Updates localUri if the file exists at a different path
+   * - Archives attachments with missing local files that haven't been uploaded
+   * - Requeues synced attachments for download if their local files are missing
+   */
+  verifyAttachments = async (): Promise<void> => {
```
stevensJourney: For my curiosity, whats the reason for this being an arrow function?

packages/common/src/attachments/SyncingService.ts
```diff
@@ -0,0 +1,187 @@
...
+      const fileBlob = await this.localStorage.readFile(attachment.localUri);
+      await this.remoteStorage.uploadFile(fileBlob, attachment);
+
+      return {
+        ...attachment,
+        state: AttachmentState.SYNCED,
+        hasSynced: true
+      };
+    } catch (error) {
+      const shouldRetry = this.errorHandler?.onUploadError(attachment, error) ?? false;
```
stevensJourney: I think the default here should be `true`. We only want to skip retries if the `errorHandler` explicitly says to do so.

packages/common/tests/attachments.test.ts
```diff
@@ -0,0 +1,481 @@
+import { describe, expect, it, vi } from 'vitest';
+import { vol } from 'memfs';
+import { PowerSyncDatabase, Schema, Table, column, NodeFileSystemAdapter } from '../../node';
```
stevensJourney: It might be easier to just have these tests in the node package's unit tests.

packages/react-native/package.json
```diff
@@ -45,6 +45,7 @@
   "dependencies": {
     "@powersync/common": "workspace:*",
     "@powersync/react": "workspace:*",
+    "expo-file-system": "18.0.12",
```
stevensJourney: I assume this would not work on non-Expo React Native projects? One would need to confirm this dependency does not cause any issues on bare React native projects.

packages/attachments/package.json
khawarizmus: Yes that is a sensible thing to do here

packages/common/tests/attachments.test.ts
khawarizmus: I was thinking of adding the browser tests as well but didn't get the chance to do it. If we add that and perhaps find a way to support [react-native](https://github.com/sheremet-va/vitest-react-native) would you say the tests would need to be moved? if not i would say we keep them for now and slowly add tests to the other platforms.

PS: that's the only package i found for RN and it seems not actively developed (last commit one year ago)

stevensJourney: Having the tests be in the common package is a cool idea. It's been a todo for a while, to get some sort of shared tests between SDKs configured. Having the tests in common could be an option. I don't think we have any here, but one thing to be concerned about is circular dependencies. 

A React Native Vitest suite looks cool. I imagine this could be challenging to get stable.

The current approach we usually take is to test the generic common logic and then test the platform specific logic in their own SDKs' tests. Testing of the common logic can sometimes be tricky to do with mocks, in those cases we sometimes defer this generic testing to some SDK like `node` (e.g. the trigger functionality) or `web`. One could theoretically then just write encapsulated tests for the SDK specific storage adapters in the relevant SDK packages' tests.

packages/common/src/attachments/AttachmentQueue.ts
joshuabrink: I had an idea of passing this as a callback for one of the watchers and wanted to retain the correct `this` reference but ended up not using it for that use case.

demos/react-native-supabase-todolist/library/powersync/system.ts
joshuabrink: We were thinking of creating a AbortController in the AttachmentsQueue and passing it to `watchAttachments` (similar concept to the [swift implementation](https://github.com/powersync-ja/powersync.dart/blob/a34349ce860a85d62aa574f89d5cac502846b5b2/packages/powersync_core/lib/src/attachments/attachment_queue_service.dart#L212))

```ts
  watchAttachments(
    onUpdate: (attachment: WatchedAttachmentItem[]) => Promise<void>,
    signal: AbortSignal
  ): void {
    throw new Error('watchAttachments should be implemented by the user of AttachmentQueue');
  }
```
Then in `startSync`:

```ts
this.watchAttachmentsAbortController = new AbortController();
this.watchAttachments(async (watchedAttachments) => {
...
}, this.watchAttachmentsAbortController.signal)
```

Which we abort in the `stopSync`:
```ts
  async stopSync(): Promise<void> {
    ...
    if (this.watchAttachmentsAbortController) {
      this.watchAttachmentsAbortController.abort();
    }
  }

```

Then in user land they can pass the signal to their watchers:
```ts
watchAttachments: async (onUpdate, signal) => {
  const watcher = this.powersync.watch(
    `SELECT photo_id as id FROM ${TODO_TABLE} WHERE photo_id IS NOT NULL`,
    [],
    {
      signal
    }
 );
...
}
```

You see any issues with this approach?

packages/react-native/package.json
joshuabrink: It requires an expo project unfortunately. But, currently looking at `@dr.pogodin/react-native-fs` an active fork of react-native-fs, which works on native.

One question about including it as a direct dependency. I was thinking of going the dynamic import + optional peer dependency route (similar to [react-native-quick-sqlite](https://github.com/powersync-ja/powersync-js/blob/34a80d8e2c24ffbf4cb33852d9e7f72c25074a5a/packages/react-native/src/db/adapters/react-native-quick-sqlite/ReactNativeQuickSQLiteOpenFactory.ts#L21)), you see any potential pitfalls?

demos/react-native-supabase-todolist/library/powersync/system.ts
stevensJourney: > (similar concept to the [swift implementation](https://github.com/powersync-ja/powersync.dart/blob/a34349ce860a85d62aa574f89d5cac502846b5b2/packages/powersync_core/lib/src/attachments/attachment_queue_service.dart#L212))

You mean the Dart implementation? The Dart implementation returns a cancellable subscription, which is not really directly the same thing. But I think an abort signal as mentioned above should be fine.

Side note. It looks like the Dart implementation there is creating two subscriptions with `_watchAttachments` - that seems strange.

packages/react-native/package.json
stevensJourney: I've seen before, in the past, that nested direct dependencies sometimes aren't detected and installed correctly in the Expo build process. This might have changed at some point though. 
Generally, an optional peer dependency would follow our current patterns. Dynamic imports with Metro have been a pain in the past, however if one follows the same patters as React Native Quick SQLite, I think we should be fine.

demos/react-native-supabase-todolist/library/powersync/system.ts
joshuabrink: I did mean Dart apologies, and I saw that duplicate too - I think it slipped through the cracks, might be worth investigating.

packages/react-native/package.json
joshuabrink: I kept both, just for better coverage. I think catering for both Expo and vanilla React Native projects would be worthwhile. I understand there is a maintenance burden with supporting both adapters.

The vanilla implementation works on Expo as well, so let me know if you think this is redundant.

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,437 @@
...
+   * Callback function to watch for changes in attachment references in your data model.
+   *
+   * This method should be implemented to monitor changes in your application's
+   * data that reference attachments. When attachments are added, removed, or modified,
+   * this callback should trigger the onUpdate function with the current set of attachments.
+   *
+   * @param onUpdate - Callback to invoke when attachment references change
+   * @throws Error indicating this method must be implemented by the user
+   */
+  watchAttachments(onUpdate: (attachment: WatchedAttachmentItem[]) => Promise<void>, signal: AbortSignal): void {
```
stevensJourney: This looks a bit weird as a function definition with no actual implementation. This should probably be a class property instead (since it's assigned in the constructor)

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,437 @@
...
+  /** Service for synchronizing attachments between local and remote storage */
+  syncingService: SyncingService;
+
+  /** Adapter for local file storage operations */
+  localStorage: LocalStorageAdapter;
+
+  /** Adapter for remote file storage operations */
+  remoteStorage: RemoteStorageAdapter;
+
+  /** @deprecated Directory path for storing attachments  */
```
stevensJourney: Since this is a new version of the implementation which originates from a new source (the SDK packages themselves). I don't think we need to be deprecating things at this point - rather remove them.

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,437 @@
...
+/**
+ * AttachmentQueue manages the lifecycle and synchronization of attachments
+ * between local and remote storage.
+ *
+ * Provides automatic synchronization, upload/download queuing, attachment monitoring,
+ * verification and repair of local files, and cleanup of archived attachments.
+ */
+export class AttachmentQueue {
+  /** Timer for periodic synchronization operations */
+  periodicSyncTimer?: ReturnType<typeof setInterval>;
```
stevensJourney: The default here for these class members is `public` - which I think we should avoid for some of these items. For items which make sense to be public, maybe consider making them `readonly` or clearly handle/document their update behaviour.

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,437 @@
...
+  }) {
+    this.remoteStorage = remoteStorage;
+    this.localStorage = localStorage;
+    this.watchAttachments = watchAttachments;
+    this.tableName = tableName;
+    this.syncIntervalMs = syncIntervalMs;
+    this.syncThrottleDuration = syncThrottleDuration;
+    this.archivedCacheLimit = archivedCacheLimit;
+    this.downloadAttachments = downloadAttachments;
+    this.context = new AttachmentContext(db, tableName, logger ?? db.logger, archivedCacheLimit);
```
stevensJourney: One of the main purposes of the `AttachmentService` is to provide a locked/exclusive access to an `AttachmentContext`. Creating a context manually here defeats that purpose, please add a `withContext` method to `AttachmentService` which provides leased access to the context.

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,437 @@
...
+  remoteStorage: RemoteStorageAdapter;
+
+  /** @deprecated Directory path for storing attachments  */
+  attachmentsDirectory?: string;
+
+  /** Name of the database table storing attachment records */
+  tableName?: string;
+
+  /** Logger instance for diagnostic information */
+  logger?: ILogger;
```
stevensJourney: You can remove the optionality of this member since the constructor provides a default

packages/react-native/package.json
```diff
@@ -33,13 +33,21 @@
   "homepage": "https://docs.powersync.com/",
   "peerDependencies": {
     "@journeyapps/react-native-quick-sqlite": "^2.4.9",
+    "@dr.pogodin/react-native-fs": "^2.36.1",
```
stevensJourney: These should probably also be `devDependencies`. This probably currently works since the packages are being hoisted from the demos. This will soon no-longer be the case.

packages/react-native/rollup.config.mjs
```diff
@@ -58,6 +58,7 @@ export default () => {
       'js-logger',
       'react-native',
       'react',
+      '@dr.pogodin/react-native-fs',
```
stevensJourney: Should probably also include `expo-filesystem` here?

packages/react-native/src/attachments/ReactNativeFileSystemAdapter.ts
```diff
@@ -0,0 +1,101 @@
+import { decode as decodeBase64, encode as encodeBase64 } from 'base64-arraybuffer';
+import { AttachmentData, EncodingType, LocalStorageAdapter } from '@powersync/common';
+
+try {
```
stevensJourney: Is it possible to lazy load `RNFS` in the constructor of `ReactNativeFileSystemStorageAdapter ` instead of doing this globally?

packages/react-native/package.json
```diff
@@ -58,6 +67,7 @@
     "@rollup/plugin-replace": "^5.0.7",
     "@rollup/plugin-terser": "^0.4.4",
     "bson": "^6.10.4",
+    "expo-file-system": "^15.2.2",
```
stevensJourney: While reviewing https://github.com/powersync-ja/powersync-js/issues/803, I noticed that the latest version of expo-file-system is `19.0.21`.

The current latest version uses a `*` `peerDependency` constraint for [expo](https://github.com/expo/expo/blob/974fb4dc876db2face0b04f2f312ee4017c7895c/packages/expo-file-system/package.json#L55), which should theoretically be compatible with any version of Expo.

However, we're already 4 major versions behind `expo-file-system`, which raises an important question: what happens when the next major version (with breaking changes) is released? Bumping our `peerDependency` range for `expo-file-system` by a major version could be considered a breaking change for our users, requiring a major version bump of `@powersync/react-native`. I'd prefer to avoid making a major version bump solely for a filesystem dependency upgrade that only affects a small portion of this package's functionality (the attachment helpers).

We don't really have this issue for other JS platforms (or other SDKs) since the storage APIs (and general dependencies) we're using are quite standard. 

Moving the FileSystem adapter out of this package is also not ideal, but does seem like a potentially more manageable alternative.

joshuabrink: Good call, while not ideal, I'm not sure if there is a cleaner way to workaround the Expo dependency churn.

The lowest friction route I could think of, was to just create a straightforward`@powersync/attachments-react-native` package that handles both Expo and React Native file system adapters.

It's still a work-in-progress (docs and demos need updating) but the idea is there.

packages/attachments-react-native/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# @powersync/attachments-react-native
+
+## 0.1.0
+
+### Features
+
+- Initial release
```
stevensJourney: It's generally safer to add a `minor` changeset entry for this new package which contains this content. 

While it's not the case here (since we have a changeset for the PowerSync SDKs). Merging this PR to `main` without a changeset, for a new package, could result in that package getting published immediately - while having a changeset would create a release PR where one can smoke check versions before release.

packages/attachments-react-native/package.json
```diff
@@ -0,0 +1,82 @@
...
+  "scripts": {
+    "build": "tsc -b && rollup --config",
+    "build:prod": "tsc -b --sourceMap false && rollup --config",
+    "clean": "rm -rf lib dist tsconfig.tsbuildinfo",
+    "watch": "tsc -b -w",
+    "test:exports": "attw --pack ."
+  },
+  "peerDependencies": {
+    "@powersync/common": "workspace:^1.43.1",
+    "expo-file-system": ">=15.0.0 <20.0.0",
```
stevensJourney: Is this version constraint correct? My understanding is that version 19 included breaking changes (compared to previous versions). How does our implementation here handle both APIs?

packages/attachments-react-native/src/ExpoFileSystemStorageAdapter.ts
```diff
@@ -0,0 +1,152 @@
+import { decode as decodeBase64, encode as encodeBase64 } from 'base64-arraybuffer';
+import { AttachmentData, EncodingType, LocalStorageAdapter } from '@powersync/common';
+
+type ExpoFileSystemModule = {
```
stevensJourney: Can't we import the type only from the package?

```typescript
import type ExpoFS from 'expo-file-system'

// ...

export class ExpoFileSystemStorageAdapter implements LocalStorageAdapter {
  private fs: typeof ExpoFS;
```

One I made that change, and was using `expo-file-system-19` I quickly ran into other issues due to the different API.

I'd suggest having this lib only support version 19.

packages/attachments-react-native/src/ReactNativeFileSystemStorageAdapter.ts
```diff
@@ -0,0 +1,134 @@
+import { decode as decodeBase64, encode as encodeBase64 } from 'base64-arraybuffer';
+import { AttachmentData, EncodingType, LocalStorageAdapter } from '@powersync/common';
+
+type ReactNativeFsModule = {
```
stevensJourney: I'd also recommend just doing ` type` import here.

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,437 @@
...
+  /** Whether to automatically download remote attachments. Default: true */
+  readonly downloadAttachments: boolean = true;
+
+  /** Maximum number of archived attachments to keep before cleanup. Default: 100 */
+  readonly archivedCacheLimit: number;
+
+  /** Service for managing attachment-related database operations */
+  readonly attachmentService: AttachmentService;
+
+  watchActiveAttachments: DifferentialWatchedQuery<AttachmentRecord>;
```
stevensJourney: Should these, and `watchAttachments` not be private/protected?

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,437 @@
...
+   * data that reference attachments. When attachments are added, removed, or modified,
+   * this callback should trigger the onUpdate function with the current set of attachments.
+   */
+  readonly watchAttachments: (
+    onUpdate: (attachment: WatchedAttachmentItem[]) => Promise<void>,
+    signal: AbortSignal
+  ) => void;
+
+  /** Name of the database table storing attachment records */
+  readonly tableName?: string;
```
stevensJourney: should this be optional/nullable?

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,437 @@
...
+    this.syncingService = new SyncingService(this.attachmentService, localStorage, remoteStorage, this.logger, errorHandler);
+  }
+
+  /**
+   * Generates a new attachment ID using a SQLite UUID function.
+   *
+   * @returns Promise resolving to the new attachment ID
+   */
+  async generateAttachmentId(): Promise<string> {
+    return this.attachmentService.withContext(async (ctx) => {
```
stevensJourney: Nit: this doesn't need an attachment context (since it can't possibly interfere with attachment sync). this can just use the DB.

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,437 @@
...
+    });
+
+    // immediately invoke the sync storage to initialize local storage
+    await this.localStorage.initialize();
+
+    await this.verifyAttachments();
+
+    // Sync storage periodically
+    this.#periodicSyncTimer = setInterval(async () => {
+      await this.syncStorage();
```
stevensJourney, line 172: Also call this if the sync status changed to `connected: true` this allows for attachments to be processed as soon as the device becomes online (after being offline)

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -0,0 +1,437 @@
...
+  readonly syncIntervalMs: number = 30 * 1000;
+
+  /** Duration in milliseconds to throttle sync operations */
+  readonly syncThrottleDuration: number;
+
+  /** Whether to automatically download remote attachments. Default: true */
+  readonly downloadAttachments: boolean = true;
+
+  /** Maximum number of archived attachments to keep before cleanup. Default: 100 */
+  readonly archivedCacheLimit: number;
```
stevensJourney, line 64: For consistency with other SDKs, I think a lot of these members can be `protected` - or, I don't think they need to be public.

packages/common/src/attachments/SyncingService.ts
```diff
@@ -0,0 +1,190 @@
...
+/**
+ * Orchestrates attachment synchronization between local and remote storage.
+ * Handles uploads, downloads, deletions, and state transitions.
+ */
+export class SyncingService {
+  attachmentService: AttachmentService;
+  localStorage: LocalStorageAdapter;
+  remoteStorage: RemoteStorageAdapter;
+  logger: ILogger;
+  errorHandler?: AttachmentErrorHandler;
```
stevensJourney: Also, a lot of these members don't need to be public. I don't imagine there's much of a use case for code of the form `syncingService.localStorage. ...` etc.

packages/web/src/attachments/IndexDBFileSystemAdapter.ts
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        if (options?.encoding === EncodingType.UTF8) {
+          const encoder = new TextEncoder();
+          const arrayBuffer = encoder.encode(req.result).buffer;
+          resolve(arrayBuffer);
+        }
+
+        // Default base64 encoding
+        const base64String = req.result.replace(/^data:\w+;base64,/, '');
```
stevensJourney: It looks like the `saveFile` method stores the data as an ArrayBuffer, where the `req.result` should be an ArrayBuffer. When reading a file, the `replace` method is not found on an ArrayBuffer.
image

packages/attachments-react-native/src/ExpoFileSystemStorageAdapter.ts
joshuabrink: For importing the types directly, I was hesitant to include it as a type import - as I assumed it would force consumers to have expo-file-system types available (or use skipLibCheck) even if they are only importing the vanilla ReactNative adapter. But looking at it again, the types aren't being leaked in the way I thought, so will change it.

I'm also thinking of breaking it up into 2 adapters. Legacy and the current adapter. It couldn't hurt also including support for users using Expo 53 and below, maintenance burden also seems minimal. Let me know if think this isn't worth it.

stevensJourney: > But looking at it again, the types aren't being leaked in the way I thought, so will change it.

Correct! My understanding is that the `import` statement should not be present in `d.ts` files if the type is not actually exposed in anyway (which is the case here)

> I'm also thinking of breaking it up into 2 adapters. Legacy and the current adapter. It couldn't hurt also including support for users using Expo 53 and below, maintenance burden also seems minimal. Let me know if think this isn't worth it.

If it's not too tricky to implement reliably, then I think this is fine. I'm also happy with just supporting the latest version for now.

packages/attachments-react-native/package.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "@powersync/attachments-react-native",
```
stevensJourney: The last nit I can think of, is maybe we should have a naming convention like `@powersync/attachments-storage-react-native` - since this package just exposes storage adapters.
I just want to avoid situations where one might think `attachments-react-native` is everything one would need for attachments in React Native - which is not the case.

packages/attachments-react-native/package.json
```diff
@@ -0,0 +1,82 @@
...
+  "version": "0.1.0",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org/",
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/powersync-ja/powersync-js.git"
+  },
+  "author": "JOURNEYAPPS",
```
stevensJourney: confirmed with @kobiebotha , this should be `POWERSYNC`
```suggestion
  "author": "POWERSYNC",
```

stevensJourney (dismissed):
Happy with these changes from my side. 🚀

stevensJourney (dismissed):
🚀

khawarizmus:
I have updated the Docs PR to match the latest changes made. Mainly the introduction of the `@powersync/attachments-storage-react-native` package.

Feel free to review: https://github.com/powersync-ja/powersync-docs/pull/289

We should be good to go with both 🚀

packages/attachments-storage-react-native/README.md
```diff
@@ -0,0 +1,99 @@
+# @powersync/attachments-storage-react-native
+
+> [!NOTE]
+> Attachment helpers are currently in an **alpha** state, intended strictly for testing. Expect breaking changes and instability as development continues.
+>
+> Do not rely on this package for production use.
+
```
khawarizmus, line 7: It would be good to mention Expo version requirements here as well

stevensJourney:
Note: it looks like the attachments tests are failing here. Other Node.js tests should no longer be flaky. Please fix these new attachment tests before proceeding.

packages/web/src/attachments/IndexDBFileSystemAdapter.ts
```diff
@@ -0,0 +1,132 @@
...
+    });
+  }
+
+  async makeDir(path: string): Promise<void> {
+    // No-op for IndexedDB as it does not have a directory structure
+  }
+
+  async rmDir(path: string): Promise<void> {
+    const store = await this.getStore('readwrite');
+    // TODO: Test this to ensure it deletes all files under the directory
```
Chriztiaan: Fact checked?

joshuabrink: That one slipped through. Good catch, there is an issue where it doesn't delete subpaths.

packages/common/src/attachments/README.md
Chriztiaan, line 1: As discussed, I believe this is valuable and I would want this to be surfaced to our readmes.

packages/attachments-storage-react-native/package.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "@powersync/attachments-storage-react-native",
+  "version": "0.1.0",
```
Chriztiaan: Not sure if I missed something, but I normally start on 0.0.1 for new/alpha packages.

joshuabrink: The reasoning here was for consumers of the package, if they are using caret ^0.1.0 they can receive patch updates (0.1.x) automatically, while ^0.0.1 locks them to only patch of 0.0.1, making upgrades a little more cumbersome.

Issue #736, opened 2025-10-16
SQL Console doesn't consistently find tables in Sync Diagnostics Client / Diagnostics App


benitav:
With the following sync rules:

```
bucket_definitions:
  global:
    data:
      # Sync all rows
      - SELECT * FROM lists
      - SELECT * FROM todos
```

Running `select * from lists` sometimes finds the table as expected:

<img width="1581" height="664" alt="Image" src="https://github.com/user-attachments/assets/ca2fd49e-ff0a-4c1e-b807-d1eb229db536" />

And other times not:

<img width="1912" height="204" alt="Image" src="https://github.com/user-attachments/assets/cc3f030a-42e3-421c-adfe-5bd75cb112d9" />

Occasionally I've been able to get the query to work by running "Clear & Redownload" on the Sync Overview page, but this also doesn't work consistently.

I found that this happens both when using the JS and Rust-based sync client. 

In all cases the tables appear as expected on the Sync Overview page:

<img width="1542" height="314" alt="Image" src="https://github.com/user-attachments/assets/2411243d-b770-47af-9067-210ecaed4c79" />

stevensJourney:
Fixes for this have been released in `@powersync/diagnostics-app@0.9.14`

PR #737, opened 2025-10-20
fix: diagnostics app dynamic schema

stevensJourney:
# Overview

The sync diagnostics tool creates a dynamic schema from the bucket operations received from the PowerSync service. This dynamic schema is used to populate the PowerSync SQLite views which are typically used in various SQL queries - such as in the `SQL console` page.

The Rust implementation currently intercepts the PowerSync service commands and correctly attempts to process the operations into a dynamic schema. Unfortunately, processing the bucket operations currently only marks the dynamic schema as `dirty` - we still need to actually update the schema with the PowerSync client in order for SQL queries to use the views.

This typically results in errors such as https://github.com/powersync-ja/powersync-js/issues/736

This applies similar logic, as found in the [RecordingStorageAdapter](https://github.com/powersync-ja/powersync-js/blob/708b24b9950874c7f92d6c17419e5427eec40a6c/tools/diagnostics-app/src/library/powersync/RecordingStorageAdapter.ts#L48), for the `RustClientInterceptor`.

**Additional improvement:**

We currently seem to be manually implementing `refreshSchema` using a `DBAdapter`. This implementation does not trigger any `schemaChanged` listeners - which are used by watched queries. It is possible for an error condition as follows to occur:

- User connects
- User quickly navigates to the `SQL Console page` and uses a query for a PowerSync view, before the View was detected
- The watched query evaluates the dependent tables, due to the current schema an error `no such table ...` is displayed
- The table is eventually downloaded and the schema is applied (manually). The watched query is not informed of this change, it does not re-evaluate.
- The error persists

We now use the `AbstractPowerSyncDatabase` `refreshSchema` method to notify all listeners when the schema has changed. The watched queries should reevaluate the dependent tables and update.

### DIFF
--- a/demos/angular-supabase-todolist/extra-webpack.config.js
+++ b/demos/angular-supabase-todolist/extra-webpack.config.js
@@ -18,7 +18,7 @@ module.exports = (config, options, targetOptions) => {
       ...config.plugins,
       new webpack.DefinePlugin({
         // Embed environment variables starting with `WEBPACK_PUBLIC_`
-        'process.env': JSON.stringify(
+        env: JSON.stringify(
           Object.fromEntries(Object.entries(process.env).filter(([key]) => key.startsWith('WEBPACK_PUBLIC_')))
         )
       })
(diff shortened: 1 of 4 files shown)


simolus3 (dismissed):
The changes look good to me 👍

PR #746, opened 2025-10-28
Prepare for trusted publishing

simolus3:
This:

- Updates the Node version we use for tests to v22 (v20 will become unsupported soon). 
- Adopts corepack for managing the pnpm version.
- Uses dependency caching from `actions/setup-node` instead of the custom pnpm cache.
- Merges the dev and prod publishing workflows into a single workflow (npm will only let us do trusted publishing from a single workflow).
- Stops passing the NPM token, instead using trusted publishing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,5 +50,5 @@
     "typescript": "^5.9.3",
     "vitest": "^3.2.4"
   },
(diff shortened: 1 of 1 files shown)


simolus3:
Closing in favor of https://github.com/powersync-ja/powersync-js/pull/841

PR #749, opened 2025-10-29
[Node.js] Prevent database is locked errors

stevensJourney:
# Overview

I've noticed frequent `database is locked` errors when using the Node.js sdk for unit tests (using the `node:sqlite` implementation).

This seems to happen when the multiple read connections are concurrently opened. We set the journal mode to `WAL` on each of the read connections (concurrently).

We only need to set the journal mode once, with the writer connection. This updates the SDK to only set the journal mode for the writer connection.

I added a unit test which will fail on `main`, but passes on this branch.

### DIFF
--- a/packages/node/tests/PowerSyncDatabase.test.ts
+++ b/packages/node/tests/PowerSyncDatabase.test.ts
@@ -205,12 +205,9 @@ databaseTest('getCrudTransactions', async ({ database }) => {
   expect(remainingTransaction?.crud).toHaveLength(15);
 });
 
-tempDirectoryTest(
+// This is not a SemVer check, but is basic enough to skip this test on older versions of Node.js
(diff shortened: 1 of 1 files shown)


pnpm-workspace.yaml
```diff
@@ -3,3 +3,26 @@ packages:
   - packages/*
   - tools/*
   - docs/
+
+ignoredBuiltDependencies:
```
simolus3: Not sure if we should remove this or pick sensible values that we then commit.

stevensJourney: Yeah, this was automatically added by pnpm. I've only ran into issues with `better-sqlite` myself. So I've manually trimmed this list to automatically approve better-sqlite.
The PNPM [docs](https://pnpm.io/cli/approve-builds) mention it is fine to manually edit this.

PR #750, opened 2025-10-30
Vitest 4 upgrade

stevensJourney:
(PR description updated by @simolus3)

The audit check reports critical issues in vitest 3, so this updates to vitest version 4. The only significant change is that support for browsers has moved into separate packages (so we have to depend on and import `@vitest/browser-playwright`).

~~Note: Locally, I can't run `@powersync/web` tests anymore. It seems to work in CI though 🙃~~ Works after a `git clean -dfx`.

### DIFF
--- a/demos/example-vite/package.json
+++ b/demos/example-vite/package.json
@@ -17,7 +17,7 @@
   },
   "devDependencies": {
     "@swc/core": "~1.6.0",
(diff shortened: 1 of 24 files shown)


Chriztiaan (approved):
Tested package and demo tests locally. Looks good.

PR #753, opened 2025-11-03
Refactor build logic

simolus3:
This changes builds for powersync packages to make published libraries easier to debug:

1. Unconditionally emit source maps (todo: test).
3. Unconditionally disable minification. This seems to be what most other libraries are doing as well, users are likely to be running minifiers for prod builds anyway.
4. Bundle fewer dependencies: Dependencies which don't require polyfills don't really need to be bundled. By not bundling them, there's a potential for reducing application code size if other packages also depend on PowerSync-dependencies (because the libraries would be de-duplicated in that case).

Listing all packages:

- `adapter-sql-js`: Bundle our custom sql.js fork, no longer bundle `async-mutex`. Use source maps, non-minified builds.
- `attachments`: Enable source maps.
- `capacitor`: TODO
- `common`:
    - We had to inject a `TextDecoder` polyfill when we were still using `can-ndjson-stream`. We're not doing that anymore, I've added a `createTextDecoder()` method on `AbstractRemote` that the `react-native` package can override. This avoids us having to bundle that polyfill on all platforms.
    - I've adopted conditional exports and a separate Node build to stop bundling the `buffer` package for Node (we can and should just use `node:buffer` instead).
    - We should bundle `js-logger` (because it's a legacy UMD module that's hard to integrate) and rsocket (because it references `Buffer` as a global and we need to rewrite that to imports). However, I don't see why we would want to bundle `async-mutex`, `buffer` or `event-iterator`. These don't seem to create any issues, and this marks them as external again.
    - Side-node: There's more work to be done here. For instance, we can stop depending on `BSON` after removing the legacy sync client. I also think we should look into only supporting WebSockets on React Native to not pay the rsocket cost for all SDKs.
- `drizzle-driver`: Enable source maps.
- `kysely-driver`: Enable source maps.
- `node`: Replace `async-locks` with `async-mutex` (that package is already used in `common`).
- `powersync-op-sqlite`: No changes necessary.
- `react`: Enable source maps.
- `react-native`: Enable source maps, disable minification.
- `tanstack-react-query`: Enable source maps.
- `vue`:  Enable source maps.
- `web`:  Update dependencies.

TODOS:

- [x] Test packages.
- [x] Test vite web demo.
- [x] Test webpack web demo.
- [x] Test react native demo.
- [x] Test react native web demo.
- [x] Test node demo.
- [x] Test node electron demo.
- [ ] Test pre-release / published packages to ensure source maps don't cause vite warnings.

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -56,7 +56,6 @@
   },
   "dependencies": {
     "async-mutex": "^0.5.0",
-    "bson": "^6.10.4",
     "buffer": "^6.0.3",
     "event-iterator": "^2.0.0"
   },
@@ -71,6 +70,7 @@
     "cross-fetch": "^4.1.0",
     "js-logger": "^1.6.1",
     "rsocket-core": "1.0.0-alpha.3",
-    "rsocket-websocket-client": "1.0.0-alpha.3"
+    "rsocket-websocket-client": "1.0.0-alpha.3",
+    "bson": "^6.10.4"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1970,9 +1970,6 @@ importers:
       async-mutex:
         specifier: ^0.5.0
         version: 0.5.0
-      bson:
-        specifier: ^6.10.4
-        version: 6.10.4
       buffer:
         specifier: ^6.0.3
         version: 6.0.3
@@ -1998,6 +1995,9 @@ importers:
       '@types/uuid':
         specifier: ^9.0.1
         version: 9.0.8
+      bson:
+        specifier: ^6.10.4
+        version: 6.10.4
       cross-fetch:
         specifier: ^4.1.0
         version: 4.1.0(encoding@0.1.13)


packages/common/package.json
```diff
@@ -38,31 +49,28 @@
...
-    "build:prod": "tsc -b --sourceMap false && rollup -c rollup.config.mjs --sourceMap false",
+    "build:prod": "tsc -b && rollup -c rollup.config.mjs",
     "clean": "rm -rf lib dist tsconfig.tsbuildinfo",
     "test": "vitest",
     "test:exports": "attw --pack ."
   },
   "dependencies": {
-    "js-logger": "^1.6.1"
+    "async-mutex": "^0.5.0",
+    "bson": "^6.10.4",
```
stevensJourney: I think `bson` here could be a `devDependency`. We only `import type` in `AbstractRemote.ts`

packages/common/rollup.config.mjs
```diff
@@ -1,52 +1,49 @@
...
+      {
+        file: `dist/bundle${suffix}.cjs`,
+        format: 'cjs',
+        sourcemap: true
+      }
+    ],
+    plugins: [
+      [
         json(),
         nodeResolve({ preferBuiltins: false, browser: true }),
```
stevensJourney, line 29: Just wondering. Should the `browser` value be `!isNode`? I know it wasn't like this before, so this is still backwards compatible.

packages/react-native/src/sync/stream/ReactNativeRemote.ts
```diff
@@ -55,6 +56,10 @@ export class ReactNativeRemote extends AbstractRemote {
     return BSON;
   }
 
+  protected createTextDecoder(): TextDecoder {
+    return new TextDecoder();
```
stevensJourney, line 60: Expo seems to support a global [TextDecoder](https://docs.expo.dev/versions/v52.0.0/sdk/encoding/) now. It might be interesting to see if that could be used for our use case.

packages/common/rollup.config.mjs
```diff
@@ -1,52 +1,49 @@
...
-        },
-        {
-          file: 'dist/bundle.cjs',
-          format: 'cjs',
-          sourcemap: sourceMap
-        }
-      ],
-      plugins: [
+  return {
+    input: 'lib/index.js',
```
stevensJourney, line 13: This the sourcemaps here are a nice improvement. 
E.g. if I cause a manual error in `AbstractRemote.getCredentials()` I now get a nice stack trace.
```
logger.js:197 [PowerSyncDatabase[example.db]] Error: Not implemented
    at WebRemote.getCredentials (AbstractRemote.js:83:15)
    at WebRemote.buildRequest (AbstractRemote.js:124:40)
    at WebRemote.socketStreamRaw (AbstractRemote.js:209:36)
    at WebStreamingSyncImplementation.legacyStreamingSyncIteration (AbstractStreamingSyncImplementation.js:501:48)
    at async callback (AbstractStreamingSyncImplementation.js:452:21)
```
The stack trace could be even better if we has TypeScript source mappings. 
```
logger.js:197 [PowerSyncDatabase[example.db]] Error: Not implemented
    at WebRemote.getCredentials (AbstractRemote.ts:147:11)
    at WebRemote.buildRequest (AbstractRemote.ts:197:36)
    at WebRemote.socketStreamRaw (AbstractRemote.ts:305:32)
    at WebStreamingSyncImplementation.legacyStreamingSyncIteration (AbstractStreamingSyncImplementation.ts:718:42)
    at async callback (AbstractStreamingSyncImplementation.ts:665:11)
```
This could be achieved by setting the `input` to `src/index.ts` and using the `typescript` rollup plugin

```
      typescript({
        tsconfig: './tsconfig.json',
        outDir: 'dist',
        sourceMap: true
      }),
```

I might be missing if this causes other issues though. It would be nice if we enabled TypeScript source maps for all packages.

packages/common/rollup.config.mjs
simolus3: Our dependencies happened to not have conditional browser exports, but I still think `browser: !isNode` would be a better option here to guard against this. Nice catch!

packages/react-native/src/sync/stream/ReactNativeRemote.ts
simolus3: Very interesting! The package we use [picks up existing instances](https://github.com/inexorabletash/text-encoding/blob/3f330964c0e97e1ed344c2a3e963f4598610a7ad/lib/encoding.js#L3298-L3301) if they're defined, so using the wrapper package in the `react-native` package should still be fine (well, except for it increasing code size).

stevensJourney: Nice. That conditional export is nice.

packages/common/rollup.config.mjs
simolus3: I am getting a lot of build errors in other packages with this. I would prefer to add the plugin later updating tsc and rollup across the workspace to make this easier.

stevensJourney (approved):

packages/common/rollup.config.mjs
```diff
@@ -1,52 +1,49 @@
...
-      // This makes life easier
-      external: [
-        // This has dynamic logic - makes bundling hard
-        'cross-fetch'
+          Buffer: isNode ? ['node:buffer', 'Buffer'] : ['buffer/', 'Buffer']
+        })
       ]
-    },
+    ],
+    external: ['async-mutex', 'bson', 'buffer/', 'event-iterator']
```
Chriztiaan, line 36: FYI @simolus3, we might want to consider removing event-iterator from the external list - otherwise end users may need to have an include rule for it when using vite: https://github.com/aguynamedloren/powersync-js/blob/main/demos/example-vite/vite.config.ts#L23 

We have had to do this for the new dashboard project in the interim, do you perhaps see an alternative path forward?

stevensJourney: Due to the effect this has on Vite users, I think we should revert the externalization of `event-iterator`. I'll make a PR for this.

PR #754, opened 2025-11-03
[Demo] TanStack DB

stevensJourney:
# Overview

This adds a React web app demo using the TanStackDB integration for local queries and mutations.

This demo has all the same functions (and looks exactly the same) as the React Supabase Todolist demo. The main differences are in the implementation.

- The types used throughout the demo are derived from Zod Schemas with advanced validations and type transforms
- We use the `useLiveQuery` hook from TanStackDB for reactivity
- Full Text Search is replaced by a regular TanStackDB query.

This demo uses `@tanstack/db`, `@tanstack/powersync-db-collection` and `@tanstack/react-db`. 

The demo here is mainly focused on the PowerSync collection integration. We don't use `@tanstack/react-router` here (yet).

### DIFF
--- a/.changeset/fifty-cats-occur.md
+++ b/.changeset/fifty-cats-occur.md
@@ -1,0 +1,5 @@
+---
+'@powersync/capacitor': patch
+---
+
+Updated limitations in README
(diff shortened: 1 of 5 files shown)


Chriztiaan:
Tested on my side, LG2M.

Chriztiaan (approved):

PR #761, opened 2025-11-06
Fix attachments build

rkistner:
Fix build regression from #753.

This caused prod builds (package releases) to hang on the build step. Weirdly, it doesn't happen when running `pnpm build:prod` directly in the attachments package, but does happen when running the global `pnpm build:packages:prod`.


simolus3 (approved):
Ouch, sorry for that and thanks for adding the `:prod` step to catch this in the future 👍

Issue #763, opened 2025-11-07
Drizzle queries appear to use writeLock for SELECT statements - need help understanding if this is expected


wturon:
Hey folks! 

I'm experiencing a pretty bad performance issues in my React Native app where all database queries seem to block each other (5-10 second freezes). I've traced the execution paths and it appears that there's a writeLock being set for all queries.

**What I'm Observing**

When I trace the execution paths of different query types, I'm seeing that all Drizzle queries route through `executeRaw() -> writeLock()`, while raw PowerSync queries use `readLock()`:

Drizzle queries (all use writeLock):
- `db.query.findMany() → executeRaw() → writeLock`
- `db.query.findFirst() → executeRaw() → writeLock`
- `db.select().from() → executeRaw() → writeLock`
- `db.select({ cols }).from() → executeRaw() → writeLock`

Raw PowerSync queries (use readLock):
- `powerSyncDb.getAll() → getAll() → readLock`
- `powerSyncDb.get() → get() → readLock`

Environment

- @powersync/drizzle-driver: ^0.6.0
- @powersync/op-sqlite: ^0.7.11
- @powersync/react-native: ^1.25.1
- @powersync/common: ^1.40.0
- drizzle-orm: ^0.39.3
- Platform: React Native (Expo)
- SQLite: WAL mode enabled

**Code Path I'm Seeing**

Looking at the source, it appears the execution goes:

```javascript
  PowerSyncSQLitePreparedQuery:
  async values(placeholderValues) {
      const params = fillPlaceholders(this.query.params, placeholderValues ?? {});
      this.logger.logQuery(this.query.sql, params);
      return await this.db.executeRaw(this.query.sql, params);
  }
```

  This values() method is called by both .all() and .get() when field mappings exist (which seems to always be the case with Drizzle's query builder).

```javascript
  Then in OPSqliteAdapter:
  executeRaw(query, params) {
      return this.writeLock(ctx => ctx.executeRaw(query, params));
  }
```

My Questions

1. Is this the expected behavior for executeRaw()? Should it use writeLock() even for SELECT queries?
2. Is there a different code path I should be using for read queries with Drizzle?
3. Should .values() be calling a different method for SELECT queries?

I have test code that traces the exact execution paths if that would be helpful!

Thanks so much for any guidance! Really appreciate all the work you've done on PowerSync - it's been great otherwise. 🙏

PR #765, opened 2025-11-11
fix: Drizzle Read Connection Concurrency

stevensJourney:
# Overview

should close https://github.com/powersync-ja/powersync-js/issues/763

The current Drizzle driver uses the `AbstractPowerSyncDatabase` `writeLock` method for most query operations. Read locks are only used if an explicit `read only` transaction is started.

Using a write lock for Select queries hampers concurrency in applications using environments which support multiple read connections - such as React Native and Node.js. Thanks @rkistner for adding this!

This PR updates the query execution logic to use `queryMetaData` from https://github.com/drizzle-team/drizzle-orm/pull/4447 in order to detect `SELECT` queries. This approach has some limitations:
- `queryMetaData` is not available for  `db.query[collectionName]` queries. We use a slightly nasty override for `db.query[collectionName]` queries. The current version only has `findMany` and `firstFirst` options - which should be read queries

### DIFF
--- a/demos/react-supabase-todolist-tanstackdb/package.json
+++ b/demos/react-supabase-todolist-tanstackdb/package.json
@@ -17,6 +17,7 @@
     "@mui/x-data-grid": "^6.19.6",
     "@powersync/react": "workspace:*",
     "@powersync/web": "workspace:*",
+    "@powersync/common": "workspace:*",
     "@supabase/supabase-js": "^2.39.7",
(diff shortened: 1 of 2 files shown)


stevensJourney:
The `Test Packages` action is still super flaky. I've seen the Node.JS sync tests fail almost 15 times :(. Skipping this for now in order to get this out.

PR #767, opened 2025-11-11
Update to pnpm version 10

simolus3:
This adopts pnpm version 10 and switches to corepack to install it in GH actions instead of using an explicit setup step. That then allows using the package cache built in to `actions/setup-node`.

### DIFF
--- a/demos/example-vite/vite.config.ts
+++ b/demos/example-vite/vite.config.ts
@@ -16,10 +16,8 @@ export default defineConfig({
     // https://github.com/vitejs/vite/issues/11672#issuecomment-1415820673
     exclude: ['@journeyapps/wa-sqlite', '@powersync/web']
   },
-  plugins: [],
   worker: {
-    format: 'es',
-    plugins: () => []
+    format: 'es'
   },
   test: {
     globals: true,
--- a/demos/example-vite/package.json
+++ b/demos/example-vite/package.json
@@ -12,7 +12,8 @@
     "test:build": "pnpm build && vitest run"
   },
   "dependencies": {
-    "@powersync/web": "^1.31.0"
+    "@powersync/web": "^1.31.0",
+    "@journeyapps/wa-sqlite": "^1.4.1"
   },
   "devDependencies": {
     "@swc/core": "~1.6.0",
--- a/scripts/demos-inject-local.ts
+++ b/scripts/demos-inject-local.ts
@@ -32,19 +32,12 @@ const filterDemosPattern = (demos: string[], options: OptionValues): string[] =>
 };
 
 const ensureLocalLinkingBuilt = async () => {
-  try {
-    await fs.access(PNPMFILE_DIST_PATH);
-    return;
-  } catch {}
-
   console.log('Building @powersync/local-linking ...');
   // Run build from repo root
   execSync('pnpm -C tools/local-linking build', {
     cwd: path.resolve(import.meta.dirname, '..', '..'),
     stdio: 'inherit'
   });
-
-  await fs.access(PNPMFILE_DIST_PATH);
 };
 
 const injectPackagesAll = async (demos: string[], options: OptionValues) => {
--- a/tools/local-linking/src/pnpmfile.ts
+++ b/tools/local-linking/src/pnpmfile.ts
@@ -91,7 +91,7 @@ export = {
 
       scanDeps(pkg.dependencies);
       scanDeps(pkg.devDependencies);
-
+      scanDeps(pkg.peerDependencies);
       return pkg;
     }
   }


rkistner:
This is the first time I see the version catalogs feature - looks really useful for monorepos like this.

packages/drizzle-driver/rollup.config.js
```diff
@@ -30,7 +30,9 @@ export default (commandLineArgs) => {
            * The Typescript plugin complains about internal Drizzle types not matching when selecting
            * other moduleResolution settings.
            */
-          moduleResolution: 'bundler'
+          moduleResolution: 'bundler',
+          // ES2017+ has native async/await, so no __awaiter helper needed (avoids tslib dependency)
+          target: 'ES2017'
```
stevensJourney: The import tests were failing due to the `dist/index.cjs` containing

```javascript
require('tslib')
```

The `drizzle-driver` package doesn't depend on `tslib`, which with our updated hoisting config - now exposes this error. Bumping the target removes the requirement for `tslib`.

packages/web/webpack.config.js
```diff
@@ -39,9 +39,7 @@ export default () => {
         stream: require.resolve('stream-browserify'),
         vm: require.resolve('vm-browserify')
       },
-      alias: {
-        bson: path.resolve(__dirname, '../../node_modules/bson/lib/bson.cjs')
-      }
+      alias: {}
```
stevensJourney, line 42: It looks like we no longer need this `alias`. I believe this has been fixed after we bumped the `bson` package to [6.10](https://github.com/mongodb/js-bson/releases/tag/v6.10.4). That release removed the requirement for `top-level-await`. Which is now reflected in more of our Vite configs (We should make another PR to remove this from all other demos and configs).

stevensJourney:
I've updated this PR with a few changes in order to get it ready for release.

## PNPM Hoisting changes

This PR removed the `node-linker=hoisted` setting in the `.npmrc` file (a welcomed changed). This means packages are no longer hoisted to the main `node_modules` folder. This PR contains a few minor changes to explicitly introduce phantom dependencies (as regular dependencies) to packages which used them.

## PNPM Catalogs

I've tried to share as many common NPM dependency versions with `catalog:` version specifiers. Some packages are difficult to unify to a single version without breaking internal peer dependency ranges. One big example of this is `react` which in some cases we are currently building against React 18 (like in the Hook packages) while some React Native packages require React 19. 

The React tests currently fail when upgrading the dev dependency to React 19. We should properly investigate that soon.

## Linking Local Packages

Our current tooling to link local SDK packages to the isolated demos in the `demos/` folder uses a [pnpmfile](https://pnpm.io/pnpmfile) to override the PNPM package resolving of workspace packages to use `file: ` dependencies.

This version of linking breaks when using `catalog: ` dependency version specifiers in our local packages. This is due to the Demo's PNPM workspace not having any definition for `catalog: ` version specifiers.

We need some way to replace `catalog: ` version ranges for linked packages.

I've updated the current pnpmfile in order to cater for this:
- The `pnpmfile` is now no longer a JS string literal created in the `demos-inject-local.ts` file. The file is now a dedicated TypeScript file which is compiled to CommonJS (as required by PNPM). The file is in it's own dedicated package since the other scripts are not intended for CommonJS targets.
- The script now uses `@pnpm/workspace.find-packages` and `@pnpm/workspace.read-manifest` to resolve locally linked packages when `pnpm install` is executed inside demo folders. These dependencies are resolved from the parent monorepo - this seems to work fine for now. If there are issues, we can consider bundling those deps.

## React Native Tests

The PNPM hoisting changes usually cause issues with React Native and Metro. Our automated `powersynctests` were broken after this change. I've updated the Android build to resolve Node.js dependencies to the new locations. The project was also updated to React Native `0.83` as this version of Metro works much better with PNPM. 

## BSON Changes

We recently updated our internal BSON package dependency to [^6.10.4](https://github.com/mongodb/js-bson/releases/tag/v6.10.4). This release removed the use of `top-level-await`, but also seems to have fixed the requirement for other bundling hacks we've been doing for `@powersync/web` and `@powersync/react-native`. I've removed the stale bundling configs. This has been tested by running the `react-supabase-todolist` and `react-native-web-supabase-todolist` demos.

It also seems like our current Vite config examples are out-of-date. It seems like we also no longer need the `vite-plugin-wasm` plugin. I haven't updated all the configs for this yet. This PR is already large enough. I'll make an issue for this to track the status.

packages/drizzle-driver/rollup.config.js
stevensJourney: The `import-tests` only seem to fail in CI for some reason. I've opted to add `tslib` as a dependency instead of bumping the minimum supported target.

From the package [docs](https://www.npmjs.com/package/tslib)
> Because this can avoid duplicate declarations of things like __extends, __assign, etc., this means delivering users smaller files on average, as well as less runtime overhead

demos/example-vite/vite.config.ts
```diff
@@ -18,10 +16,10 @@ export default defineConfig({
     // https://github.com/vitejs/vite/issues/11672#issuecomment-1415820673
     exclude: ['@journeyapps/wa-sqlite', '@powersync/web']
   },
-  plugins: [wasm(), topLevelAwait()],
+  plugins: [],
```
simolus3: Isn't this the default? I suppose we could also remove these lines entirely as well then (also applies to some of the other packages, but we can also do that cleanup in a follow-up PR).

Chriztiaan (dismissed):
Happy with the changeset from my side, I am keen for the Vite config sweep!

Chriztiaan (approved):

PR #775, opened 2025-11-19
fix: Potential Multitab Transaction Issues With IndexedDB

stevensJourney:
## Problem

A db transaction death-loop could occur when using IndexedDB across multiple tabs, manifesting as the error:
```
cannot start a transaction within a transaction
```

This happens when a tab holding an active SQLite transaction is closed before committing or rolling back that transaction. Since IndexedDB shares a single SQLite connection (via a shared web worker) across all tabs, the uncommitted transaction persists and blocks subsequent transaction attempts.

## Root Cause

- Each tab shares its SQLite connection with the shared sync worker
- The shared DBAdapter can be closed during an active transaction (e.g., when a tab closes)
- Uncommitted transactions remain active on the shared connection
- Subsequent `BEGIN` statements fail because a transaction is already in progress

## Solution

This PR implements a connection "hold" mechanism:

1. **Hold acquisition**: A hold is created when a worker starts using a connection
2. **Hold release**: The hold is released when work completes normally
3. **Orphaned hold detection**: If the shared worker detects a hold changed without being properly released (indicating the tab closed mid-transaction), it executes `ROLLBACK` to clean up the orphaned transaction

This ensures transactions are always resolved, even when tabs close unexpectedly.

### DIFF
--- a/packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
+++ b/packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
@@ -14,6 +14,8 @@ import { getNavigatorLocks } from '../..//shared/navigator';
 import { AsyncDatabaseConnection } from './AsyncDatabaseConnection';
 import { SharedConnectionWorker, WebDBAdapter } from './WebDBAdapter';
 import { WorkerWrappedAsyncDatabaseConnection } from './WorkerWrappedAsyncDatabaseConnection';
+import { WASQLiteVFS } from './wa-sqlite/WASQLiteConnection';
+import { ResolvedWASQLiteOpenFactoryOptions } from './wa-sqlite/WASQLiteOpenFactory';
 import { ResolvedWebSQLOpenOptions } from './web-sql-flags';
 
 /**
@@ -48,6 +50,7 @@ export class LockedAsyncDatabaseAdapter
   protected _disposeTableChangeListener: (() => void) | null = null;
   private _config: ResolvedWebSQLOpenOptions | null = null;
   protected pendingAbortControllers: Set<AbortController>;
+  protected requiresHolds: boolean | null;
 
   closing: boolean;
   closed: boolean;
@@ -59,6 +62,7 @@ export class LockedAsyncDatabaseAdapter
     this.pendingAbortControllers = new Set<AbortController>();
     this.closed = false;
     this.closing = false;
+    this.requiresHolds = null;
     // Set the name if provided. We can query for the name if not available yet
     this.debugMode = options.debugMode ?? false;
     if (this.debugMode) {
@@ -107,6 +111,10 @@ export class LockedAsyncDatabaseAdapter
     this._config = await this._db.getConfig();
     await this.registerOnChangeListener(this._db);
     this.iterateListeners((cb) => cb.initialized?.());
+    /**
+     * This is only required for the long-lived shared IndexedDB connections.
+     */
+    this.requiresHolds = (this._config as ResolvedWASQLiteOpenFactoryOptions).vfs == WASQLiteVFS.IDBBatchAtomicVFS;
   }
 
   getConfiguration(): ResolvedWebSQLOpenOptions {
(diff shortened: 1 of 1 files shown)


.changeset/silver-insects-unite.md
```diff
@@ -0,0 +1,6 @@
+---
+'@powersync/web': patch
+---
+
+- Fixed an issue where IndexedDB could cause "cannot start a transaction within a transaction" errors.
+- Improved reconnect logic when multiple tabs are closed.
```
stevensJourney, line 6: This is a tangential, but related fix. We'll now await the closing of shared connections in the shared sync webworker.

packages/web/src/db/adapters/wa-sqlite/WASQLiteConnection.ts
```diff
@@ -172,6 +193,30 @@ export class WASqliteConnection
...
+    const previousHoldId = this._holdId;
+    this._holdId = `${++this._holdCounter}`;
+    if (previousHoldId) {
+      await this.iterateAsyncListeners(async (cb) => cb.holdOverwritten?.(previousHoldId));
+    }
+    return this._holdId;
+  }
+
+  async releaseHold(holdId: string): Promise<void> {
+    if (holdId != this._holdId) {
```
stevensJourney, line 214: This should never happen in practice, hence the exception being thrown.

packages/web/src/worker/sync/SharedSyncImplementation.ts
```diff
@@ -334,12 +334,15 @@ export class SharedSyncImplementation extends BaseObserver<SharedSyncImplementat
...
     for (const closeListener of trackedPort.closeListeners) {
-      closeListener();
+      await closeListener();
     }
 
     if (this.dbAdapter && this.dbAdapter == trackedPort.db) {
       // Unconditionally close the connection because the database it's writing to has just been closed.
-      await this.connectionManager.disconnect();
+      // The connection has been closed previously, this might throw. We should be able to ignore it.
+      await this.connectionManager
```
stevensJourney, line 343: For OPFS, the disconnect performs a DB operation. This might throw if the tab has closed. We can log and error here, we'll need to continue in order to restore connection if required.

packages/web/src/db/adapters/AsyncDatabaseConnection.ts
```diff
@@ -25,6 +25,21 @@ export type OnTableChangeCallback = (event: BatchedUpdateNotification) => void;
 export interface AsyncDatabaseConnection<Config extends ResolvedWebSQLOpenOptions = ResolvedWebSQLOpenOptions> {
   init(): Promise<void>;
   close(): Promise<void>;
+  /**
+   * Marks the connection as in-use by a certain actor.
+   * @returns A hold ID which can be used to release the hold.
+   */
+  markHold(): Promise<string>;
```
rkistner, line 32: I wonder if we could reduce overhead introduced by the holds if we do the following?
1. Let the client generate the `holdId`, and pass it in as a parameter here.
2. Make `markHold` and `releaseHold` synchronous, performing the events in the background.

Potential risks:
1. Uncaught errors - could just catch and log.
2. The hold release only actually happening after the lock is released, potentially leading to another client marking a hold before it's released. Not sure if this could happen in practice - maybe keep releaseHold asynchronous in that case?
3. I assume releasing before marking shouldn't be possible either way, since the underlying MessagePort should be ordered.

stevensJourney: > Make markHold and releaseHold synchronous, performing the events in the background.

Unfortunately we cannot run the `markHold` operation in the background either. The check for if the hold has been overwritten needs to happen before operations from the LockContext run - since it has to cleanup any potentially orphaned transactions.

stevensJourney: I can't really think of a safe way to optimize the current logic. I did make one optimisation now, where we only need to request these holds for IndexedDB connections. For OPFS don't need them since there are many dedicated connections which are shared - once their tab is closed the connection dies.
Given that OPFS usually is the faster alternative, I hope this is satisfactory.

Issue #778, opened 2025-11-21
SQLite Sequence Error on Update/Delete Mutations: 'SQLite Sequence should not be empty'


mikecfisher:
### Environment
  - **Package**: \`@powersync/web\` v1.29.0
  - **Database**: SQLite WASM (\`WASQLiteOpenFactory\`)
  - **Framework**: Next.js 16, React 19
  - **Integration**: TanStack DB (\`@tanstack/react-db\`)

  ### Problem

  UPDATE/DELETE mutations randomly fail with:
  ```
  Error: SQLite Sequence should not be empty
  ```

  Error occurs in PowerSync worker at line ~16931:
  ```
  const rs = await tx.execute("SELECT seq FROM sqlite_sequence WHERE name = 'ps_crud'");
  if (!rs.rows?.length) {
      throw new Error('SQLite Sequence should not be empty');
  }
  ```

  ### Root Cause

  PowerSync's internal `ps_crud` table uses `INTEGER PRIMARY KEY AUTOINCREMENT`, which requires a `sqlite_sequence` entry. The worker **checks for
   the sequence entry BEFORE inserting**,  but the entry doesn't exist on a fresh database until the first INSERT.

  **Race condition:**
  1. Fresh page load → PowerSync initializes → `ps_crud` is empty
  2. User triggers update/delete mutation immediately
  3. PowerSync checks `sqlite_sequence` for `ps_crud` entry
  4. Entry doesn't exist (never inserted yet) → **Error thrown**
  5. INSERT never happens → sequence never auto-created

  ### Why It's Intermittent

  **Scenario 1 (ERROR):**
  - Fresh load → immediate update/delete
  - No prior writes to `ps_crud` → no sequence entry
  - ❌ Error

  **Scenario 2 (WORKS):**
  - Fresh load → background sync runs first
  - Sync writes to `ps_crud` → SQLite auto-creates sequence entry
  - User triggers mutation
  - ✅ Works (sequence exists)

INSERT operations work fine because they're usually slower (user fills form). UPDATE/DELETE happen immediately (single button click),
   so they race the background sync.

  ### Reproduction

  ```typescript
  import { PowerSyncDatabase } from '@powersync/web';
  import { powerSyncCollectionOptions } from '@tanstack/react-db';

  // 1. Initialize PowerSync (fresh IndexedDB)
  const db = new PowerSyncDatabase({ schema, database: factory });
  await db.init();

  // 2. Create TanStack collection
[...]

PR #781, opened 2025-11-24
fix: Syncing After Uploads when using Triggers

stevensJourney:
# Overview

When using triggers (e.g., `database.triggers.createDiffTrigger()`), the `updateLocalTarget()` method could fail to read the `ps_crud` sequence value, causing sync to be blocked after performing uploads.

## Root Cause

The trigger system creates temporary tables and triggers (`CREATE TEMP TABLE` and `CREATE TEMP TRIGGER`) in the `temp` schema on the connection. When these temporary objects exist, SQLite's schema resolution mechanism prioritizes the `temp` schema when resolving unqualified table names. This means that queries to `sqlite_sequence` without an explicit schema qualifier could resolve to `temp.sqlite_sequence` instead of `main.sqlite_sequence`, returning no rows even though the sequence was actually advancing correctly.

### Solution

Explicitly qualify `sqlite_sequence` queries with the `main.` schema prefix in `SqliteBucketStorage.updateLocalTarget()`:
- Changed `SELECT seq FROM sqlite_sequence` → `SELECT seq FROM main.sqlite_sequence`

This ensures we always query the main database's sequence table, regardless of temporary schema context.

closes https://github.com/powersync-ja/powersync-js/issues/778

### DIFF
--- a/.changeset/four-falcons-arrive.md
+++ b/.changeset/four-falcons-arrive.md
@@ -1,0 +1,5 @@
+---
+'@powersync/common': patch
+---
+
+Fixed issue where using triggers could block syncing after performing uploads.
--- a/packages/common/src/client/sync/bucket/SqliteBucketStorage.ts
+++ b/packages/common/src/client/sync/bucket/SqliteBucketStorage.ts
@@ -255,7 +255,7 @@ export class SqliteBucketStorage extends BaseObserver<BucketStorageListener> imp
       // Nothing to update
       return false;
     }
-    const rs = await this.db.getAll<{ seq: number }>("SELECT seq FROM sqlite_sequence WHERE name = 'ps_crud'");
+    const rs = await this.db.getAll<{ seq: number }>("SELECT seq FROM main.sqlite_sequence WHERE name = 'ps_crud'");
     if (!rs.length) {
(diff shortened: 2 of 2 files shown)


stevensJourney:
> Great find! I would never have guessed this could be the cause.
> 
> Coukd there be any other places in our codebase affected by this type of issue? I.e. any other queries where we should add the `main` prefix to be safe?

This certainly had me scratching my head for a while before I found it.

The only other thing in this codebase I could think of are the queries to `sqlite_master` [here](https://github.com/journeyapps/powersync-react-native-sdk/blob/f3ea14653097ff1aa7c1e4cd1c56aa37dc512a1c/packages/common/src/client/sync/bucket/SqliteBucketStorage.ts#L46) and [here](https://github.com/journeyapps/powersync-react-native-sdk/blob/f3ea14653097ff1aa7c1e4cd1c56aa37dc512a1c/packages/common/src/client/AbstractPowerSyncDatabase.ts#L1155) - however in testing it seems like querying `sqlite_master` includes the data for the main database.
There might be some cases in the Rust core - I'm not very familiar with that codebase though, I could not see any glaring issues at a glance.

PR #784, opened 2025-11-26
Isolate `demos` folder

LucDeCaf:
Currently, the demos in the `demos` folder use pnpm's workspace feature to link to workspace packages. This is convenient for testing changes to workspace packages, as demos will automatically use the local version - however, it causes issues for other users of the demos, as they are required to install _all_ demos and _all_ packages' dependencies in order to run even a single demo. Also, copying demos out of the monorepo requires one to figure out the correct package versions manually.

This PR replaces the `@powersync/*` package versions from `workspace:*` to the latest published releases and adds a `pnpm-workspace.yaml` file to each demo. This isolates the demos from the rest of the workspace and ensures that each demo can run independently from the `powersync-js` repo's context. For convenience, this PR also adds a set of scripts for re-linking the `@powersync/*` packages to the workspace (for ease of development and testing) and for upgrading `@powersync/*` package versions across demos.

Importantly, this PR also includes a partial rewrite of the `react-native-supabase-group-chat` demo, which was using old packages and failing to build on my system. [The PR which made these changes](#789) was merged into this branch, since it also used published PowerSync package versions instead of workspace versions.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -55,7 +55,7 @@ importers:
         version: 3.7.4
       prettier-plugin-embed:
         specifier: ^0.4.15
-        version: 0.4.15
+        version: 0.4.15(babel-plugin-macros@3.1.0)
       prettier-plugin-sql:
         specifier: ^0.18.1
         version: 0.18.1(prettier@3.7.4)
@@ -622,18 +622,24 @@ importers:
 
   tools/diagnostics-app:
     dependencies:
+      '@emotion/react':
+        specifier: ^11.14.0
+        version: 11.14.0(@types/react@18.3.1)(react@18.3.1)
+      '@emotion/styled':
+        specifier: ^11.14.1
+        version: 11.14.1(@emotion/react@11.14.0(@types/react@18.3.1)(react@18.3.1))(@types/react@18.3.1)(react@18.3.1)
       '@journeyapps/wa-sqlite':
         specifier: ^1.4.0
         version: 1.4.1
       '@mui/icons-material':
         specifier: ^7.3.5
-        version: 7.3.6(@mui/material@5.18.0(@types/react@18.3.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@types/react@18.3.1)(react@18.3.1)
+        version: 7.3.6(@mui/material@5.18.0(@emotion/react@11.14.0(@types/react@18.3.1)(react@18.3.1))(@emotion/styled@11.14.1(@emotion/react@11.14.0(@types/react@18.3.1)(react@18.3.1))(@types/react@18.3.1)(react@18.3.1))(@types/react@18.3.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@types/react@18.3.1)(react@18.3.1)
       '@mui/material':
         specifier: ^5.15.12
-        version: 5.18.0(@types/react@18.3.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
+        version: 5.18.0(@emotion/react@11.14.0(@types/react@18.3.1)(react@18.3.1))(@emotion/styled@11.14.1(@emotion/react@11.14.0(@types/react@18.3.1)(react@18.3.1))(@types/react@18.3.1)(react@18.3.1))(@types/react@18.3.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
       '@mui/x-data-grid':
         specifier: ^6.19.6
-        version: 6.20.4(@mui/material@5.18.0(@types/react@18.3.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@mui/system@5.18.0(@types/react@18.3.1)(react@18.3.1))(@types/react@18.3.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
+        version: 6.20.4(@mui/material@5.18.0(@emotion/react@11.14.0(@types/react@18.3.1)(react@18.3.1))(@emotion/styled@11.14.1(@emotion/react@11.14.0(@types/react@18.3.1)(react@18.3.1))(@types/react@18.3.1)(react@18.3.1))(@types/react@18.3.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@mui/system@5.18.0(@emotion/react@11.14.0(@types/react@18.3.1)(react@18.3.1))(@emotion/styled@11.14.1(@emotion/react@11.14.0(@types/react@18.3.1)(react@18.3.1))(@types/react@18.3.1)(react@18.3.1))(@types/react@18.3.1)(react@18.3.1))(@types/react@18.3.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
       '@powersync/react':
         specifier: workspace:*
         version: link:../../packages/react
@@ -800,7 +806,7 @@ importers:
         version: 0.78.0(@babel/core@7.28.5)(@babel/preset-env@7.28.5(@babel/core@7.28.5))
       '@react-native/eslint-config':
         specifier: 0.78.0
-        version: 0.78.0(eslint@8.57.1)(jest@29.7.0(@types/node@22.19.3))(prettier@2.8.8)(typescript@5.0.4)
+        version: 0.78.0(eslint@8.57.1)(jest@29.7.0(@types/node@22.19.3)(babel-plugin-macros@3.1.0))(prettier@2.8.8)(typescript@5.0.4)
       '@react-native/metro-config':
         specifier: 0.78.0
         version: 0.78.0(@babel/core@7.28.5)(@babel/preset-env@7.28.5(@babel/core@7.28.5))
@@ -842,7 +848,7 @@ importers:
         version: 4.1.0
       detox:
         specifier: ^20.34.4
-        version: 20.46.3(@jest/environment@29.7.0)(@jest/types@29.6.3)(expect@29.7.0)(jest-environment-node@29.7.0)(jest@29.7.0(@types/node@22.19.3))
+        version: 20.46.3(@jest/environment@29.7.0)(@jest/types@29.6.3)(expect@29.7.0)(jest-environment-node@29.7.0)(jest@29.7.0(@types/node@22.19.3)(babel-plugin-macros@3.1.0))
       eslint:
         specifier: ^8.19.0
         version: 8.57.1
@@ -851,7 +857,7 @@ importers:
         version: 3.3.0
       jest:
         specifier: ^29.7.0
-        version: 29.7.0(@types/node@22.19.3)
+        version: 29.7.0(@types/node@22.19.3)(babel-plugin-macros@3.1.0)
       prettier:
         specifier: 2.8.8
         version: 2.8.8
@@ -860,7 +866,7 @@ importers:
         version: 19.0.0(react@19.0.0)
       ts-jest:
         specifier: ^29.2.6
-        version: 29.4.6(@babel/core@7.28.5)(@jest/transform@29.7.0)(@jest/types@29.6.3)(babel-jest@29.7.0(@babel/core@7.28.5))(jest-util@29.7.0)(jest@29.7.0(@types/node@22.19.3))(typescript@5.0.4)
+        version: 29.4.6(@babel/core@7.28.5)(@jest/transform@29.7.0)(@jest/types@29.6.3)(babel-jest@29.7.0(@babel/core@7.28.5))(jest-util@29.7.0)(jest@29.7.0(@types/node@22.19.3)(babel-plugin-macros@3.1.0))(typescript@5.0.4)
       typescript:
         specifier: 5.0.4
         version: 5.0.4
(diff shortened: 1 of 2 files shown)


simolus3:
Can we add demo lockfiles to `.gitignore`? I imagine they make up the bulk of this PR, and I'm not sure they're that helpful (we should definitely keep the root lockfile to ensure we're all using the same version of tools).

demos/django-react-native-todolist/metro.config.js
```diff
@@ -4,19 +4,14 @@ const path = require('node:path');
...
-config.resolver.nodeModulesPaths = [
-  path.resolve(projectRoot, 'node_modules'),
-  path.resolve(workspaceRoot, 'node_modules')
-];
+config.resolver.nodeModulesPaths = [path.resolve(projectRoot, 'node_modules')];
 // #3 - Force resolving nested modules to the folders below
-config.resolver.disableHierarchicalLookup = true;
-config.resolver.unstable_enableSymlinks = true;
+// config.resolver.disableHierarchicalLookup = true;
+// config.resolver.unstable_enableSymlinks = true;
```
Chriztiaan: Let's queue up a test, we need to make sure switching to workspace dependencies still works for RN demos with these changes removed/commented out.

demos/angular-supabase-todolist/package.json
```diff
@@ -23,8 +23,8 @@
     "@angular/platform-browser-dynamic": "^19.2.4",
     "@angular/router": "^19.2.4",
     "@angular/service-worker": "^19.2.4",
-    "@journeyapps/wa-sqlite": "^1.4.0",
-    "@powersync/web": "workspace:*",
+    "@journeyapps/wa-sqlite": "^1.3.2",
```
Chriztiaan: This downgrade (across a few of the demos) is probably due to a bad merge conflict resolution.

LucDeCaf: Yep, fixed

demos/django-react-native-todolist/metro.config.js
LucDeCaf: Seems to work after regenerating `ios` and `.expo` dirs

demos/example-capacitor/package.json
```diff
@@ -25,20 +25,20 @@
     "@capacitor/core": "latest",
     "@capacitor/ios": "^7.4.3",
     "@capacitor/splash-screen": "latest",
-    "@journeyapps/wa-sqlite": "^1.4.0",
-    "@powersync/capacitor": "workspace:*",
-    "@powersync/react": "workspace:*",
-    "@powersync/web": "workspace:*",
+    "@powersync/capacitor": "^0.1.3",
+    "@journeyapps/wa-sqlite": "^1.3.2",
```
Chriztiaan: Revert version please

demos/example-electron/package.json
```diff
@@ -21,14 +21,14 @@
   "dependencies": {
     "@emotion/react": "^11.13.0",
     "@emotion/styled": "^11.13.0",
-    "@journeyapps/wa-sqlite": "^1.4.0",
+    "@journeyapps/wa-sqlite": "^1.3.2",
```
Chriztiaan: Revert

demos/react-multi-client/package.json
```diff
@@ -10,9 +10,9 @@
     "test:build": "pnpm build"
   },
   "dependencies": {
-    "@journeyapps/wa-sqlite": "^1.4.0",
-    "@powersync/react": "workspace:*",
-    "@powersync/web": "workspace:*",
+    "@journeyapps/wa-sqlite": "^1.3.2",
```
Chriztiaan: Revert

demos/react-native-supabase-todolist/package.json
```diff
@@ -4,17 +4,18 @@
...
   },
   "dependencies": {
     "@azure/core-asynciterator-polyfill": "^1.0.2",
     "@expo/vector-icons": "^14.0.3",
-    "@journeyapps/react-native-quick-sqlite": "^2.5.0",
-    "@powersync/attachments": "workspace:*",
-    "@powersync/common": "workspace:*",
-    "@powersync/react": "workspace:*",
-    "@powersync/react-native": "workspace:*",
+    "@journeyapps/react-native-quick-sqlite": "^2.4.9",
```
Chriztiaan: Revert

demos/react-native-web-supabase-todolist/package.json
```diff
@@ -12,22 +12,23 @@
...
     "@expo/metro-runtime": "^4.0.1",
-    "@expo/vector-icons": "^14.0.2",
-    "@journeyapps/react-native-quick-sqlite": "^2.5.0",
-    "@journeyapps/wa-sqlite": "^1.4.0",
-    "@powersync/attachments": "workspace:*",
-    "@powersync/react": "workspace:*",
-    "@powersync/react-native": "workspace:*",
-    "@powersync/web": "workspace:*",
+    "@expo/vector-icons": "^14.0.4",
+    "@journeyapps/react-native-quick-sqlite": "^2.4.9",
```
Chriztiaan: Revert

demos/react-supabase-todolist-optional-sync/package.json
```diff
@@ -9,14 +9,14 @@
     "start": "pnpm build && pnpm preview"
   },
   "dependencies": {
+    "@powersync/react": "^1.8.2",
+    "@powersync/web": "^1.29.1",
     "@emotion/react": "11.11.4",
     "@emotion/styled": "11.11.5",
-    "@journeyapps/wa-sqlite": "^1.4.0",
+    "@journeyapps/wa-sqlite": "^1.3.2",
```
Chriztiaan: Revert

Chriztiaan (dismissed):
Last bit of reverting version drops, then it can go in.

Chriztiaan:
Happy with the changes.

demos/angular-supabase-todolist/package.json
```diff
@@ -24,7 +24,7 @@
     "@angular/router": "^19.2.4",
     "@angular/service-worker": "^19.2.4",
     "@journeyapps/wa-sqlite": "^1.4.0",
-    "@powersync/web": "workspace:*",
+    "@powersync/web": "^1.29.1",
```
stevensJourney: There appears to be some inconsistency between demos. E.g. here the version is specified as `^1.29.1`, while the `example-capacitor` demo has `^1.30.0`. Should the script be executed again here?

demos/django-react-native-todolist/package.json
```diff
@@ -12,9 +12,9 @@
     "@expo/metro-runtime": "^4.0.1",
     "@expo/vector-icons": "^14.0.0",
     "@journeyapps/react-native-quick-sqlite": "^2.5.0",
-    "@powersync/common": "workspace:*",
-    "@powersync/react": "workspace:*",
-    "@powersync/react-native": "workspace:*",
+    "@powersync/common": "^1.43.1",
```
stevensJourney: I see it was here before, but, do we need the `@powersync/common` dependency here? I'd hope it would be resolved by `@powersync/react-native`

demos/example-electron-node/src/types.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'better-sqlite3';
```
stevensJourney, line 1: Could we use `@types/better-sqlite3` for this?

demos/react-native-barebones-opsqlite/package.json
```diff
@@ -11,27 +11,29 @@
...
     "@azure/core-asynciterator-polyfill": "^1.0.2",
     "@op-engineering/op-sqlite": "^14.0.2",
-    "@powersync/op-sqlite": "workspace:*",
-    "@powersync/common": "workspace:*",
-    "@powersync/react-native": "workspace:*",
-    "@powersync/react": "workspace:*",
+    "@powersync/op-sqlite": "^0.7.16",
+    "@powersync/react": "^1.8.2",
+    "@powersync/react-native": "^1.27.1",
+    "@powersync/common": "^1.43.1",
```
stevensJourney: I suppose we should remove this as a dependency from all demos that might not require it directly.

demos/react-native-supabase-group-chat/src/providers/AuthProvider.tsx
```diff
@@ -59,10 +60,13 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
 
-  useEffect(() => {
-    if (!session) getSession();
-    // if (session && !user) getUser();
-  }, [session, user]);
+  useEffect(
+    () => {
+      if (!session) getSession();
+    },
+    // [session, user]
```
stevensJourney: For my context: Why is the `user` dependency commented out here?

scripts/bump-demo-packages.ts
```diff
@@ -0,0 +1,209 @@
...
+  '--dry-run': boolean;
+};
+
+const DEMOS_DIR = 'demos';
+
+const workspacePackages = await findWorkspacePackages(path.resolve('.'), {
+  patterns: ['./packages/*']
+});
+
+const defaultFlags = (): Flags => {
```
stevensJourney: a bit of a nit: One can use a workspace devDependency for something like [Commander](https://www.npmjs.com/package/commander) or Yargs to do a lot of this CLI work for us.

scripts/demos-use-workspace.ts
```diff
@@ -0,0 +1,166 @@
...
+import { execSync } from 'child_process';
+import * as fs from 'fs';
+import * as path from 'path';
+
+const demosDir = path.resolve('demos');
+
+const workspacePackages = await findWorkspacePackages(path.resolve('.'));
+
+// Function to split user-provided demos into found and not found demos
+const filterDemos = (allDemos: string[], providedDemos: string[]): [string[], string[]] => {
```
stevensJourney: A bit of a nit: The tuple result here, could be in my opinion, a bit neater if we returned a keyed object e.g. `{found: string[], notFound: string[]}`

scripts/demos-use-workspace.ts
```diff
@@ -0,0 +1,166 @@
...
+
+// Main function to read demos directory and process each demo
+const main = () => {
+  const args: string[] = [];
+  const opts = {
+    noInstall: false
+  };
+
+  for (const arg of process.argv.slice(2)) {
+    if (arg === '--no-install') {
```
stevensJourney: Just mentioning again, something like Commander would provide details and documentation for the supported arguments if a user executed this script with `--help` or provided an invalid value.

scripts/demos-use-workspace.ts
```diff
@@ -0,0 +1,166 @@
...
+      opts.noInstall = true;
+    } else {
+      args.push(arg);
+    }
+  }
+
+  const allDemos = fs.readdirSync(demosDir);
+  let demoNames: string[];
+
+  if (args.length > 0) {
```
stevensJourney: Nit: One could perhaps use something like https://www.npmjs.com/package/inquirer for an optional interactive selection (for those of us who are too lazy to type the full demo name).

scripts/demos-use-workspace.ts
```diff
@@ -0,0 +1,166 @@
...
+  for (const demoName of demoNames) {
+    linkDemo(demoName);
+  }
+  console.log('\nDone.');
+
+  if (opts.noInstall) {
+    process.exit(0);
+  }
+
+  console.log('\nInstalling packages...');
```
stevensJourney: I got this error when running the script locally

```
Updating packages...
- @powersync/react: '^1.8.2' => 'workspace:*'
- @powersync/web: '^1.29.1' => 'workspace:*'
Linking tsconfig.demo.json
- Added '{"path":"../../tsconfig.demo.json"}'

Done.

Installing packages...
Scope: all 18 workspace projects
Lockfile is up to date, resolution step is skipped
Packages: -374
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Progress: resolved 0, reused 201, downloaded 0, added 0, done

╭ Warning ────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                 │
│   Ignored build scripts: @journeyapps/wa-sqlite, @swc/core, core-js, core-js-pure, detox, esbuild, swiftlint.   │
│   Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.                        │
│                                                                                                                 │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

Done in 2.5s using pnpm v10.8.0

Done.
undefined:1
/* To learn more about this file see: https://angular.io/config/tsconfig. */
^

SyntaxError: Unexpected token '/', "/* To lear"... is not valid JSON
    at JSON.parse (<anonymous>)
    at linkDemo (file:///Users/stevenontong/Documents/platform_code/powersync/powersync-react-native-sdk/scripts/dist/demos-use-workspace.js:69:31)
    at main (file:///Users/stevenontong/Documents/platform_code/powersync/powersync-react-native-sdk/scripts/dist/demos-use-workspace.js:127:9)
    at file:///Users/stevenontong/Documents/platform_code/powersync/powersync-react-native-sdk/scripts/dist/demos-use-workspace.js:143:1

Node.js v20.9.0
```

It might be related to this JSON in a demo's `tsconfig.json`
image

Could you also please give some context for how this script works.

I'm a bit confused by the current structure. Where:
- The demos have their own `pnpm-workspace.yaml` file which defines the demo folder as its own monorepo (in order to avoid issues from nested workspaces I presume)
- The monorepo root now does not include the `demos` folder as part of its workspace (and we don't change that with this script)
- This script updates the demo's package.json dependencies to use `workspace:*` for packages defined in the root workspace

If I run the script for a single demo like the `react-supabase-todolist` demo. The `@powersync/web` package is indeed located in the `demos/react-supabase-todolist/node_modules` folder - I'm not sure why this happens, since we execute a `pnpm install` command in the repo root which does not include the `demos` folder in the workspace. However, if I clear my node modules for the entire repo with `rm -rf **/node_modules` and do `pnpm install` in the root, then the web sdk is not present in the demo's node_modules folder. If I try to run `pnpm install` in the demo, I get the following (expected) error
```
❯ pnpm install
 ERR_PNPM_WORKSPACE_PKG_NOT_FOUND  In : "@powersync/react@workspace:*" is in the dependencies but no package named "@powersync/react" is present in the workspace
```

This could be problematic, if for example a dev wants to even install or update a different dependency in a demo while developing.

Furthermore. This method of updating the Git tracked `package.json` files with `workspace:*` dependencies would result in changes which are hard to avoid committing during development. For this reason alone, I can say I would not use this script in my personal development. 

I would rather use the `npm link` [command](https://docs.npmjs.com/cli/v9/commands/npm-link) inside the demo, e.g.
```
npm link # in packages/web
npm link @powersync/web # in demo
```
Note the above uses NPM (not PNPM) since PNPM has historically been tricky with linking. PNPM links will also modify tracked Git files. There might be some issues for using NPM with the local PNPM workspace (haven't thoroughly tested this myself)

That being said, if one wants to use PNPM and have linking without modifying tracked source files, it is also possible by using dynamic overrides in a [pnpmfile](https://pnpm.io/pnpmfile)

Update: After typing this comment, I noticed there are instructions for adding the `demos` folder to the root workspace and using `git restore` to clear changes to tracked files. This is probably personal preference, but I still would not want to run `git restore` each time before committing during development.

scripts/bump-demo-packages.ts
```diff
@@ -0,0 +1,209 @@
...
+        message: `Error writing package.json: ${e.message}`
+      };
+    }
+  }
+
+  console.log('Done\n');
+  return result;
+};
+
+const main = () => {
```
stevensJourney: We should add this script to the `changeset:version` script. This would ensure that the demos are updated as part of the release process.

stevensJourney: For added peace-of-mind, once could also do a `changeset version` in the `Test Isolated Demos` action, before running the Verdaccio publishing. This could help catch versioning and package issues sooner.

stevensJourney (commented):
I can see a lot of hard work has gone into this. This is already a large improvement from the previous setup. 
I added some comments for some potential improvements to make it potentially even better.

demos/example-electron-node/src/types.d.ts
LucDeCaf: Currently, it looks like we need to use `"moduleResolution": "bundler"` in `tsconfig.json`, which causes the following error when trying to run the project (even when installing `@types/better-sqlite3`:
Screenshot 2025-12-15 at 14 32 57

demos/react-native-supabase-group-chat/src/providers/AuthProvider.tsx
LucDeCaf: I simply forgot to delete it after testing if it worked.

scripts/demos-use-workspace.ts
LucDeCaf: To clarify on `.pnpmfile.cjs` - is the idea to have a development script that generates a `.pnpmfile.cjs` in each demo which intercepts the installation of `'@powersync/*'` packages by linking to the local versions instead of the published versions?

stevensJourney: On a high-level, that is what would be done. There are a few options to actually implement this. 

The `.pnpmfile.cjs` could contain the relevant dependency overrides in the file, or it could read the overrides from another file - e.g. a json file which contains a Map of package names and paths to each package to be overriden.

Generally, I think having a standard `pnpmfile.cjs` implementation which reads the overrides dynamically from a JSON file is cleaner - that way one does not need to edit JavaScript files to configure which packages are to be linked. 

The linking script could create the pnpmfile in the demo folder, or the file could always be present. I don't really have a strong opinion on this.

scripts/bump-demo-packages.ts
stevensJourney: Nice!

scripts/demos-inject-local.ts
```diff
@@ -0,0 +1,181 @@
...
+  hooks: {
+    readPackage(pkg) {
+      const injectPeers = (manifestPath) => {
+        try {
+          const content = fs.readFileSync(manifestPath, "utf-8");
+          const localPkg = JSON.parse(content);
+          if (localPkg.peerDependencies) {
+            pkg.dependencies = pkg.dependencies || {};
+            Object.keys(localPkg.peerDependencies).forEach((peer) => {
+              if (peer.startsWith("@powersync/")) {
```
stevensJourney: This check might be fine for now, but could cause issues with other packages scoped with `@powersync` which are not from this repo. I'd rather suggest we also only override if the package is found in `localPaths`

LucDeCaf: done 👍

stevensJourney (approved):
I'm very happy with these changes. This is going to make life much more enjoyable for everyone :D

Issue #785, opened 2025-11-26
OPFSCoopSyncVFS causes multi-tab deadlocks in Chrome and Safari


mikecfisher:
## Problem

I was experiencing deadlocks when using multiple browser tabs with `OPFSCoopSyncVFS`. This happened in **both Chrome and Safari** (not just Safari as I initially thought).

### Reproduction Steps

1. Open Tab A with PowerSync app - works fine
2. Open Tab B - works fine  
3. Refresh Tab B - still fine
4. Switch back to Tab A - **both tabs hang/break**

The app would get stuck on "Syncing..." and never recover. Had to close all tabs and start fresh.

### My Setup

- `@powersync/web` (latest, updated packages recently)
- Next.js 16 with Turbopack
- Manual worker setup (copying from node_modules to public/ since I can't use webpack)

### What I Tried

1. **Safari-only IDBBatchAtomicVFS** - Detected Safari and used IDB for Safari, OPFS for Chrome
   - Result: Safari worked, Chrome still broke

2. **enableMultiTabs: false** - Disabled SharedWorker coordination entirely
   - Result: Second tab would never load data (hung waiting for DB access)

3. **enableMultiTabs: true + IDBBatchAtomicVFS for all browsers** 
   - Result: Works perfectly in both Chrome and Safari!

### The Fix

I stopped using `OPFSCoopSyncVFS` entirely and let it default to `IDBBatchAtomicVFS`:

```ts
const dbFactory = new WASQLiteOpenFactory({
  dbFilename: 'app.db',
  flags: { enableMultiTabs: true },
  // No vfs specified = uses default IDBBatchAtomicVFS
  worker: '/powersync/worker/WASQLiteDB.umd.js',
});

const db = new PowerSyncDatabase({
  database: dbFactory,
  flags: { enableMultiTabs: true },
  schema: MySchema,
  sync: { worker: '/powersync/worker/SharedSyncImplementation.umd.js' },
});
```

### Root Cause

Seems like OPFS file locking doesn't play well with SharedWorker coordination when tabs fight for sync leadership. When you switch from Tab B back to Tab A, something about reclaiming the sync leader role causes an OPFS lock deadlock.

IDBBatchAtomicVFS (IndexedDB) handles concurrent access better and doesn't have this issue.

Any thoughts on this and am missing something? I didn't see much in the docs regarding this. 
[...]

Issue #664, opened 2025-07-11
intermittent error occurs when using OPFSCoopSyncVFS


ImSingee:
An intermittent error occurs when using OPFSCoopSyncVFS:

```
Failed to execute 'createSyncAccessHandle' on 'FileSystemFileHandle': Access Handles cannot be created if there is another open Access Handle or Writable stream associated with the same file.
```

This error has been observed in all versions since OPFS support was introduced.

The issue is intermittent and seems more likely to occur under the following conditions:

- Leaving the page open in the background for several hours, especially after the computer has been asleep.
- Quickly switching away from the page and then back to it

The error can occur with either a single tab or multiple tabs open (with the same powersync db).

If there's more than one tabs open, when the error occurs, all other pages can no longer execute SQL queries, and the sync status gets stuck on "connecting".

The only way to recover is to close all application tabs and reopen them.

michaelbarnes:
Thanks @ImSingee we'll look into it.

Chriztiaan:
Hey @ImSingee, I've been having trouble reproducing this issue on my side.
Are you using the latest versions of our Web SDK? If you already are it would be good to get a bit more information about the error. Is there any more context (or stack traces) when you hit this problem?

ImSingee:
Hi @Chriztiaan, Thanks for your follow-up!

This happens very rarely (less than once a month). I didn't set up a trace for this previously. I've set up the Sentry trace now, and if this issue occurs again, I will do my best to post more information here.

stevensJourney:
This should be fixed by updating `@journeyapps/wa-sqlite` to a version above `1.4.1`

PR #786, opened 2025-11-26
[Fix] OPFS Multitab Deadlocks

Chriztiaan:
## Summary

This PR fixes potential deadlocks in the shared sync implementation when using OPFS (Origin Private File System) with multiple browser tabs. The solution introduces a distributed database adapter that automatically handles connection reconfiguration when tabs close, eliminating the need for explicit reconnection logic in many scenarios.

## The Problem

When using OPFS with multiple tabs, we use a shared web worker for processing sync events. The shared web worker shares a SQLite connection from one of the tabs. For OPFS, we have a connection per tab, and tabs connect to the shared sync worker. The shared sync worker requests a connection from a tab when needed.

Historically, we requested a connection from the last connected tab. When that tab closed, it would inform the shared sync worker (or a navigator lock would detect the tab closed), and we would then disconnect and reconnect using a DB connection from another tab. This approach worked in normal scenarios, but proved to be very racy when tabs were opening and closing at extremely fast rates. The race conditions could lead to deadlocks, especially in Safari with OPFS, where the timing of tab lifecycle events could cause the sync worker to get stuck waiting for a connection that would never respond.

### Related issues

- https://github.com/powersync-ja/powersync-js/issues/785
   - This should be mostly resolved with the fixes here. Safari still has some freezing tabs issues which this does not solve entirely.
- https://github.com/powersync-ja/powersync-js/issues/664
   - Should be solved by this work. The shared worker will reopen it's connection if this is detected. Note that individual tabs will not recover if their connection failed to open with this error - we could improve this. 

## The Solution

This PR removes the direct need for reconnection in many circumstances when a tab closes. Instead of the sync implementation managing connections directly, we now use a **distributed DB adapter** (`LockedAsyncDatabaseAdapter`) in the shared sync worker. This distributed adapter serves as the DBAdapter API surface for the sync implementation and handles reconfiguring the internal connection from another tab if the current tab closes.

DB reconnection is now transparent to the sync implementation. If a reconnect happens during an idle period of sync, the sync implementation will be completely unaware and can proceed normally. If an operation is attempted while reconnecting, the sync manager will detect the error and retry after a delay, just as it does for all errors. This eliminates many of the race conditions that occurred during rapid tab open/close scenarios and provides better error handling with improved error serialization for connection closed scenarios.

It's important to note that this PR does not yet address suspected frozen tab issues in Safari. While random connection selection helps avoid relying on potentially frozen tabs, additional work may be needed to fully handle Safari's tab freezing behavior.

### The Distributed Database Adapter

At the heart of the solution is the `LockedAsyncDatabaseAdapter`, which was already in use and had pluggable ability for opening connections. We've now extended this adapter to support automatically reopening the database connection on specific errors. The adapter detects `ConnectionClosedError` and `NoModificationAllowedError` during operations and automatically re-opens the database connection from another available tab when needed. This provides a stable API surface that hides connection lifecycle details from the sync implementation, allowing it to operate as if it always has a reliable connection.

It's important to note that the distributed database adapter reconnect logic (`reOpenOnConnectionClosed: true`) is only used by the shared sync implementation. Single-tab implementations and other database adapters do not use this automatic reconnection feature, as they don't face the same multi-tab connection challenges.

### Shared Sync Implementation

The shared sync implementation has been refactored to use `LockedAsyncDatabaseAdapter` instead of directly managing connections. The adapter is initialized with `reOpenOnConnectionClosed: true` to enable automatic reconnection, and connection selection is now randomized using `getRandomWrappedPort()` instead of always using the last port. This randomization is particularly important because in some cases, like Safari, tabs can be frozen by the browser, and random selection ensures we don't rely on a potentially frozen tab, improving reliability across different tab scenarios.

We've also tightened the setup flow for the shared sync worker. The sync manager now only uses tabs once the navigator lock-based close detection has been configured. This ensures that tabs which are opened and closed very quickly (closed before the lock could have been configured) will never be used by the sync implementation, since we can't reliably detect if they died. The client is only added to the `SharedSyncImplementation` after the navigator lock has been successfully registered with the shared worker, creating a safety gate that prevents dead tabs from entering the system.

### Connection Management Improvements

The `WorkerWrappedAsyncDatabaseConnection` now properly handles unexpected remote closures. While `markRemoteClosed()` existed before to immediately abort pending operations when a tab closes, we've now added detection if the tab was closed while we were busy configuring or opening the database. This prevents operations from hanging indefinitely during the critical database initialization phase. We've also added better timeout handling for database connection operations with a 10-second timeout, and improved abort controller management for pending operations to ensure resources are properly cleaned up.

Additionally, we've now put the DB open operations in the same Navigator lock as used by the specific write/read operations. This prevents write conflicts when opening databases while another tab might be writing, ensuring that database initialization and operations are properly serialized across tabs.

We've also created a temporary fork of the OPFS Coop VFS. It appears that in some circumstances the initialization logic could grab hold of a navigator lock and not release it, which could contribute to deadlock scenarios. The fork addresses this potential lock retention issue.

### Error Handling Enhancements

Error handling has been significantly improved throughout the system. Previously, some errors such as DOM exceptions were not serializable and were not reported from the shared sync worker to main tabs. We now handle these cases better when doing JSON serialization of errors in sync statuses, ensuring all errors can be properly transmitted across worker boundaries. Upload and download errors are now serialized for `SyncStatus` events, and we've implemented better error propagation and retry logic. Connection closed errors are now properly caught and trigger reconnection automatically.

### Key Files Modified

The main changes are spread across several key files:

- `packages/web/src/worker/sync/SharedSyncImplementation.ts` - Main sync implementation using distributed adapter
- `packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts` - Distributed adapter with auto-reconnection
- `packages/web/src/db/adapters/WorkerWrappedAsyncDatabaseConnection.ts` - Improved remote closure handling
- `packages/web/src/worker/db/opfs.ts` - Temporary fork of OPFS Coop VFS to fix potential navigator lock retention issues
- `packages/common/src/db/crud/SyncStatus.ts` - Error serialization improvements
- `packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts` - Retry logic improvements

### Testing

This PR includes comprehensive test coverage to ensure the solution works correctly. We've added a multiple tabs test (`multiple_tabs_iframe.test.ts`) that tests multi-tab scenarios with rapid open/close operations. We've also created new mock sync service tests with testing utilities for isolated sync service testing, and added error serialization tests to ensure errors are properly serialized across worker boundaries. The existing multi-instance tests have been updated to use the new mock sync service.

### DIFF
--- a/packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
+++ b/packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
@@ -234,7 +234,14 @@ export class LockedAsyncDatabaseAdapter
    */
   async close() {
     this.closing = true;
-    this._disposeTableChangeListener?.();
+    /**
+     * Note that we obtain a reference to the callback to avoid calling the callback with `this` as the context.
+     * This is to avoid Comlink attempting to clone `this` when calling the method.
+     */
+    const dispose = this._disposeTableChangeListener;
+    if (dispose) {
+      dispose();
+    }
     this.pendingAbortControllers.forEach((controller) => controller.abort('Closed'));
     await this.baseDB?.close?.();
     this.closed = true;
--- a/.changeset/fix-comlink-dispose-listener.md
+++ b/.changeset/fix-comlink-dispose-listener.md
@@ -1,0 +1,5 @@
+---
+"@powersync/web": patch
+---
+
+Avoid binding `this` when disposing table change listeners in the web adapter to prevent Comlink serialization errors on close.
--- a/packages/web/tests/locked_adapter_close.test.ts
+++ b/packages/web/tests/locked_adapter_close.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from 'vitest';
+import { LockedAsyncDatabaseAdapter } from '../src/db/adapters/LockedAsyncDatabaseAdapter';
+import { AsyncDatabaseConnection } from '../src/db/adapters/AsyncDatabaseConnection';
+import { DEFAULT_CACHE_SIZE_KB, TemporaryStorageOption } from '../src/db/adapters/web-sql-flags';
+import { ResolvedWASQLiteOpenFactoryOptions } from '../src/db/adapters/wa-sqlite/WASQLiteOpenFactory';
+import { WASQLiteVFS } from '../src/db/adapters/wa-sqlite/WASQLiteConnection';
+
+describe('LockedAsyncDatabaseAdapter.close', () => {
+  it('calls the table change disposer without binding this', async () => {
+    let thisValue: unknown;
+    let called = false;
+
+    const disposer = function (this: unknown) {
+      thisValue = this;
+      called = true;
+    };
+
+    const config: ResolvedWASQLiteOpenFactoryOptions = {
+      dbFilename: 'test.db',
+      flags: {
+        broadcastLogs: true,
+        disableSSRWarning: false,
+        enableMultiTabs: false,
+        useWebWorker: false,
+        ssrMode: false
+      },
+      temporaryStorage: TemporaryStorageOption.MEMORY,
+      cacheSizeKb: DEFAULT_CACHE_SIZE_KB,
+      vfs: WASQLiteVFS.OPFSCoopSyncVFS
+    };
+
+    const db: AsyncDatabaseConnection = {
+      init: async () => {},
+      close: async () => {},
+      markHold: async () => 'hold',
+      releaseHold: async () => {},
+      isAutoCommit: async () => true,
+      execute: async () => ({ rows: { _array: [], length: 0 } } as any),
+      executeRaw: async () => [],
+      executeBatch: async () => ({ rows: { _array: [], length: 0 } } as any),
+      registerOnTableChange: async () => disposer,
+      getConfig: async () => config
+    };
+
+    const adapter = new LockedAsyncDatabaseAdapter({
+      name: 'test-close',
+      openConnection: async () => db
+    });
+
+    await adapter.init();
+    await adapter.close();
+
+    expect(called).toBe(true);
+    expect(thisValue).toBeUndefined();
+  });
+});


packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
```diff
@@ -196,7 +197,7 @@ export class LockedAsyncDatabaseAdapter
     return this.acquireLock(
       async () => fn(this.generateDBHelpers({ execute: this._execute, executeRaw: this._executeRaw })),
       {
-        timeoutMs: options?.timeoutMs
+        timeoutMs: options?.timeoutMs ?? this.options.defaultLockTimeoutMs
```
stevensJourney, line 270: The timeouts here are meant to help expose potentially held locks. E.g. in the past we've seen some DBAdapters created in the shared sync worker - which were not completely closed - could hold on to a lock indefinitely. 
Without a timeout it just seems like the whole process is stuck. With a timeout, we'll at least get some indication of this.

packages/web/src/db/sync/SharedWebStreamingSyncImplementation.ts
```diff
@@ -157,10 +175,47 @@ export class SharedWebStreamingSyncImplementation extends WebStreamingSyncImplem
...
+     *    - The client side lock is held until the client is disposed.
+     *    - We resolve the top-level promise after the lock has been registered with the shared worker.
+     * - The client sends the params to the shared worker after locks have been registered.
+     */
+    await new Promise<void>((resolve) => {
+      // Request a random lock until this client is disposed. The name of the lock is sent to the shared worker, which
+      // will also attempt to acquire it. Since the lock is returned when the tab is closed, this allows the share worker
+      // to free resources associated with this tab.
+      // We take hold of this lock as soon-as-possible in order to cater for potentially closed tabs.
+      getNavigatorLocks().request(`tab-close-signal-${crypto.randomUUID()}`, async (lock) => {
```
stevensJourney, line 200: The initialisation flow here is meant to assist the Shared Sync worker to only use clients which have a navigator-lock based close detection established. We don't want to use clients which might have been registered, but closed before the lock has been configured.

packages/web/src/worker/sync/SharedSyncImplementation.ts
```diff
@@ -321,42 +341,53 @@ export class SharedSyncImplementation extends BaseObserver<SharedSyncImplementat
...
+       * connect operations coming from other tabs.
+       * Re-index subscriptions, the subscriptions of the removed port would no longer be considered.
+       */
       if (shouldReconnect) {
-        await this.connectionManager.connect(CONNECTOR_PLACEHOLDER, this.lastConnectOptions ?? {});
+        this.logger.debug(`Disconnecting due to closed database: should reconnect: ${shouldReconnect}`);
+        this.dbAdapter = null;
+        // Unconditionally close the connection because the database it's writing to has just been closed.
+        // The connection has been closed previously, this might throw. We should be able to ignore it.
+        await this.connectionManager
```
stevensJourney: Starting the disconnect and reconnect inside the `portMutex` lock should help prevent race conditions between multiple tabs potentially requesting connection while we're busy handing this removal operation.

packages/web/tests/mocks/MockWebRemote.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Mock WebRemote that throws 401 Unauthorized errors for all HTTP requests.
+ * Used for testing error handling in the shared sync worker.
+ * Other tests may override this for managed streams.
+ */
+class MockFetchProvider extends FetchImplementationProvider {
+  getFetch(): FetchImplementation {
+    // Return a mock fetch that always returns 401
+    return async (input: RequestInfo | URL, init?: RequestInit): Promise<Response> => {
+      const response = new Response(null, {
```
stevensJourney: This is just a basic mock for now. The 401 should help invalidate any potentially cached credentials for tests. In future we could use a similar, but extended, approach to mock streaming connections in shared workers.

stevensJourney: This mock has now been updated to allow for controlled responses.

packages/web/src/worker/sync/SharedSyncImplementation.ts
stevensJourney: This is no longer applicable after using a distributed database adapter

packages/common/src/db/crud/SyncStatus.ts
```diff
@@ -250,13 +250,28 @@ export class SyncStatus {
...
+        uploadError: this.serializeError(this.dataFlowStatus.uploadError),
+        downloadError: this.serializeError(this.dataFlowStatus.downloadError)
+      },
       lastSyncedAt: this.lastSyncedAt,
       hasSynced: this.hasSynced,
       priorityStatusEntries: this.priorityStatusEntries
     };
   }
 
+  protected serializeError(error?: Error) {
```
stevensJourney, line 268: Not all errors are serialisable over a MessagePort. E.g. some `DomException`s fail to be passed across workers. This explicitly serialises errors in the `SyncStatus` now.

packages/web/src/db/PowerSyncDatabase.ts
```diff
@@ -160,14 +159,13 @@ export class PowerSyncDatabase extends AbstractPowerSyncDatabase {
...
    */
-  close(options: PowerSyncCloseOptions = DEFAULT_POWERSYNC_CLOSE_OPTIONS): Promise<void> {
+  close(options?: PowerSyncCloseOptions): Promise<void> {
     if (this.unloadListener) {
       window.removeEventListener('unload', this.unloadListener);
     }
-
     return super.close({
       // Don't disconnect by default if multiple tabs are enabled
-      disconnect: options.disconnect ?? !this.resolvedFlags.enableMultiTabs
```
stevensJourney, line 170: The default param above actually causes the opposite of this to occur. A `close` would perform a disconnect on web if no argument was supplied. Removing the default fixes this.

packages/web/src/worker/db/opfs.ts
```diff
@@ -0,0 +1,600 @@
...
+                const persistentFile = this.persistentFiles.get(file.path + suffix);
+                if (persistentFile) {
+                  persistentFile.accessHandle = await persistentFile.fileHandle.createSyncAccessHandle();
+                }
+              })
+            );
+          } catch (e) {
+            this.log?.(`failed to create access handles for ${file.path}`, e);
+            // Release the lock, if we failed here, we'd need to obtain the lock later in order to retry
+            file.persistentFile.handleLockReleaser();
```
stevensJourney: This is the only change in this file. When frequently refeshing a page, we noticed 

```
[PowerSyncDatabase [example.dbll Error: Failed to execute
'createSyncAccessHandle'
on 'FileSystemFileHandle': Access Handles cannot be created if
there is another open Access Handle or Writable stream associated with the same file.
```

It seems like if this error happened during page load (not entirely sure why this error did happen at this stage), the lock would never be released - which would prevent other tabs from using the DB.

package.json
```diff
@@ -49,5 +49,10 @@
     "rollup-plugin-dts": "^6.2.1",
     "typescript": "^5.7.2",
     "vitest": "^3.2.4"
+  },
+  "pnpm": {
+    "overrides": {
+      "@journeyapps/wa-sqlite": "0.0.0-dev-20251201120934"
```
stevensJourney: TODO remove this

packages/web/src/worker/db/opfs.ts
stevensJourney: moved to https://github.com/powersync-ja/wa-sqlite/pull/56

demos/react-supabase-todolist/src/components/providers/SystemProvider.tsx
```diff
@@ -12,8 +19,15 @@ export const useSupabase = () => React.useContext(SupabaseContext);
 
 export const db = new PowerSyncDatabase({
   schema: AppSchema,
-  database: {
-    dbFilename: 'example.db'
+  database: new WASQLiteOpenFactory({
```
Chriztiaan, line 24: Do we want to check this in or have a commented out example?

stevensJourney: I think it would be nice to have an example/demo of OPFS being used.

packages/web/src/worker/sync/SharedSyncImplementation.ts
```diff
@@ -171,6 +169,32 @@ export class SharedSyncImplementation extends BaseObserver<SharedSyncImplementat
...
+      }
+      return;
+    });
+  }
+
+  /**
+   * In some very rare cases a specific tab might not respond to requests.
+   * This returns a random port which is not closing.
+   */
+  protected async getRandomWrappedPort(): Promise<WrappedSyncPort | undefined> {
```
simolus3, line 191: The PR description mentions

> This randomization is particularly important because in some cases, like Safari, tabs can be frozen by the browser, and random selection ensures we don't rely on a potentially frozen tab, improving reliability across different tab scenarios.

Am I right to assume that "ensures we don't rely on a potentially frozen tab" is a bit optimistic here because we could still hit a frozen tab due to bad luck?
This also wouldn't completely fix the issue, but what if we send a ping (or call a method doing nothing through comlink) to all non-closing tabs concurrently and pick the first one that responds?

stevensJourney: Indeed. This by no means is a complete solution. Properly mitigating frozen tabs (to the best of our ability) is something for future work. 
The changes here should help for some cases though. Luck certainly is involved, but it's not the main strategy employed. Essentially, if we hit a frozen tab, the `open` operation timeout would be hit and we'd retry - hopefully with a different client. This is a little better than if the last tab was frozen and we constantly tried that one.

packages/common/src/db/crud/SyncStatus.ts
simolus3: Would perhaps be nice to mention that as a reason in a comment on this method or on line 255 where it's used.

packages/web/src/db/adapters/AsyncDatabaseConnection.ts
```diff
@@ -17,6 +17,18 @@ export type ProxiedQueryResult = Omit<QueryResult, 'rows'> & {
  */
 export type OnTableChangeCallback = (event: BatchedUpdateNotification) => void;
 
+/**
+ * Thrown when an underlying database connection is closed.
+ * This is particularly relevant when worker connections are marked as closed while
+ * operations are still in progress.
+ */
+export class ConnectionClosedError extends Error {
```
simolus3: nit: We could consider moving and exporting this from `common`, since we also throw a similar error in the Node.JS SDK [here](https://github.com/powersync-ja/powersync-js/blob/8bf5fd78cd6b96adbb3c9a4c3c587c86ea5d4fde/packages/node/src/db/RemoteConnection.ts#L43).

packages/web/src/worker/sync/SharedSyncImplementation.ts
```diff
@@ -217,33 +241,51 @@ export class SharedSyncImplementation extends BaseObserver<SharedSyncImplementat
...
+      reOpenOnConnectionClosed: true
+    });
+    this.distributedDB = lockedAdapter;
+    await lockedAdapter.init();
+
+    lockedAdapter.registerListener({
+      databaseReOpened: () => {
+        // We may have missed some table updates while the database was closed.
+        // We can poke the crud in case we missed any updates.
+        this.connectionManager.syncStreamImplementation?.triggerCrudUpload();
```
simolus3, line 279: The implementation should be fine as-is since it reconnects on errors, but we could eventually use this to reconnect if we're using the Rust client implementation.

As discussed offline, sync state with the Rust client is inherently bound to a connection. So if the database is reopened, we should ideally reconnect because the client won't know about the connection state otherwise. I don't think this needs to be fixed in this PR, but we could add a TODO comment mentioning the issue.

stevensJourney: Yeah, I can add a todo for this. I believe the sync implementation should automatically reconnect if the Rust client throws an exception (due to the broken state), but it would be good to catch this early since that would only happen when the next sync event is received.
We should also note that we probably don't need to do this reconnect if using IndexedDB VFS (since it would be the same connection).

packages/web/src/db/adapters/AsyncDatabaseConnection.ts
stevensJourney: That's a fair point. Will move this.

stevensJourney:
> The fixes here are clearly a significant improvement, so I'm very happy with them 👍 The iframe / mock sync worker setup also looks impressive.
> 
> One thing I am a bit worried about is the increasing complexity in the sync worker setup. E.g. the Dart implementation in comparison uses an actor-like pattern for the worker to handle all events in order, which makes that code a bit easier to reason about and potentially decreases the likelihood of race conditions (although I'm certain Dart would deadlock on sleeping tabs as well, it's just not seeing enough usage for us to get reports). To ensure all our web SDKs work correctly here, it might be worth eventually restructuring the worker setup and communication protocol a bit to align the Dart and JS implementations (or to use the JS sync worker for the Dart SDK as well, if we have a stable protocol).

I agree that aligning both implementations would be a significant win. The actor approach in the Dart client does have less potential for race conditions.

One consideration with that approach is that if any ordered action were to freeze, the entire actor could remain stuck without strict timeout enforcement. 

Additionally, in some circumstances the ordering of events could counteract the work from https://github.com/powersync-ja/powersync-js/pull/607, which prevents queuing multiple requests. That said, the Dart implementation is less susceptible to this since it only reconnects when the Sync request changes (we should implement this in JS at some point) and doesn't await the connect operation (In my opinion, I like that the JS client does this). 

These are just some points for context. I do agree and think we should use the learnings from both implementations to create a standardized approach going forward.

packages/common/src/db/crud/SyncStatus.ts
stevensJourney: 👍 Added a comment.

packages/web/src/db/adapters/AsyncDatabaseConnection.ts
stevensJourney: Moved this to `@powersync/common`. Updated the Node.JS SDK to use this.

packages/web/src/worker/sync/SharedSyncImplementation.ts
stevensJourney: Added a comment for this 😄

PR #789, opened 2025-12-03
Migrate PowerChat demo

LucDeCaf:
The PowerChat demo (found [here](https://github.com/powersync-ja/powersync-js/tree/main/demos/react-native-supabase-group-chat)) is one of the older React Native demos in the repository. As such, it has naturally become less stable and harder to debug as PowerSync, Expo, React Native and other packages all increase their versions. This naturally leads to wanting to bump the package versions - however, some packages are so far behind that it would frankly be easier to start a new project and copy the application code, while also writing new config files that work with the latest packages.

This PR is a rewrite of the application using the latest compatible package versions and package configurations. Most of the app code is copied and should work the same, but most of the configuration code (Metro config, Babel config, Expo config) is new.

Also see PR #784, which is partially the reason this demo is being updated.


demos/react-native-supabase-group-chat/src/lib/connector.ts
```diff
@@ -1,19 +1,25 @@
-import { AbstractPowerSyncDatabase, CrudEntry, PowerSyncBackendConnector, UpdateType, type PowerSyncCredentials } from '@powersync/react-native';
-import type { SupabaseClient } from '@supabase/supabase-js';
+import {
+  AbstractPowerSyncDatabase,
+  CrudEntry,
+  PowerSyncBackendConnector,
+  UpdateType,
+  type PowerSyncCredentials
+} from "@powersync/react-native";
+import type { SupabaseClient } from "@supabase/supabase-js";
```
Chriztiaan: A few files have the incorrect string formatting.

demos/react-native-supabase-group-chat/package.json
```diff
@@ -20,51 +20,59 @@
...
-    "react-native-reanimated": "~3.16.7",
-    "react-native-safe-area-context": "4.12.0",
-    "react-native-screens": "~4.4.0",
-    "react-native-svg": "15.8.0",
-    "react-native-web": "0.19.13",
-    "tamagui": "1.79.6"
+    "@expo/vector-icons": "^15.0.3",
+    "@faker-js/faker": "^10.1.0",
+    "@journeyapps/react-native-quick-sqlite": "^2.4.11",
+    "@powersync/attachments": "^2.4.1",
```
Chriztiaan: Technically this is stealth changing the dropping of the "workspace" value.
Either we use workspace for this PR agains't main, or we target the the other PR's branch from this PR for now while reviewing.

demos/react-native-supabase-group-chat/src/lib/connector.ts
LucDeCaf: Prettier seems to struggle to detect folders named `lib` - I moved the files to a folder called `library` and it seems to work now.

demos/react-native-supabase-group-chat/package.json
LucDeCaf: Yeah, this should probably point to [isolate demos](/powersync-ja/powersync-js/tree/isolate-demos)

demos/react-native-supabase-group-chat/src/lib/connector.ts
Chriztiaan: Awesome, thanks for fixing that.

PR #794, opened 2025-12-11
fix(drizzle-driver): reorder export attributes preventing "Error: Default condition should be last one"

flochaz:
# Context

While using `@powersync/drizzle-driver` in an imported library, compilation with webpack fail with `Module not found: Error: Default condition should be last one` due to wrong attribute order in package.json of drizzle-driver. This constraint is coming from https://github.com/webpack/enhanced-resolve package.


Chriztiaan:
Hey @flochaz, thank you for the PR! Could you add a patch changeset entry with `pnpm changeset`?

PR #795, opened 2025-12-11
fix(node): load properly better-sqlite3

flochaz:
# Context

Using the default worker fail with bad import of better-sqlite3. Patching `loadBetterSqlite3` function to return the `default` if exposed fixed the issue.

### DIFF
--- a/packages/node/src/db/SqliteWorker.ts
+++ b/packages/node/src/db/SqliteWorker.ts
@@ -84,9 +84,8 @@ export function startPowerSyncWorker(options?: Partial<PowerSyncWorkerOptions>) 
     },
     async loadBetterSqlite3() {
       const module = await dynamicImport('better-sqlite3');
-      return module && typeof module === 'object' && 'default' in module
-        ? module.default
-        : module;    
+      // require() gives us the default directly, for an ESM import() we need to use the default export.
+      return isBundledToCommonJs ? module : module.default;
     },
     ...options
   };


.changeset/fix-better-sqlite3.md
```diff
@@ -0,0 +1,7 @@
+---
+"@powersync/node": patch
+---
+
+Fix loading of `better-sqlite3` in the Node worker so the module is dynamically loaded correctly.
+
+This ensures the Node SDK loads the correct Database constructor when `better-sqlite3` is installed.
```
simolus3:
```suggestion
Fix loading `better-sqlite3` on CommonJS builds.
```

packages/node/src/db/SqliteWorker.ts
```diff
@@ -84,7 +84,9 @@ export function startPowerSyncWorker(options?: Partial<PowerSyncWorkerOptions>)
     },
     async loadBetterSqlite3() {
       const module = await dynamicImport('better-sqlite3');
-      return module.default;
+      return module && typeof module === 'object' && 'default' in module
+        ? module.default
+        : module;
```
simolus3: I think this might make it clearer what's going on:

```suggestion
      // require() gives us the default directly, for an ESM import() we need to use the default export.
      return isBundledToCommonJs ? module : module.default;
```

simolus3: @flochaz You've marked this as resolved but the branch still has the original variant. Is there a reason against this change?

flochaz:
My bad will will fix

Florian Chazal

simolus3:
Released in version `0.15.1` of `@powersync/node`, thanks for your contribution 👍

PR #796, opened 2025-12-12
[Demo] Tanstack db demo update

stevensJourney:
# Fix TanStack DB Demo Module Resolution & Add E2E Tests

## Problem

The TanStack DB demo fails to run with `pnpm dev`, showing the following error:

```
✘ [ERROR] Could not resolve "@powersync/common"

    ../../node_modules/@tanstack/powersync-db-collection/dist/esm/schema.js:1:27:
      1 │ import { ColumnType } from "@powersync/common";
        ╵                            ~~~~~~~~~~~~~~~~~~~
```

### Root Cause

This is caused by pnpm's package hoisting behavior in the monorepo:

1. `@tanstack/powersync-db-collection` is hoisted to the **root** `node_modules` folder
2. It declares `@powersync/common` as a **peer dependency**
3. The demo had `@powersync/common` as a direct `workspace:*` dependency, which symlinked it to the **demo's** `node_modules` only
4. When Vite/esbuild bundles `@tanstack/powersync-db-collection` from the root, it cannot resolve `@powersync/common` because it's not present at the root level

## Solution

Remove the direct `@powersync/common` dependency from the demo. The package is already available transitively through `@powersync/web`, which resolves the hoisting issue.

## Additional Changes

Added end-to-end tests for the demo that run outside of the monorepo context. These tests:

- Verify the demo loads correctly in a Vitest browser mode session
- Perform basic mutations and validate that TanStack DB detects and renders the changes correctly


demos/react-supabase-todolist-tanstackdb/e2e/todo-lists.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+          completed ? MOCK_USER_ID : null
+        ]
+      );
+    });
+
+    return todoId;
+  }
+
+  it('should load the app and show the Todo Lists screen', async () => {
+    await renderAppAndWaitForTodoListsScreen();
```
LucDeCaf: Is it possible to move `await renderAppAndWaitForTodoListsScreen()` into beforeEach? It seems redundant to call it manually in every test.

Chriztiaan:
Tested this on my side, happy with the changes.

Chriztiaan (approved):

demos/react-supabase-todolist-tanstackdb/e2e/todo-lists.test.tsx
stevensJourney: That's a fair point. Good catch.

LucDeCaf (approved):
Tested it, looks good to me.

PR #797, opened 2025-12-15
[Package] PowerSync Nuxt Module

khawarizmus:
This PR introduces a new `nuxt` package to offer first-class support for Nuxt with PowerSync.

The PR also introduces a new demo showcasing a reference implementation showing PowerSync + Nuxt 4 + Supabase

- [x] Add Nuxt module package
- [x] Add PowerSync x Nuxt x Supabase demo
- [x] Write the changset
- [x] Test a dev release outside the repo

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,5 +56,10 @@
     "typescript": "catalog:",
     "vitest": "catalog:"
   },
+  "pnpm": {
+    "overrides": {
+      "fast-xml-parser": ">=5.3.5"
+    }
+  },
   "packageManager": "pnpm@10.27.0+sha512.72d699da16b1179c14ba9e64dc71c9a40988cbdc65c264cb0e489db7de917f20dcf4d64d8723625f2969ba52d4b7e2a1170682d9ac2a5dcaeaab732b7e16f04a"
 }
--- a/audit-ci.jsonc
+++ b/audit-ci.jsonc
@@ -1,6 +1,5 @@
 {
   "critical": true,
   "package-manager": "auto",
-  "registry": "https://registry.npmjs.org",
-  "allowlist": ["GHSA-m7jm-9gc2-mpf2"] // CLI-only dev tooling (cli-doctor), never processes untrusted XML
+  "registry": "https://registry.npmjs.org"
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -106,6 +106,9 @@ catalogs:
       specifier: ^3.2.4
       version: 3.2.4
 
+overrides:
+  fast-xml-parser: '>=5.3.5'
+
 importers:
 
   .:
@@ -178,10 +181,10 @@ importers:
     dependencies:
       '@docusaurus/core':
         specifier: ^3.7.0
-        version: 3.9.2(@docusaurus/faster@3.9.2(@docusaurus/types@3.7.0(@swc/core@1.15.11(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@swc/helpers@0.5.19))(@mdx-js/react@3.1.1(@types/react@19.2.14)(react@18.3.1))(@rspack/core@1.7.6(@swc/helpers@0.5.19))(@swc/core@1.15.11(@swc/helpers@0.5.19))(lightningcss@1.31.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(typescript@5.9.3)
+        version: 3.9.2(@docusaurus/faster@3.9.2(@docusaurus/types@3.7.0(@swc/core@1.15.13(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@swc/helpers@0.5.19))(@mdx-js/react@3.1.1(@types/react@19.2.14)(react@18.3.1))(@rspack/core@1.7.6(@swc/helpers@0.5.19))(@swc/core@1.15.13(@swc/helpers@0.5.19))(lightningcss@1.31.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(typescript@5.9.3)
       '@docusaurus/preset-classic':
         specifier: ^3.7.0
-        version: 3.9.2(@algolia/client-search@5.49.0)(@docusaurus/faster@3.9.2(@docusaurus/types@3.7.0(@swc/core@1.15.11(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@swc/helpers@0.5.19))(@mdx-js/react@3.1.1(@types/react@19.2.14)(react@18.3.1))(@rspack/core@1.7.6(@swc/helpers@0.5.19))(@swc/core@1.15.11(@swc/helpers@0.5.19))(@types/react@19.2.14)(lightningcss@1.31.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(search-insights@2.17.3)(typescript@5.9.3)
+        version: 3.9.2(@algolia/client-search@5.49.0)(@docusaurus/faster@3.9.2(@docusaurus/types@3.7.0(@swc/core@1.15.13(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@swc/helpers@0.5.19))(@mdx-js/react@3.1.1(@types/react@19.2.14)(react@18.3.1))(@rspack/core@1.7.6(@swc/helpers@0.5.19))(@swc/core@1.15.13(@swc/helpers@0.5.19))(@types/react@19.2.14)(lightningcss@1.31.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(search-insights@2.17.3)(typescript@5.9.3)
       '@mdx-js/react':
         specifier: ^3.1.0
         version: 3.1.1(@types/react@19.2.14)(react@18.3.1)
@@ -200,19 +203,19 @@ importers:
     devDependencies:
       '@docusaurus/faster':
         specifier: ^3.7.0
-        version: 3.9.2(@docusaurus/types@3.7.0(@swc/core@1.15.11(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@swc/helpers@0.5.19)
+        version: 3.9.2(@docusaurus/types@3.7.0(@swc/core@1.15.13(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@swc/helpers@0.5.19)
       '@docusaurus/module-type-aliases':
         specifier: ^3.7.0
-        version: 3.9.2(@swc/core@1.15.11(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
+        version: 3.9.2(@swc/core@1.15.13(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
       '@docusaurus/theme-classic':
         specifier: ^3.7.0
-        version: 3.9.2(@docusaurus/faster@3.9.2(@docusaurus/types@3.7.0(@swc/core@1.15.11(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@swc/helpers@0.5.19))(@rspack/core@1.7.6(@swc/helpers@0.5.19))(@swc/core@1.15.11(@swc/helpers@0.5.19))(@types/react@19.2.14)(lightningcss@1.31.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(typescript@5.9.3)
+        version: 3.9.2(@docusaurus/faster@3.9.2(@docusaurus/types@3.7.0(@swc/core@1.15.13(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(@swc/helpers@0.5.19))(@rspack/core@1.7.6(@swc/helpers@0.5.19))(@swc/core@1.15.13(@swc/helpers@0.5.19))(@types/react@19.2.14)(lightningcss@1.31.1)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(typescript@5.9.3)
       '@docusaurus/tsconfig':
         specifier: 3.7.0
         version: 3.7.0
       '@docusaurus/types':
         specifier: 3.7.0
-        version: 3.7.0(@swc/core@1.15.11(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
+        version: 3.7.0(@swc/core@1.15.13(@swc/helpers@0.5.19))(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
       '@types/node':
         specifier: ^20.17.12
         version: 20.19.33
@@ -990,7 +993,7 @@ importers:
         version: 1.6.13(@swc/helpers@0.5.19)
       '@tailwindcss/vite':
         specifier: ^4.1.17
-        version: 4.2.0(vite@5.4.21(@types/node@22.19.11)(lightningcss@1.31.1)(terser@5.46.0))
+        version: 4.2.1(vite@5.4.21(@types/node@22.19.11)(lightningcss@1.31.1)(terser@5.46.0))
       '@tanstack/router-plugin':
         specifier: ^1.153.2
         version: 1.162.4(@tanstack/react-router@1.162.4(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(vite@5.4.21(@types/node@22.19.11)(lightningcss@1.31.1)(terser@5.46.0))(webpack@5.105.2(@swc/core@1.6.13(@swc/helpers@0.5.19)))
@@ -1008,7 +1011,7 @@ importers:
         version: 4.7.0(vite@5.4.21(@types/node@22.19.11)(lightningcss@1.31.1)(terser@5.46.0))
       tailwindcss:
         specifier: ^4.1.17
-        version: 4.2.0
+        version: 4.2.1
       typescript:
         specifier: ^5.5.3
         version: 5.9.3
@@ -3317,8 +3320,8 @@ packages:
   '@expo/sudo-prompt@9.3.2':
     resolution: {integrity: sha512-HHQigo3rQWKMDzYDLkubN5WQOYXJJE2eNqIQC2axC2iO3mHdwnIR7FgZVvHWtBwAdzBgAP0ECp8KqS8TiMKvgw==}
 
-  '@expo/vector-icons@15.0.3':
-    resolution: {integrity: sha512-SBUyYKphmlfUBqxSfDdJ3jAdEVSALS2VUPOUyqn48oZmb2TL/O7t7/PQm5v4NQujYEPLPMTLn9KVw6H7twwbTA==}
+  '@expo/vector-icons@15.1.1':
+    resolution: {integrity: sha512-Iu2VkcoI5vygbtYngm7jb4ifxElNVXQYdDrYkT7UCEIiKLeWnQY0wf2ZhHZ+Wro6Sc5TaumpKUOqDRpLi5rkvw==}
     peerDependencies:
       expo-font: '>=14.0.4'
       react: '*'
@@ -3327,8 +3330,8 @@ packages:
   '@expo/ws-tunnel@1.0.6':
     resolution: {integrity: sha512-nDRbLmSrJar7abvUjp3smDwH8HcbZcoOEa5jVPUv9/9CajgmWw20JNRwTuBRzWIWIkEJDkz20GoNA+tSwUqk0Q==}
 
-  '@expo/xcpretty@4.4.0':
-    resolution: {integrity: sha512-o2qDlTqJ606h4xR36H2zWTywmZ3v3842K6TU8Ik2n1mfW0S580VHlt3eItVYdLYz+klaPp7CXqanja8eASZjRw==}
+  '@expo/xcpretty@4.4.1':
+    resolution: {integrity: sha512-KZNxZvnGCtiM2aYYZ6Wz0Ix5r47dAvpNLApFtZWnSoERzAdOMzVBOPysBoM0JlF6FKWZ8GPqgn6qt3dV/8Zlpg==}
     hasBin: true
 
   '@fastify/busboy@2.1.1':
@@ -6176,8 +6179,8 @@ packages:
     resolution: {integrity: sha512-LnhVjMWyMQV9ZmeEy26maJk+8HTIbd59cH4F2MJ439k9DqejRisfFNGAPvRYlKETuh9LrImlS8aKsBgKjMA8WA==}
     engines: {node: '>=14'}
 
-  '@swc/core-darwin-arm64@1.15.11':
-    resolution: {integrity: sha512-QoIupRWVH8AF1TgxYyeA5nS18dtqMuxNwchjBIwJo3RdwLEFiJq6onOx9JAxHtuPwUkIVuU2Xbp+jCJ7Vzmgtg==}
+  '@swc/core-darwin-arm64@1.15.13':
+    resolution: {integrity: sha512-ztXusRuC5NV2w+a6pDhX13CGioMLq8CjX5P4XgVJ21ocqz9t19288Do0y8LklplDtwcEhYGTNdMbkmUT7+lDTg==}
     engines: {node: '>=10'}
     cpu: [arm64]
     os: [darwin]
@@ -6188,8 +6191,8 @@ packages:
     cpu: [arm64]
     os: [darwin]
 
-  '@swc/core-darwin-x64@1.15.11':
-    resolution: {integrity: sha512-S52Gu1QtPSfBYDiejlcfp9GlN+NjTZBRRNsz8PNwBgSE626/FUf2PcllVUix7jqkoMC+t0rS8t+2/aSWlMuQtA==}
+  '@swc/core-darwin-x64@1.15.13':
+    resolution: {integrity: sha512-cVifxQUKhaE7qcO/y9Mq6PEhoyvN9tSLzCnnFZ4EIabFHBuLtDDO6a+vLveOy98hAs5Qu1+bb5Nv0oa1Pihe3Q==}
     engines: {node: '>=10'}
     cpu: [x64]
     os: [darwin]
@@ -6200,8 +6203,8 @@ packages:
     cpu: [x64]
     os: [darwin]
 
-  '@swc/core-linux-arm-gnueabihf@1.15.11':
-    resolution: {integrity: sha512-lXJs8oXo6Z4yCpimpQ8vPeCjkgoHu5NoMvmJZ8qxDyU99KVdg6KwU9H79vzrmB+HfH+dCZ7JGMqMF//f8Cfvdg==}
+  '@swc/core-linux-arm-gnueabihf@1.15.13':
+    resolution: {integrity: sha512-t+xxEzZ48enl/wGGy7SRYd7kImWQ/+wvVFD7g5JZo234g6/QnIgZ+YdfIyjHB+ZJI3F7a2IQHS7RNjxF29UkWw==}
     engines: {node: '>=10'}
     cpu: [arm]
     os: [linux]
@@ -6212,8 +6215,8 @@ packages:
     cpu: [arm]
     os: [linux]
 
-  '@swc/core-linux-arm64-gnu@1.15.11':
-    resolution: {integrity: sha512-chRsz1K52/vj8Mfq/QOugVphlKPWlMh10V99qfH41hbGvwAU6xSPd681upO4bKiOr9+mRIZZW+EfJqY42ZzRyA==}
+  '@swc/core-linux-arm64-gnu@1.15.13':
+    resolution: {integrity: sha512-VndeGvKmTXFn6AGwjy0Kg8i7HccOCE7Jt/vmZwRxGtOfNZM1RLYRQ7MfDLo6T0h1Bq6eYzps3L5Ma4zBmjOnOg==}
     engines: {node: '>=10'}
     cpu: [arm64]
     os: [linux]
@@ -6224,8 +6227,8 @@ packages:
     cpu: [arm64]
     os: [linux]
 
-  '@swc/core-linux-arm64-musl@1.15.11':
-    resolution: {integrity: sha512-PYftgsTaGnfDK4m6/dty9ryK1FbLk+LosDJ/RJR2nkXGc8rd+WenXIlvHjWULiBVnS1RsjHHOXmTS4nDhe0v0w==}
+  '@swc/core-linux-arm64-musl@1.15.13':
+    resolution: {integrity: sha512-SmZ9m+XqCB35NddHCctvHFLqPZDAs5j8IgD36GoutufDJmeq2VNfgk5rQoqNqKmAK3Y7iFdEmI76QoHIWiCLyw==}
     engines: {node: '>=10'}
     cpu: [arm64]
     os: [linux]
@@ -6236,8 +6239,8 @@ packages:
     cpu: [arm64]
     os: [linux]
 
-  '@swc/core-linux-x64-gnu@1.15.11':
-    resolution: {integrity: sha512-DKtnJKIHiZdARyTKiX7zdRjiDS1KihkQWatQiCHMv+zc2sfwb4Glrodx2VLOX4rsa92NLR0Sw8WLcPEMFY1szQ==}
+  '@swc/core-linux-x64-gnu@1.15.13':
+    resolution: {integrity: sha512-5rij+vB9a29aNkHq72EXI2ihDZPszJb4zlApJY4aCC/q6utgqFA6CkrfTfIb+O8hxtG3zP5KERETz8mfFK6A0A==}
     engines: {node: '>=10'}
     cpu: [x64]
     os: [linux]
@@ -6248,8 +6251,8 @@ packages:
     cpu: [x64]
     os: [linux]
 
-  '@swc/core-linux-x64-musl@1.15.11':
-    resolution: {integrity: sha512-mUjjntHj4+8WBaiDe5UwRNHuEzLjIWBTSGTw0JT9+C9/Yyuh4KQqlcEQ3ro6GkHmBGXBFpGIj/o5VMyRWfVfWw==}
+  '@swc/core-linux-x64-musl@1.15.13':
+    resolution: {integrity: sha512-OlSlaOK9JplQ5qn07WiBLibkOw7iml2++ojEXhhR3rbWrNEKCD7sd8+6wSavsInyFdw4PhLA+Hy6YyDBIE23Yw==}
     engines: {node: '>=10'}
     cpu: [x64]
     os: [linux]
@@ -6260,8 +6263,8 @@ packages:
     cpu: [x64]
     os: [linux]
 
-  '@swc/core-win32-arm64-msvc@1.15.11':
-    resolution: {integrity: sha512-ZkNNG5zL49YpaFzfl6fskNOSxtcZ5uOYmWBkY4wVAvgbSAQzLRVBp+xArGWh2oXlY/WgL99zQSGTv7RI5E6nzA==}
+  '@swc/core-win32-arm64-msvc@1.15.13':
+    resolution: {integrity: sha512-zwQii5YVdsfG8Ti9gIKgBKZg8qMkRZxl+OlYWUT5D93Jl4NuNBRausP20tfEkQdAPSRrMCSUZBM6FhW7izAZRg==}
     engines: {node: '>=10'}
     cpu: [arm64]
     os: [win32]
@@ -6272,8 +6275,8 @@ packages:
     cpu: [arm64]
     os: [win32]
 
-  '@swc/core-win32-ia32-msvc@1.15.11':
-    resolution: {integrity: sha512-6XnzORkZCQzvTQ6cPrU7iaT9+i145oLwnin8JrfsLG41wl26+5cNQ2XV3zcbrnFEV6esjOceom9YO1w9mGJByw==}
+  '@swc/core-win32-ia32-msvc@1.15.13':
+    resolution: {integrity: sha512-hYXvyVVntqRlYoAIDwNzkS3tL2ijP3rxyWQMNKaxcCxxkCDto/w3meOK/OB6rbQSkNw0qTUcBfU9k+T0ptYdfQ==}
     engines: {node: '>=10'}
     cpu: [ia32]
     os: [win32]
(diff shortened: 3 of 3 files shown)


khawarizmus:
Not sure what the chanset should include when releasing a new package

demos/nuxt-supabase-todolist/components/AppHeader.vue
```diff
@@ -0,0 +1,43 @@
...
+
+<template>
+  <UHeader :toggle="false">
+    <template #left>
+      <UButton
+        variant="link"
+        @click="navigateTo('/')"
+      >
+        <img
+          src="https://cdn.prod.website-files.com/67eea61902e19994e7054ea0/67f910109a12edc930f8ffb6_powersync-icon.svg"
```
stevensJourney: Is this link copied from some site, or, where is it managed? Could we use an asset for this?

demos/nuxt-supabase-todolist/README.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+### 2. Create project on Supabase and set up Postgres
+
+This demo app uses Supabase as its Postgres database and backend:
+
+1. [Create a new project on the Supabase dashboard](https://supabase.com/dashboard/projects).
+2. Go to the Supabase SQL Editor for your new project and execute the SQL statements in [`db/seed.sql`](db/seed.sql) to create the database schema, PowerSync replication role, and publication needed for PowerSync.
+
+**Important:** When connecting PowerSync to your Supabase database, you'll use the `powersync_role` credentials instead of the default Supabase connection string. This role has the necessary replication privileges and bypasses Row Level Security (RLS).
```
stevensJourney, line 71: We should probably add a note to update the password in `db/seed.sql` before running?

demos/nuxt-supabase-todolist/README.md
```diff
@@ -0,0 +1,95 @@
...
+│   └── confirm.vue           # Auth confirmation page
+├── components/
+│   └── AppHeader.vue         # Header component
+├── db/
+│   └── seed.sql              # Database setup SQL
+├── sync-rules.yaml           # PowerSync sync rules
+└── nuxt.config.ts            # Nuxt configuration
+```
+
+## Learn More
```
stevensJourney, line 138: It would be nice if this demo had a local Supabase config and quickstart for local development - which can be used to start the demo in only a few commands - like the YJS Demo has [here](https://github.com/powersync-ja/powersync-js/blob/5771c583ffe99ac572b7d28c6c18116e33c7e8b0/demos/yjs-react-supabase-text-collab/README.md#quick-start-local-development).

demos/nuxt-supabase-todolist/nuxt.config.ts
```diff
@@ -0,0 +1,76 @@
...
+      exclude: ['@journeyapps/wa-sqlite', '@powersync/web', '@powersync/common', '@powersync/vue', '@powersync/kysely-driver'],
+      include: [
+        '@powersync/web > js-logger',
+        '@supabase/postgrest-js',
+      ],
+    },
+
+    worker: {
+      format: 'es',
+      plugins: () => [wasm(), topLevelAwait()],
```
stevensJourney: I believe you should be able to remove the `topLevelAwait` plugin these days - due to the BSON lib no longer requiring it. It's only our list of todos for other demos.

demos/nuxt-supabase-todolist/package.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "dependencies": {
+    "@nuxt/ui": "^4.0.0",
+    "@nuxtjs/supabase": "2.0.1",
+    "@powersync/nuxt": "workspace:*"
+  },
+  "devDependencies": {
+    "@nuxt/devtools": "^1.0.6",
+    "@nuxt/eslint": "^1.9.0",
+    "@supabase/supabase-js": "2.75.0",
```
stevensJourney: Nit: This should probably be a `dependency`?

demos/nuxt-supabase-todolist/powersync/AppSchema.ts
```diff
@@ -0,0 +1,28 @@
...
+    user_id: column.text,
+  },
+  { indexes: { user: ['user_id'] } },
+)
+
+export const AppSchema = new Schema({
+  tasks,
+})
+
+export const AppSchemaWithDiagnostics = new Schema([
```
stevensJourney, line 20: It might make sense to remove the requirement for devs to create two variants of the schema. 

One could (hopefuly) encapsulate the diagnostics setup using an existing PowerSync instance by using `updateSchema`. e.g. if the Diagnostics package had a helper of the form

``` typescript
export async function configureDiagnostics(database: AbstractPowerSyncDatabase) {
  await database.updateSchema(new Schema([...database.schema.tables, ...diagnosticsSchema.tables]));
}
```

demos/nuxt-supabase-todolist/sync-rules.yaml
```diff
@@ -0,0 +1,9 @@
+config:
```
stevensJourney, line 2: Adding a schema for this file is always nice

```yaml
# yaml-language-server: $schema=https://unpkg.com/@powersync/service-sync-rules@latest/schema/sync_rules.json
```

package.json
```diff
@@ -49,5 +49,10 @@
     "rollup-plugin-dts": "^6.2.1",
     "typescript": "^5.7.2",
     "vitest": "^3.2.4"
+  },
+  "pnpm": {
+    "overrides": {
+      "cookie": "1.0.2"
```
stevensJourney: Can this override be applied on the demo level instead of the workspace level?

demos/nuxt-supabase-todolist/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "name": "powersync-nuxt-supabase-todolist",
+  "scripts": {
+    "dev": "nuxt dev",
+    "build": "nuxt build",
+    "start": "nuxt preview"
+  },
+  "dependencies": {
+    "@nuxt/ui": "^4.0.0",
+    "@nuxtjs/supabase": "2.0.1",
+    "@powersync/nuxt": "workspace:*"
```
stevensJourney: This should depend on `@powersync/web`? When running locally, it seems this package is not linked to the demo - which causes type errors for me.

packages/nuxt/README
```diff
@@ -0,0 +1,667 @@
+<!--
+Get your module up and running quickly.
+
+Find and replace all on all files (CMD+SHIFT+F):
```
stevensJourney: Should this still be here?

packages/nuxt/README
```diff
@@ -0,0 +1,667 @@
...
+    const plugin = createPowerSyncPlugin({ database: db })
+    nuxtApp.vueApp.use(plugin)
+  },
+})
+```
+
+At this point, you're all set to use the module composables. The module automatically exposes all `@powersync/vue` composables, so you can use them directly:
+
+- `usePowerSync()` - Access the PowerSync database instance
+- `usePowerSyncQuery()` - Query the database reactively
```
stevensJourney: `usePowerSyncQuery` is deprecated, rather link to `useQuery` here. Also the same for the items below this one.

packages/nuxt/README
```diff
@@ -0,0 +1,667 @@
...
+
+#### Nuxt Devtools
+
+The inspector is also available in the Nuxt Devtools as a tab, providing seamless integration with your development workflow.
+
+> [!WARNING]  
+> Multitab support is still not fully supported when diagnostics are enabled which causes the inspector to malfunction in the devtool. see [Known Issues section](#known-issues)
+
+
+## API Reference
```
stevensJourney: These kinds of API references are generated automatically for other packages by Docusaurus - see the `docs` folder. I'd recommend not having this here and rather adding this new package to those docs.

packages/nuxt/README
```diff
@@ -0,0 +1,667 @@
...
+    },
+  },
+})
+```
+
+## Development
+
+```bash
+# Install dependencies
+npm install
```
stevensJourney: This monorepo uses `pnpm` - so not sure why we'd recommend `npm` here for development?
This content will also be visible on our package's NPM.js page, which is why I'd rather recommend moving this section to its own file - to avoid it reflecting there.

packages/nuxt/package.json
```diff
@@ -0,0 +1,81 @@
...
+      "access": "public"
+    },
+    "description": "PowerSync Nuxt module",
+    "license": "Apache-2.0",
+    "type": "module",
+    "repository": {
+      "type": "git",
+      "url": "git+https://github.com/powersync-ja/powersync-js.git"
+    },
+    "author": "JOURNEYAPPS",
```
stevensJourney: I believe you can change this to `POWERSYNC`, that was recommended for other new packages.

packages/nuxt/package.json
```diff
@@ -0,0 +1,81 @@
...
+      "fuse.js": "^7.1.0",
+      "mitt": "^3.0.1",
+      "reka-ui": "^2.5.0",
+      "shiki": "^3.13.0",
+      "unocss": "^66.5.2",
+      "unstorage": "^1.17.1"
+    },
+    "peerDependencies": {
+      "@journeyapps/wa-sqlite": "^1.2.6",
+      "@powersync/kysely-driver": "workspace:*",
```
stevensJourney: I believe this dependency should be optional?

packages/nuxt/src/runtime/pages/__powersync-inspector.vue
```diff
@@ -0,0 +1,182 @@
+<template>
```
stevensJourney, line 1: I'm not sure why this is the case, but I've noticed when this page loads, an unstyled version of the layout is present for a noticeable period before the page renders correctly. 

image

stevensJourney: The page rendering also seems very flashy on my machine. E.g. if I have an upload error, the page flickers when updating the status.

https://github.com/user-attachments/assets/e00c49a5-e188-4cdd-9136-17bf2b7bb21e

packages/nuxt/src/runtime/composables/usePowerSyncInspectorDiagnostics.ts
```diff
@@ -0,0 +1,337 @@
...
+  const isUploading = ref(syncStatus.value?.dataFlowStatus.uploading || false)
+  const lastSyncedAt = ref<Date | null>(syncStatus.value?.lastSyncedAt || null)
+  const uploadError = ref(syncStatus.value?.dataFlowStatus.uploadError || null)
+  const downloadError = ref(
+    syncStatus.value?.dataFlowStatus.downloadError || null,
+  )
+  const downloadProgressDetails = ref(
+    syncStatus.value?.dataFlowStatus.downloadProgress || null,
+  )
+  const bucketRows = ref<null | any[]>(null)
```
stevensJourney: Please type these query results accordingly.

packages/nuxt/src/runtime/composables/usePowerSyncInspectorDiagnostics.ts
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    // Clean up listener on unmount
+    onUnmounted(() => {
+      unregisterListener()
+    })
+  })
+
+  // exposed state
+  return {
```
stevensJourney, line 396: Please create and export a type for this return type.

packages/nuxt/src/runtime/utils/AppSchema.ts
```diff
@@ -0,0 +1,27 @@
+import { column, Schema, Table } from '@powersync/web'
+
+export const local_bucket_data = new Table(
```
stevensJourney, line 6: Nit: We typically name these kinds of constants in upper case.

packages/nuxt/src/runtime/utils/DynamicSchemaManager.ts
```diff
@@ -0,0 +1,116 @@
+import type { DBAdapter, SyncDataBatch } from '@powersync/web'
+import { Column, ColumnType, OpTypeEnum, Schema, Table } from '@powersync/web'
+import { DiagnosticsAppSchema as AppSchema } from './AppSchema'
+import { JsSchemaGenerator } from './JsSchemaGenerator'
+
+/**
+ * Record fields from downloaded data, then build a schema from it.
+ */
+export class DynamicSchemaManager {
```
stevensJourney, line 9: This looks like a copy from `tools/diagnostics-app`? It would be nice if we had this functionality in some shared location.

packages/nuxt/src/runtime/utils/NuxtPowerSyncDatabase.ts
```diff
@@ -0,0 +1,145 @@
...
+      const currentSchemaManager = getCurrentSchemaManager()
+
+      // override settings to disable multitab as we can't use it right now
+      // options.flags = {
+      //   ...options.flags,
+      //   enableMultiTabs: true, // to support multitabe we need to write our won worker implementation
+      //   broadcastLogs: true, // need to be enabled for multitab support
+      // }
+
+      // @ts-expect-error - type error because we are forcing the vfs to be the OPFSCoopSyncVFS
```
stevensJourney: This seems fishy. Why do we need to make this force assertion?

packages/nuxt/src/runtime/utils/NuxtPowerSyncDatabase.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (useDiagnostics) {
+      const { logger } = useDiagnosticsLogger()
+      const { getCurrentSchemaManager } = usePowerSyncInspector()
+      // Create schema manager before calling super
+      const currentSchemaManager = getCurrentSchemaManager()
+
+      // override settings to disable multitab as we can't use it right now
+      // options.flags = {
+      //   ...options.flags,
+      //   enableMultiTabs: true, // to support multitabe we need to write our won worker implementation
```
stevensJourney: If you're overriding the PowerSyncDatabase, it might be possible (with some minor modifications) to actually implement your own web worker implementation, using the current primitives, which contains the required sync hooks.

packages/nuxt/src/runtime/utils/RustClientInterceptor.ts
```diff
@@ -0,0 +1,141 @@
...
+  isStreamingSyncCheckpointPartiallyComplete,
+  isStreamingSyncData,
+  PowerSyncControlCommand,
+  SqliteBucketStorage,
+  SyncDataBucket,
+} from '@powersync/web'
+import type { DynamicSchemaManager } from './DynamicSchemaManager'
+import type { Ref } from 'vue'
+/**
+ * Tracks per-byte and per-operation progress for the Rust client.
```
stevensJourney, line 24: Just sharing this again. I'd really prefer we did not duplicate these implementations.

stevensJourney (commented):
Just sharing some initial comments for some items which stood out from an initial review.

demos/nuxt-supabase-todolist/README.md
```diff
@@ -0,0 +1,95 @@
...
+2. Ensure "Email" provider is enabled
+3. You can disable email confirmation for development by going to "Authentication" -> "Email Auth" and disabling "Confirm email"
+
+You'll need to create a user account when you first access the application.
+
+### 4. Set up PowerSync
+
+You can use either PowerSync Cloud or self-host PowerSync:
+
+- **PowerSync Cloud**: [Create a new project on the PowerSync dashboard](https://powersync.journeyapps.com/) and connect it to your Supabase database using the `powersync_role` credentials created in step 2.
```
LucDeCaf: Nit: should probably link to https://dashboard.powersync.com/

demos/nuxt-supabase-todolist/nuxt.config.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  // enable hot reloading when we make changes to our module
+  watch: ['../src/*', './**/*'],
+
+  compatibilityDate: '2024-07-05',
+
+  vite: {
+    plugins: [topLevelAwait()],
+    optimizeDeps: {
+      exclude: ['@journeyapps/wa-sqlite', '@powersync/web', '@powersync/common', '@powersync/vue', '@powersync/kysely-driver'],
```
LucDeCaf: Do we need to exclude all `@powersync/*` packages from optimisation? AFAIK, the only ones that need to be excluded are `@powersync/web` and `@journeyapps/wa-sqlite`, since those packages include web workers and wasm respectively - the others should be fine to have optimised.

demos/nuxt-supabase-todolist/pages/login.vue
```diff
@@ -0,0 +1,118 @@
...
+            sign === "up"
+              ? "Already have an account?"
+              : "Don't have an account?"
+          }}
+          <UButton
+            variant="link"
+            class="p-0"
+            @click="sign = sign === 'up' ? 'in' : 'up'"
+          >
+            {{ sign === "in" ? "Sign up" : "Sign in" }}
```
LucDeCaf: Looks like this condition might be reversed.

packages/nuxt/src/devtools.ts
```diff
@@ -0,0 +1,22 @@
...
+  const DEVTOOLS_UI_ROUTE = `http://localhost:${port}/__powersync-inspector`
+
+  nuxt.hook('devtools:customTabs', (tabs) => {
+    tabs.push({
+      // unique identifier
+      name: 'powersync-inspector',
+      // title to display in the tab
+      title: 'Powersync Inspector',
+      // any icon from Iconify, or a URL to an image
+      icon: 'https://cdn.prod.website-files.com/67eea61902e19994e7054ea0/67f910109a12edc930f8ffb6_powersync-icon.svg',
```
LucDeCaf: This should maybe be a file in 'assets' instead of a link - not too fussed, though

packages/nuxt/src/runtime/composables/usePowerSyncInspectorDiagnostics.ts
```diff
@@ -0,0 +1,337 @@
...
+import type {
+  PowerSyncBackendConnector,
+  PowerSyncConnectionOptions,
+  UploadQueueStats,
+} from '@powersync/web'
+import { ref, computed, readonly, onMounted, onUnmounted } from 'vue'
+import { computedAsync } from '@vueuse/core'
+import { usePowerSyncInspector } from './usePowerSyncInspector'
+
+type JSONValue
```
LucDeCaf: The same JSON types are declared in `src/module.ts`. It might make sense to move these to a `types.ts` file or similar.

scripts/reset-node-modules.ts
LucDeCaf, line 1: Not entirely sure what the benefit of a dedicated script is over using `npx npkill` or even `rm -rf demos/**/node_modules`

package.json
khawarizmus: This is unfortunately not possible currently. I tried various ways to force the override locally, but it only worked at the root level. This issue stems from the dependency being a transient dependency and used by different projects within the monorepo.

packages/nuxt/README
khawarizmus: This was from the official template. It can safely be removed at this stage

packages/nuxt/README
khawarizmus: Added the module to Docosurus

packages/nuxt/README
khawarizmus: When you say it's own file, what kind of file exactly? I have updated this to use `pnpm` and removed some commands that do not apply anymore after migrating the project to the mnorepo

packages/nuxt/package.json
khawarizmus: I can change it to `POWERSYNC`, but a quick search shows that 13 (including this module) projcts are using `JOURNEYAPPS`

packages/nuxt/src/runtime/composables/usePowerSyncInspectorDiagnostics.ts
khawarizmus: It turns out the type wasn't used. But if it needed in the future, we should expose it from `@powersync/common` package instead.

scripts/reset-node-modules.ts
khawarizmus: I quickly vibe-coded this to reset the entire monorepo, not only the demos, which allowed me to iterate faster when I was having dependency and build issues. It can be improved to delete the `pnpm-lock` file and other unnecessary folders like `.nuxt` inside the nuxt demo folder.

I don't mind removing it completely either, but I found it useful tbh, that's why I pushed it.

packages/nuxt/src/runtime/composables/usePowerSyncKysely.ts
LucDeCaf, line 1: I'm not sold on bundling the Kysely integration with the Nuxt package, if users want to use another ORM they would have to pay the price of both ORMs' bundles.

demos/nuxt-supabase-todolist/app.vue
```diff
@@ -0,0 +1,58 @@
+<template>
```
Chriztiaan, line 1: I'd suggest that we keep a consistent orders for templates/script blocks across all our components in this demo.

demos/nuxt-supabase-todolist/pages/confirm.vue
```diff
@@ -0,0 +1,17 @@
...
+const user = useSupabaseUser()
+
+watch(
+  user,
+  () => {
+    if (user.value) {
+      return navigateTo('/')
+    }
+  },
+  { immediate: true },
```
Chriztiaan: A few of these watches that happen to be immediate:true, could be swapped by watchEffect() - https://vuejs.org/guide/essentials/watchers#watcheffect

demos/nuxt-supabase-todolist/pages/index.vue
```diff
@@ -0,0 +1,153 @@
+<script setup lang="ts">
+import type { Database, TaskRecord } from '~/powersync/AppSchema'
+
+const client = useSupabaseClient()
+const user = asyncComputed(async () => await client.auth.getUser().then(res => res.data.user))
```
Chriztiaan: This is not a built-in helper, it's from VueUse, which doesn't seem to be a dependency of this demo - I do see it as one for the nuxt package though. Are we autoinstalling it for projects? Or is it just leaking into this project from the package?

I would consider dropping it for the demo to make the demo easier to adopt.

package.json
Chriztiaan: Now that the demo isolate work is merged this should be a solvable. Each demo has its own workspacefile https://github.com/powersync-ja/powersync-js/blob/main/demos/react-supabase-todolist/pnpm-workspace.yaml.

demos/nuxt-supabase-todolist/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "name": "powersync-nuxt-supabase-todolist",
+  "scripts": {
+    "dev": "nuxt dev",
+    "build": "nuxt build",
+    "start": "nuxt preview"
+  },
+  "dependencies": {
+    "@nuxt/ui": "^4.0.0",
+    "@nuxtjs/supabase": "2.0.1",
+    "@powersync/nuxt": "workspace:*",
```
Chriztiaan: Our demos no longer depend on workspace dependencies, after this PR is approved we may need to split the work into two PRs to a release out that this demo can depend on.

demos/nuxt-supabase-todolist/powersync/AppSchema.ts
khawarizmus: As discussed. I will try to load this directly in the constructor instead, and I update the docs.

package.json
khawarizmus: Will try to make it work that way. When I started, the isolation wasn't there yet.

packages/nuxt/package.json
khawarizmus: Yes, will try and make it optional as per our discussion

packages/nuxt/src/runtime/pages/__powersync-inspector.vue
khawarizmus: Will investigate the flickering as for the style loading i will see if we can show a loading state before the unocss loads

packages/nuxt/src/runtime/utils/DynamicSchemaManager.ts
khawarizmus: As per our discussion, it's not a 100% copy past since I had to make some parts reactive, but there is a proposale that would make all this code disappear https://github.com/powersync-ja/powersync-sqlite-core/issues/155

packages/nuxt/src/runtime/utils/NuxtPowerSyncDatabase.ts
khawarizmus: This will be removed.

packages/nuxt/src/runtime/utils/RustClientInterceptor.ts
khawarizmus: Addressed above

packages/nuxt/src/runtime/composables/usePowerSyncKysely.ts
khawarizmus: I agree with this. I will see how to optionally include the dependency and composable via a config flag.

demos/nuxt-supabase-todolist/pages/index.vue
khawarizmus: Yes, the module installs VueUse automatically as we are using it in some components. I wouldn't worry about including it since it's a very common module.

Chriztiaan: This may be a slippery slope, generally we want our packages to be as slim as possible. If some of the other/bigger nuxt modules auto apply this dependency too maybe it's fine. But I wouldn't want us to be the only package that sneaks in a 1mb dependency that they didn't explicitly opt in for.

scripts/reset-node-modules.ts
Chriztiaan: If we do plan on keeping this script, can we consider adding to the package.json https://github.com/powersync-ja/powersync-js/blob/main/package.json#L20-L21 ?

demos/nuxt-supabase-todolist/pages/index.vue
Chriztiaan: Can you confirm which version is applied if you install a differing VueUse version in the demo too? Conflicts, or is one prioritised over the other?

demos/nuxt-supabase-todolist/README.md
khawarizmus: I have added all necessary files and instructions for this

demos/nuxt-supabase-todolist/powersync/AppSchema.ts
khawarizmus: This has been addressed

demos/nuxt-supabase-todolist/package.json
khawarizmus: Yes, pnpm doesn't install peer dependencies by default, which is different from how npm does it. So it had to be explicitly added.

packages/nuxt/package.json
khawarizmus: This is changed but my above comment still stands, currently all packages don't use `POWERSYNC` as an author

packages/nuxt/package.json
khawarizmus: Made the dependency and loading Kysely composable optional

packages/nuxt/src/runtime/composables/usePowerSyncInspectorDiagnostics.ts
khawarizmus: It seems to that the type is automatically inferred. Is there a reason for us to add this type explicitly?

packages/nuxt/src/runtime/utils/NuxtPowerSyncDatabase.ts
khawarizmus: I have used the `SharedWebStreamingSyncImplementation`, which made the inspector work across tabs now. but because it is the only way to have it work with the official Nuxt devtools, I opted into overriding the settings to force multitab.

Another way would be to display an error and say that they have to use multitab but can still access the inspector via direct URL. This would require detecting that we are inside the devtool (which I think is possible)

scripts/reset-node-modules.ts
khawarizmus: I have renamed the script to `reset-repo` and added it to the `package`json` file. 

kept the logic and made it delete `pnpm-lock` files. I was thinking of making it also delete `dist` and `lib` folders as well as `tsconfig.tsbuildinfo` files. This would make it easier to reset -> install -> build

Let me know what you guys think

packages/nuxt/src/runtime/composables/usePowerSyncKysely.ts
khawarizmus: I have made it so that the user has to explicitly opt in to use Kysely, and the composable won't be imported unless they do. The dependency is also optional, and if they opt in and don't have it installed, the module would error.

packages/nuxt/src/runtime/pages/__powersync-inspector.vue
khawarizmus: I have added a logic that detects if the Unocss is interpreted by the browser properly, which I use to toggle off a loading screen.

The loading screen uses a replica of the [loading component](https://github.com/nuxt/devtools/blob/main/packages/devtools-ui-kit/src/components/NLoading.vue) from the official Nuxt devtools ui kit, implemented in pure CSS. This makes it work independently of Unocss and still looks native to Nuxt devtools.

khawarizmus: The flickering should be fixed. it was a change of size in the icons when the status changes.

demos/nuxt-supabase-todolist/app.vue
khawarizmus: Good point. I have made things more consistent now.

demos/nuxt-supabase-todolist/pages/confirm.vue
khawarizmus: This was addressed

packages/nuxt/src/runtime/composables/usePowerSyncInspectorDiagnostics.ts
LucDeCaf: This is mostly so users who want to use the type have a convenient type name they can depend on, i.e. `let state: PowerSyncInspectorState;` instead of `let state: any;` or `let state: { db: ..., syncStatus: ... };`. FWIW, I believe the Typescript LSP has a code action to generate a named type from an object's inferred type (not sure how to run code actions on VSCode).

scripts/reset-node-modules.ts
LucDeCaf: I like that idea. One thing to be wary of if we delete `lib` files is that some projects use/used `lib` as a source folder instead of a build folder. (IIRC I replaced one of those with a `library` folder, not sure if there's still more but worth considering.)

It might also be worth letting users choose which repos are cleaned. There's some similar looking logic in the demo management scripts if you feel like implementing this.

packages/nuxt/src/runtime/utils/AppSchema.ts
khawarizmus: I exported the names of the tables as a cost.

packages/nuxt/src/runtime/composables/usePowerSyncInspectorDiagnostics.ts
khawarizmus: This has been addressed

packages/nuxt/src/runtime/composables/usePowerSyncInspectorDiagnostics.ts
khawarizmus: Makes sense. I have made the appropriate types for it.

demos/nuxt-supabase-todolist/package.json
khawarizmus: I have made a dev release and isolated the demo. now both the nuxt package version and the demo package json point to the dev release.

demos/nuxt-supabase-todolist/pages/index.vue
khawarizmus: From my tests it works fine. It also prioritizes the local version over what's in the nuxt package.

packages/nuxt/src/runtime/composables/useDiagnosticsLogger.ts
```diff
@@ -0,0 +1,70 @@
...
+import localStorageDriver from 'unstorage/drivers/session-storage'
+import mitt from 'mitt'
+
+const emitter = mitt()
+
+const logsStorage = createStorage({
+  driver: localStorageDriver({ base: 'powersync:' }),
+})
+
+const consola = createConsola({
```
Chriztiaan: I think we may need to drop consola usage in this package in favour of our logger helper.
A future case could be made to move from `js-logger` to consola - but I think it's noise for this PR.

demos/nuxt-supabase-todolist/nuxt.config.ts
```diff
@@ -0,0 +1,75 @@
...
+  // enable hot reloading when we make changes to our module
+  watch: ['../src/*', './**/*'],
+
+  compatibilityDate: '2024-07-05',
+
+  vite: {
+    optimizeDeps: {
+      exclude: ['@journeyapps/wa-sqlite', '@powersync/web'],
+      include: [
+        '@powersync/web > js-logger',
```
Chriztiaan: This may no longer be needed. See other updated demos config.

packages/nuxt/README.md
```diff
@@ -0,0 +1,390 @@
...
+This module re-exports all `@powersync/vue` composables. With **npm** (v7+), installing `@powersync/nuxt` is enough; with **pnpm**, install peer dependencies explicitly.
+
+```bash
+# Using npm
+npm install @powersync/nuxt
+npm install --save-dev vite-plugin-top-level-await vite-plugin-wasm
+
+# Using pnpm (peer deps are not auto-installed)
+pnpm add @powersync/nuxt @powersync/vue @powersync/web
+pnpm add -D vite-plugin-top-level-await vite-plugin-wasm
```
Chriztiaan, line 35: Top level await may no longer be needed. See other updated demos config.

packages/nuxt/src/runtime/components/ConfigInspectorTab.vue
```diff
@@ -0,0 +1,175 @@
...
+          class="text-sm schema-code-block h-full overflow-auto"
+          v-html="schemaHtml"
+        />
+      </NSectionBlock>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import type {
```
Chriztiaan: Just noticed, our other packages have semi colon line endings. Which isn't applied in this package?

packages/nuxt/vitest.config.ts
```diff
@@ -0,0 +1,49 @@
...
+import { resolve } from 'path';
+
+import topLevelAwait from 'vite-plugin-top-level-await';
+import wasm from 'vite-plugin-wasm';
+
+const config: UserConfigExport = {
+  // This is only needed for local tests to resolve the package name correctly
+  worker: {
+    format: 'es',
+    plugins: () => [wasm(), topLevelAwait()]
```
Chriztiaan: No longer need top level await, check other demos.

packages/nuxt/src/runtime/composables/useDiagnosticsLogger.ts
khawarizmus: I have dropped `consola` and removed it from being a dependency. But I agree that using `consola` over `js-logger` would bring a nicer DX and probably more portability.

demos/nuxt-supabase-todolist/nuxt.config.ts
khawarizmus: This is addressed

packages/nuxt/README.md
khawarizmus: This is addressed

packages/nuxt/src/runtime/components/ConfigInspectorTab.vue
khawarizmus: I had to set Prettier as my default formatter and instruct it to format Vue files, but this is addressed

packages/nuxt/vitest.config.ts
khawarizmus: This is addressed

.changeset/gentle-poems-live.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/nuxt': minor
```
Chriztiaan: As we want this to be released under 0.0.1 for the first version, this will likely need to be a patch change instead.

packages/nuxt/src/runtime/components/SyncStatusTab.vue
```diff
@@ -0,0 +1,186 @@
...
+            </span>
+          </div>
+        </div>
+      </NSectionBlock>
+
+      <span border="b" border-color="gray-100" />
+
+      <NSectionBlock icon="carbon:data-share" text="Operations">
+        <NTip v-if="!isDiagnosticSchemaSetup" n="red6 dark:red5" icon="carbon:warning-alt">
+          Make sure to extend your schema with the diagnostics schema using the `diagnosticsSchema` from the
```
Chriztiaan: This is duplicated, weirdly. Is there room for reuse in this file?

packages/nuxt/src/runtime/utils/RecordingStorageAdapter.ts
```diff
@@ -0,0 +1,75 @@
...
+export class RecordingStorageAdapter extends SqliteBucketStorage {
+  private rdb: DBAdapter;
+  private schemaManager: DynamicSchemaManager;
+
+  public tables: Record<string, Record<string, ColumnType>> = {};
+
+  constructor(db: Ref<PowerSyncDatabase>, schemaManager: Ref<DynamicSchemaManager>) {
+    super(
+      db.value.database,
+
```
Chriztiaan: Rogue newline

package.json
```diff
@@ -50,6 +51,8 @@
     "prettier": "^3.2.5",
     "prettier-plugin-embed": "^0.4.15",
     "prettier-plugin-sql": "^0.18.1",
+	  "rollup": "4.14.3",
```
Chriztiaan: Weird space?

packages/nuxt/README.md
Chriztiaan, line 1: README => README.md

demos/nuxt-supabase-todolist/powersync/AppSchema.ts
```diff
@@ -0,0 +1,25 @@
+import { column, Schema, Table } from '@powersync/web'
+
+export const TASKS_TABLE = 'tasks'
```
Chriztiaan, line 3: Is there a strong case to deviate from the existing lists/todos and customers app schemas we use in our other demos?

packages/nuxt/src/runtime/components/SyncStatusTab.vue
khawarizmus: There is a bunch of v-ifs, which makes it difficult to loop over; however i have introduced some v-fors to reduce the repetition.

package.json
khawarizmus: I don't see this in my `package.json`
image

khawarizmus:
> I tried testing the nuxt demo on my side, but got this. I had the same issue against a new dev release.

@Chriztiaan I am not able to reproduce this on my side

package.json
Chriztiaan: I deleted all the whitespace on my side, and formatted the file. The diff is now fixed.

Chriztiaan:
> > I tried testing the nuxt demo on my side, but got this. I had the same issue against a new dev release.
> 
> @Chriztiaan I am not able to reproduce this on my side

Resolved on a fresh clone of the repo.

Chriztiaan:
Screenshot 2026-02-17 at 08 57 43
Screenshot 2026-02-17 at 08 57 38

I think this is wrong. It should be:
> Don't have an account? Sign Up.
> Already have an account? Sign In.

Chriztiaan (approved):
Happy with the changes, tested locally and working. Thanks!

scripts/reset-node-modules.ts
Chriztiaan: Future iteration @khawarizmus

Issue #787, opened 2025-12-03
Triggers + OPFS can miss updates


stevensJourney:
# Overview

The triggers and High Performance Diffs functionality can produce unexpected results if used with multiple tabs and OPFS.

Note: The IndexedDB VFS implementation does not suffer from this since a single SQLite connection is shared between all tabs. Each tab has their own SQLite trigger and destination table for ops. Every change to a table will trigger all triggers. 

Related https://github.com/TanStack/db/issues/948

stevensJourney:
This has been fixed in [@powersync/web@1.32.0](https://github.com/powersync-ja/powersync-js/releases/tag/%40powersync%2Fweb%401.32.0)

PR #804, opened 2025-12-30
OPFS Multiple Tab Trigger Invocation

stevensJourney:
# Overview

Fixes:
- https://github.com/powersync-ja/powersync-js/issues/787
- https://github.com/TanStack/db/issues/948

Our SQLite trigger helpers currently allow for tracking INSERT, UPDATE, and DELETE operations on a PowerSync view. These operations are stored in a SQLite table for each managed consumer to process.

Our managed triggers currently only support **temporary** SQLite triggers, where diff operations are stored in temporary SQLite tables. This approach avoids extra storage overhead for diff records and simplifies cleanup. The temporary nature means the triggers and tables are scoped to the SQLite connection and are disposed automatically once the connection is closed.

## The Problem

For OPFS (in Web), we create a separate SQLite connection per tab. Triggers are also scoped to each tab, so each tab creates its own managed trigger (hence its own SQLite trigger and destination table). Having multiple connections across multiple tabs causes issues:

**A mutation performed by one connection does not trigger a temporary trigger defined for another connection.**

This causes our trigger consumers to miss updates from other tabs.

## The Solution

**Persisted SQLite triggers** are stored in the database. A mutation from any connection will trigger all persisted triggers, ensuring that mutations from other tabs are written to the destination table of other tabs' managed triggers.

This PR adds the ability to optionally use **persisted SQLite tables and triggers** (enabled by default for OPFS connections). While these resources do use persistence, they are still temporary in nature. The PowerSync SDK will dispose of these resources automatically when they are detected as no longer in use.

# Implementation

Using persisted triggers and destination tables solves the multiple SQLite connection problem, but introduces a new challenge: how do we dispose of the persisted tables and triggers when they are no longer needed?

## The Claim Mechanism

Disposing stale items for closed tabs (and SQLite connections) is not straightforward. When a tab is closed, there is usually no reliable method for ensuring cleanup has completed successfully.

To address this, the implementation introduces a **"claim" mechanism** through a new `TriggerClaimManager` interface. This interface provides two core capabilities: obtaining a claim on a resource (returning a release callback), and checking whether a resource is currently claimed.

A heartbeat or time-to-live mechanism was also considered, but this approach seemed vulnerable to slow or frozen tab issues. A tab that's temporarily unresponsive could have its resources incorrectly cleaned up. The hold method avoids this problem since a hold remains valid regardless of how responsive the owning tab is.

The resources and claims to resources are freed if the trigger is disposed using the callback returned from the method which created it.

```typescript
const dispose = await database.triggers.trackTableDiff({...})

// This will delete the tables
await dispose()
```

For cases where the triggers are not explicitly disposed, a cleanup process is implemented. This process runs whenever a new `PowerSyncDatabase` is created and at a 2 minute interval.

A unique UUID is associated with each managed trigger. This ID and destination SQLite table are encoded into the names of the SQLite triggers created. The naming convention for triggers is as follows:

`__ps_temp_trigger_${operation}__${destination_table}__${trigger_id}`

The cleanup process:
- queries the `sqlite_master` table for all triggers which match the `__ps_temp_trigger` naming convention. The destination table and trigger id are extracted
- Checks the `TriggerHoldManager` if any active claims are present for the trigger id
- Deletes the triggers and tables if no claim is present.

Alternative approaches were considered for the tracking of the managed trigger Ids and tables.

- Storing metadata in `ps_kv: Cant be used since the keys in `ps_kv` are cleared in `disconnectAndClear`, which will discard the state.
- Adding comments to the trigger sql, which should be present in sqlite_master: Can't be used since the comments are removed when executing (at least for WA-SQLite)

## Platform-Specific Claim Managers

The claim mechanism needs to work differently depending on the platform.

For shared-memory SDKs like Node and React Native, a `MemoryTriggerClaimManager` uses a global in-memory store to track claims. Since these environments typically share memory within a single process, this straightforward approach works well.

For web environments, a `NavigatorTriggerClaimManager` leverages the [Navigator Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) to manage claims across browser tabs. This allows us to determine if any tab is still holding on to a trigger, even when those tabs can't directly communicate with each other.

The web `PowerSyncDatabase` enables persistence by default for OPFS VFS modes, while keeping temporary triggers for `IDBBatchAtomicVFS` where multiple connections aren't an issue.

### DIFF
--- a/packages/common/src/client/triggers/MemoryTriggerClaimManager.ts
+++ b/packages/common/src/client/triggers/MemoryTriggerClaimManager.ts
@@ -1,6 +1,5 @@
 import { TriggerClaimManager } from './TriggerManager.js';
 
-// Uses a global store to share the state between potentially multiple instances
 const CLAIM_STORE = new Map<string, () => Promise<void>>();
 
 /**


packages/common/src/client/triggers/TriggerManager.ts
```diff
@@ -449,3 +458,30 @@ export interface TriggerManager {
    */
   trackTableDiff(options: TrackDiffOptions): Promise<TriggerRemoveCallback>;
 }
+
+/**
+ * @experimental
+ * @internal
+ * An interface which exposes which persisted managed SQLite triggers and destination SQLite tables
+ * are actively in use. Resource which are not reported as claimed by this interface will be disposed.
+ */
```
simolus3, line 475: The PR description gives a very nice overview of what this interface is supposed to accomplish, I think copying parts of the "The Claim Mechanism" section into this documentation comment might make the purpose of this easier to understand for someone stumbling upon this later. Or at least for me this comment is a bit abstract and doesn't make it obvious why this is necessary.

packages/common/src/client/triggers/MemoryTriggerClaimManager.ts
```diff
@@ -0,0 +1,26 @@
+import { TriggerClaimManager } from './TriggerManager.js';
+
+/**
+ * @internal
+ * @experimental
+ */
+export class MemoryTriggerClaimManager implements TriggerClaimManager {
+  // Uses a global store to share the state between potentially multiple instances
+  private static CLAIM_STORE = new Map<string, () => Promise<void>>();
```
simolus3: nit: Having a class whose only internal state is a static field feels a bit odd to me, perhaps a global constant implementing `TriggerClaimManager` as a JS object would make that clearer.

packages/common/src/client/triggers/TriggerManagerImpl.ts
```diff
@@ -110,29 +273,30 @@ export class TriggerManagerImpl implements TriggerManager {
...
           timestamp TEXT,
           value TEXT,
           previous_value TEXT
-        );
+        )
       `);
 
       if (operations.includes(DiffTriggerOperation.INSERT)) {
-        const insertTriggerId = `ps_temp_trigger_insert_${id}`;
+        const insertTriggerId = this.generateTriggerName(DiffTriggerOperation.INSERT, destination, id);
```
simolus3, line 295: Wouldn't `cleanupResources` also run for temporary triggers since they have same naming convention? These temporary triggers aren't registered in the claims tracker, so couldn't that lead to them being dropped prematurely?

I guess we're protected from this since we only select from `sqlite_master` which only includes tables from the `main` schema. But it would be good to mention this in a comment somewhere.

packages/common/src/client/triggers/TriggerManager.ts
```diff
@@ -201,6 +204,12 @@ interface BaseCreateDiffTriggerOptions {
    * Hooks which allow execution during the trigger creation process.
    */
   hooks?: TriggerCreationHooks;
+
+  /**
+   * Use storage-backed (non-TEMP) tables and triggers that persist across sessions.
+   * These resources are still automatically disposed when no longer claimed.
+   */
+  useStorage?: boolean;
```
simolus3, line 212: Given that we have mechanisms to clean up those triggers reliably (albeit after a small delay), I wonder if it could make sense to unconditionally use stored triggers and to drop support for in-memory triggers?

The only benefit of in-memory triggers is potentially better performance. But not having this as an option would allow us to remove additional complexity in opening logic added to the web database. So if the performance benefit is not that significant, reducing implementation complexity might be a good tradeoff.

packages/common/src/client/triggers/TriggerManagerImpl.ts
```diff
@@ -48,14 +108,115 @@ export class TriggerManagerImpl implements TriggerManager {
...
+    // we use the database here since cleanupResources is called during the PowerSyncDatabase initialization
+    await this.db.database.writeLock(async (ctx) => {
+      // Query sqlite_master directly to find all persisted triggers and extract destination/id
+      // Trigger naming convention: __ps_temp_trigger_<operation>__<destination>__<id>
+      // - Compute start index after the second '__' (after operation) as a CTE for clarity
+      //   _start_index = instr(substr(name, 3), '__') + 4
+      //   (add 2 to account for removed leading '__', plus 2 to skip the '__' before destination)
+      // - Destination length excludes the trailing '__' + 36-char UUID: length(name) - _start_index - 37
+      // - UUID is always last 36 chars
+      const trackedItems = await ctx.getAll<TrackedTableRecord>(/* sql */ `
```
simolus3: nit: Just doing `SELECT * FROM sqlite_schema WHERE type = 'trigger' AND name LIKE '__ps_temp_trigger_%'` and then extracting the table name in JS would probably be easier to understand.

packages/common/src/client/triggers/TriggerManagerImpl.ts
```diff
@@ -48,14 +108,115 @@ export class TriggerManagerImpl implements TriggerManager {
...
+        FROM
+          trigger_names
+      `);
+
+      if (trackedItems.length == 0) {
+        // There is nothing to cleanup
+        return;
+      }
+
+      for (const trackedItem of trackedItems) {
```
simolus3: Do we need this check? There's nothing after the loop we might have to skip if there aren't any items.

```suggestion
      for (const trackedItem of trackedItems) {
```

packages/common/src/client/triggers/TriggerManagerImpl.ts
```diff
@@ -48,14 +108,115 @@ export class TriggerManagerImpl implements TriggerManager {
...
+        this.db.logger.debug(`Clearing resources for trigger ${trackedItem.id} with table ${trackedItem.table}`);
+
+        // We need to delete the triggers and table
+        const triggerNames = Object.values(DiffTriggerOperation).map((operation) =>
+          this.generateTriggerName(operation, trackedItem.table, trackedItem.id)
+        );
+        for (const triggerName of triggerNames) {
+          // The trigger might not actually exist, we don't track each trigger name and we test all permutations
+          await ctx.execute(`DROP TRIGGER IF EXISTS ${triggerName}`);
+        }
```
simolus3: The query iterates over all triggers, finds the unique affected table names and then this method again expands that to all possible trigger names to delete? Would it be easier to remove the `DISTINCT`, add the full trigger name as a select column and then just delete that? That means we'd be running `DROP TABLE IF EXISTS` multiple times for the same table if there are multiple triggers but that should be harmless.

packages/web/src/db/NavigatorTriggerClaimManager.ts
```diff
@@ -0,0 +1,19 @@
+import { TriggerClaimManager } from '@powersync/common';
+import { getNavigatorLocks } from '../shared/navigator';
+
+export class NavigatorTriggerClaimManager implements TriggerClaimManager {
```
simolus3: nit: Since this also has no internal state it could also be an `export const navigatorTriggerClaimManager: TriggerClaimManager = {...}`.

simolus3 (commented):
I'm happy with this approach. My main comment is that I'm wondering whether we should unconditionally use stored triggers to simplify our internal logic.

packages/common/src/client/triggers/TriggerManagerImpl.ts
stevensJourney: > I guess we're protected from this since we only select from sqlite_master which only includes tables from the main schema. But it would be good to mention this in a comment somewhere.

This is correct, those triggers won't be returned from `sqlite_master`. I will add more comments for this.

packages/common/src/client/triggers/TriggerManagerImpl.ts
stevensJourney: That's a fair point. I did consider doing this. A second opinion helps here. I'll update the implementation.

packages/common/src/client/triggers/TriggerManager.ts
stevensJourney: It's a fair point to mention this. 
In my opinion, the added complexity of configuring defaults is worth the potential improved performance.

There probably are a few scenarios where we want to avoid increasing the storage size. Two which come to mind are:
- The `SQL.js` adapter which exports the entire DB to be persisted after each `writeLock`.
- The IndexedDB VFS which has historically shown that performance suffers when the database size increases.

We could potentially perform the inverse, where storage is used by default and disabled for those cases - but then we still need to be able to configure the defaults and lose some of those potential performance benefits.

packages/common/src/client/triggers/TriggerManager.ts
stevensJourney: I added more description to the comment

packages/common/src/client/triggers/MemoryTriggerClaimManager.ts
stevensJourney: 👍 That's fair. I moved this and the navigator implementation to consts.

packages/common/src/client/triggers/TriggerManagerImpl.ts
stevensJourney: I added a note about this in the `cleanupResources` method

packages/common/src/client/triggers/TriggerManagerImpl.ts
stevensJourney: I've added the trigger names to the TrackedTableRecord. That way we can delete the actual triggers and tables once.

packages/common/src/client/triggers/TriggerManagerImpl.ts
stevensJourney: 👍  The JS based implementation does look cleaner.

packages/common/src/client/triggers/MemoryTriggerClaimManager.ts
```diff
@@ -0,0 +1,26 @@
+import { TriggerClaimManager } from './TriggerManager.js';
+
+// Uses a global store to share the state between potentially multiple instances
```
simolus3: This comment is outdated now

```suggestion
```

simolus3 (dismissed):
Looks good to me 👍

PR #806, opened 2026-01-03
Avoid binding this when disposing table change listener

### LINKED ISSUE #805: DataCloneError in shared‑worker path (function sent over MessagePort)
### Summary
We’re seeing a `DataCloneError` in production when calling `disconnectAndClear()` then `close()` on a PowerSync DB. The error stack points to Comlink `postMessage` inside the PowerSync bundle. The demangled function in the error matches `SqliteBucketStorage.tablesUpdated` (a listener registered via `db.registerListener`). It looks like a listener object containing a function is being sent across a MessagePort in the shared‑worker/multi‑tab path, which is not structured‑cloneable.

### Error

DataCloneError: Failed to execute 'postMessage' on 'MessagePort': (r) => {
Si(r).includes(Ne.CRUD) && this.iterateListeners((g) => { ... })
} could not be cloned.
https://.../vendor-powersync-*.js:9122:44

### Relevant source
- `@powersync/common/src/client/sync/bucket/SqliteBucketStorage.ts`:
```ts
this.updateListener = db.registerListener({
  tablesUpdated: (update) => {
    if (tables.includes(PSInternalTable.CRUD)) {
      this.iterateListeners((l) => l.crudUpdate?.());
    }
  }
});
```

### Suspected root cause

db.registerListener(...) is being invoked on a Comlink proxy (shared‑worker path), so the listener object (with a function) is being serialized for a MessagePort call. Comlink cannot clone functions, so postMessage throws DataCloneError.

### Notes

- This is only observed in production (SharedWorker/multi‑tab path is active).
- Disabling multi‑tab (forcing enableMultiTabs: false) appears to avoid the issue.

### Environment

- @powersync/web: 1.30.0
- @powersync/react: 1.8.2
- Chrome (prod, HTTPS)

longfellowone:
## Summary
- avoid binding `this` when invoking `_disposeTableChangeListener` to prevent Comlink cloning the adapter
- add regression test for `close()` disposer context
- add changeset for @powersync/web

## Testing
- pnpm test
- pnpm --filter @powersync/web exec vitest --run tests/locked_adapter_close.test.ts

Fixes #805

### DIFF
--- a/.changeset/fix-comlink-dispose-listener.md
+++ b/.changeset/fix-comlink-dispose-listener.md
@@ -1,0 +1,5 @@
+---
+"@powersync/web": patch
+---
+
+Avoid binding `this` when disposing table change listeners in the web adapter to prevent Comlink serialization errors on close.
--- a/packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
+++ b/packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
@@ -170,7 +170,10 @@ export class LockedAsyncDatabaseAdapter
    */
   async close() {
     this.closing = true;
-    this._disposeTableChangeListener?.();
+    const dispose = this._disposeTableChangeListener;
+    if (dispose) {
+      dispose();
+    }
     this.pendingAbortControllers.forEach((controller) => controller.abort('Closed'));
     await this.baseDB?.close?.();
     this.closed = true;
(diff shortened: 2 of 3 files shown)


stevensJourney (approved):
It's very strange that this issue wasn't detected in our other tests - or the manual multiple tabs testing we've recently been doing for testing other issues. 
I can't think of a good reason for production builds being different from the development builds in this specific scenario. 
That being said. The fact that `this` is being proxied is indeed an issue and the work here will prevent that.
The only nit I have is that we should probably add some comments for this behaviour in the code. I'll add that in a follow-up PR.
Thanks for the investigation, fix and contribution!

longfellowone:
Couldn't tell you why, only other info I can offer up is I have the service worker disabled in dev. To trigger it my code (worked every time) was roughly

disconnectAndClear()
close()
window.location.href

Issue #808, opened 2026-01-04
OPFSCoopSyncVFS: Database initialization timeout on page refresh (Safari)


longfellowone:
## Description

When using OPFSCoopSyncVFS, rapid page refreshes (especially pull‑to‑refresh on Safari iOS) can cause database initialization to time out. The new page can’t acquire OPFS file locks because the previous page’s worker hasn’t released them yet.

## Environment

- @powersync/web: 1.30.0
- Browser: Safari 26.1 (iOS and macOS)
- VFS: OPFSCoopSyncVFS

## Configuration

- VFS: OPFSCoopSyncVFS (Dedicated Worker for DB)
- Multi‑tabs: enabled (SharedWorker for sync)
- Issue occurs regardless of Sync Worker type — lock appears to be held by DB worker.

## Reproduction

1. Open an app using PowerSyncDatabase with OPFSCoopSyncVFS
2. Wait for database to initialize and connect
3. Pull‑to‑refresh (iOS) or Cmd+R (macOS) rapidly
4. After 1–3 refreshes, database initialization times out

## Expected

Database initializes successfully after refresh.

## Actual

Database initialization times out.

stevensJourney:
The steps to reproduce the issue above no longer result in an issue for me when testing the changes in https://github.com/powersync-ja/powersync-js/pull/786.
Note, that this PR also updates the `@journeyapps/wa-sqlite` dependency to `1.4.1` since there were also locking issues which were fixed there. 

https://github.com/user-attachments/assets/7a3a3548-8ac9-49bf-a800-ce2d0934396f

PR #810, opened 2026-01-04
Use cloneable abort reason for shared sync port close

longfellowone:
## Summary
- Ensure abort reasons used when closing shared sync client ports are cloneable across Comlink.
- Add a regression test for both credentials and upload abort paths.

## Changes
- SharedSyncImplementation now aborts with a plain string reason when removing a port.
- New tests in `shared_sync_abort_reason.test.ts` cover fetch-credentials and upload abort paths.
- Changeset for @powersync/web.

## Tests
- pnpm --filter @powersync/common run build
- pnpm --filter @powersync/web run build:tsc
- pnpm vitest --run tests/shared_sync_abort_reason.test.ts

### DIFF
--- a/.changeset/shared-sync-cloneable-abort.md
+++ b/.changeset/shared-sync-cloneable-abort.md
@@ -1,0 +1,5 @@
+---
+"@powersync/web": patch
+---
+
+Use a cloneable abort reason when closing shared sync client ports to avoid DataCloneError across Comlink.
--- a/packages/web/src/worker/sync/SharedSyncImplementation.ts
+++ b/packages/web/src/worker/sync/SharedSyncImplementation.ts
@@ -1,5 +1,4 @@
 import {
-  AbortOperation,
   BaseObserver,
   ConnectionManager,
   createLogger,
@@ -313,11 +312,10 @@ export class SharedSyncImplementation extends BaseObserver<SharedSyncImplementat
        * The port might currently be in use. Any active functions might
        * not resolve. Abort them here.
        */
+      const abortReason = 'Closing pending requests after client port is removed';
       [this.fetchCredentialsController, this.uploadDataController].forEach((abortController) => {
         if (abortController?.activePort == port) {
-          abortController!.controller.abort(
-            new AbortOperation('Closing pending requests after client port is removed')
-          );
+          abortController!.controller.abort(abortReason);
         }
       });
 
--- a/packages/web/tests/shared_sync_abort_reason.test.ts
+++ b/packages/web/tests/shared_sync_abort_reason.test.ts
@@ -1,0 +1,96 @@
+import * as Comlink from 'comlink';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { SharedSyncImplementation, type WrappedSyncPort } from '../src/worker/sync/SharedSyncImplementation';
+import type { AbstractSharedSyncClientProvider } from '../src/worker/sync/AbstractSharedSyncClientProvider';
+
+describe('Shared sync abort reasons', { sequential: true }, () => {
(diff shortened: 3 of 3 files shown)


stevensJourney:
Could you please add more context for the issue this PR is trying to solve. Specifically the error you were observing before these changes.

From a quick refresher on the code. The abort reason is used to when aborting the `uploadDataController`, which will `resolve` the pending `uploadCrud` operation - Which means we don't actually pass any error from this flow. Which, from what I can tell, should not cause any clone errors when updating the `SyncStatus` for connected clients. 

I haven't attempted to reproduce the issue this PR is trying to solve, but believe general cloning of unserializable Errors should be handled in https://github.com/powersync-ja/powersync-js/pull/786 which will be merged and released soon. 

https://github.com/powersync-ja/powersync-js/blob/e5253c9f31cbd9c38a3b6d429a938ad433eeeae7/packages/common/src/db/crud/SyncStatus.ts#L268

longfellowone:
Saw DataCloneError while digging into #808. The only value that is directly set in that teardown path is the abort reason
passed from `removePort()`. Making that reason a plain string ensures `signal.reason` is cloneable (even if the abort Event itself still isn’t). 

Based on my understanding, the path this targets (**fetchCredentialsController**):

- `SharedSyncImplementation.fetchCredentials` creates a Promise and sets `abortController.signal.onabort = reject`.
- `removePort()` aborts the controller, so the Promise rejects with the abort `Event` (whose `target.reason` is the
  string).
- If that rejection happens while the streaming loop is building a request (via `AbstractRemote.getCredentials()` →
  `fetchCredentials()`), the error can in some cases bubble to the streaming loop and be assigned to `dataFlow.downloadError` (see
  `AbstractStreamingSyncImplementation.ts`).
- `SharedSyncImplementation.updateAllStatuses()` broadcasts that status via Comlink, and `SyncStatus.toJSON()` does
  not sanitize errors.

stevensJourney:
> Saw DataCloneError while digging into #808. The only value that is directly set in that teardown path is the abort reason passed from `removePort()`. Making that reason a plain string ensures `signal.reason` is cloneable (even if the abort Event itself still isn’t).
> 
> Based on my understanding, the path this targets (**fetchCredentialsController**):
> 
> * `SharedSyncImplementation.fetchCredentials` creates a Promise and sets `abortController.signal.onabort = reject`.
> * `removePort()` aborts the controller, so the Promise rejects with the abort `Event` (whose `target.reason` is the
>   string).
> * If that rejection happens while the streaming loop is building a request (via `AbstractRemote.getCredentials()` →
>   `fetchCredentials()`), the error can in some cases bubble to the streaming loop and be assigned to `dataFlow.downloadError` (see
>   `AbstractStreamingSyncImplementation.ts`).
> * `SharedSyncImplementation.updateAllStatuses()` broadcasts that status via Comlink, and `SyncStatus.toJSON()` does
>   not sanitize errors.

I see. The serialization in `updateAllStatuses` should be fixed in PR 786. It seems like the `fetchCredentials` error is actually caused by calling `reject` with the Abort event from `onabort`.

IMO, a nicer fix is to pair sending the `AbortOperation` as the Error for `reject` and serialising the error correctly.

```diff 
          fetchCredentials: async () => {
            const lastPort = await this.getLastWrappedPort();
            if (!lastPort) {
              throw new Error('No client port found to fetch credentials');
            }
            return new Promise(async (resolve, reject) => {
              const abortController = new AbortController();
              this.fetchCredentialsController = {
                controller: abortController,
                activePort: lastPort
              };
-              abortController.signal.onabort = reject;
+              abortController.signal.onabort = () => {
+                reject(abortController.signal.reason);
+              };
```

PR #811, opened 2026-01-04
Defer OPFS worker termination during open on pagehide (Safari refresh lockups)

### LINKED ISSUE #808: OPFSCoopSyncVFS: Database initialization timeout on page refresh (Safari)

longfellowone:
Fixes #808

## Summary
- Prevent OPFS lock stalls on rapid refresh by deferring worker termination when pagehide fires while open is in-flight.
- Make pagehide handling idempotent and ensure a deferred close runs after open resolves.

## Changes
- WASQLiteOpenFactory: ignore repeat pagehide; if open is in-flight, defer close until after open completes; terminate worker on open error.
- Add opfs_pagehide regression tests (pagehide during open, idempotent pagehide, deferred close).

## Tests
- pnpm --filter @powersync/web run build:tsc
- pnpm vitest --run tests/opfs_pagehide.test.ts
- pnpm vitest

### DIFF
--- a/.changeset/opfs-pagehide-worker-terminate.md
+++ b/.changeset/opfs-pagehide-worker-terminate.md
@@ -1,0 +1,5 @@
+---
+"@powersync/web": patch
+---
+
+Defer dedicated OPFS worker termination on pagehide until open completes and ignore repeat pagehide events to avoid lock leaks.
--- a/packages/web/src/db/adapters/wa-sqlite/WASQLiteOpenFactory.ts
+++ b/packages/web/src/db/adapters/wa-sqlite/WASQLiteOpenFactory.ts
@@ -57,6 +57,8 @@ export class WASQLiteOpenFactory extends AbstractWebSQLOpenFactory {
       cacheSizeKb = DEFAULT_CACHE_SIZE_KB,
       encryptionKey
     } = this.waOptions;
+    const shouldForceCloseOnPagehide =
+      vfs === WASQLiteVFS.OPFSCoopSyncVFS || vfs === WASQLiteVFS.AccessHandlePoolVFS;
 
     if (!enableMultiTabs) {
       this.logger.warn('Multiple tabs are not enabled in this browser');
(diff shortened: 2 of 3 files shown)


stevensJourney:
It would be good to rebase this PR on https://github.com/powersync-ja/powersync-js/pull/786 for extra sanity. However, I do acknowledge there is still room for improvement there when it comes to Safari.

It looks like the gist of this PR is to terminate the OPFS worker from the previous page before the next page is loaded. I imagine there are some timing concerns where Safari might not be automatically terminating the previous dedicated worker in time.

A few clarifying questions:

*What exactly "times out"?* Looking at the code, I don't see any current timeout for these lock aquisitions. If there's no timeout, the new worker should just wait indefinitely until the old worker is terminated and releases its lock. Is this a hang rather than a timeout? 

*Worker termination should release locks automatically*. When the browser terminates a worker, any OPFS SyncAccessHandles and locks held by that worker should be released. If the old worker is eventually terminated on page unload (even if delayed), the new worker's lock acquisition should eventually succeed.

*Is the bfcache involved?* I notice the test sets persisted: true on the `pagehide` event. Is Safari potentially keeping the old page (and its worker) alive in the back-forward cache, preventing lock release entirely?

*For extra context*: Could you share which lock is being held that causes this issue? Note that `navigator.locks.query()` will show Web Locks API locks (like the `open-wasqlite-db` lock used in the worker).

simolus3:
For context, I ran into a similar issue with Safari and the Dart SDK: If a tab navigates to a new location causing a reload _while a worker holds a navigator lock_, that lock is never returned. I have opened [this WebKit issue](https://bugs.webkit.org/show_bug.cgi?id=316904) about it.

Because we use navigator locks to coordinate access to databases on the web, this can cause deadlocks on Safari. The question is, how do we fix it. While using the `pagehide` event would likely work, there's no guarantee that it shuts down the worker in time. The underlying bug is still there and possibly reachable depending on race conditions, but it would be harder to reproduce.

Perhaps a better workaround would be to let tabs acquire navigator locks instead of doing that in the worker. But we already know that Safari has its own issues with that because tabs owning locks can be suspended on iOS, also causing deadlocks. Maybe we can use a shared worker for coordination, it could "ping" tabs to see if they're still active and otherwise manage locks in its own state.

Another question though: If a worker doesn't clean up locks consistently, is it that safe to assume it clears synchronous OPFS file handles? If not, all of this wouldn't help and the database would still be blocked. So there's definitely more to experiment with here.

@longfellowone As a workaround you could adopt earlier, does it help to close the PowerSync database in a `pagehide` listener of your app? Or does that not close resources fast enough?

PR #813, opened 2026-01-08
Added the react-native-background-sync project to demos

bean1352:
Added the react-native-background-sync project to demos. This project showcases how to use Expo's BackgroundTask API to connect to PowerSync, allowing the local SQLite database to stay in sync with the source database when the mobile app is backgrounded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Demo applications are located in the [`demos/`](./demos/) directory. Also see ou
 
 - [demos/react-native-supabase-todolist](./demos/react-native-supabase-todolist/README.md): A React Native to-do list example app using a Supabase backend.
 - [demos/react-native-supabase-group-chat](./demos/react-native-supabase-group-chat/README.md): A React Native group chat example app using a Supabase backend.
+- [demos/react-native-supabase-background-sync](./demos/react-native-supabase-background-sync/README.md): A React Native example app demonstrating background sync using PowerSync, Expo and Supabase.
 - [demos/react-native-web-supabase-todolist](./demos/react-native-web-supabase-todolist/README.md) A React Native to-do list example app using a Supabase backend that's compatible with React Native for Web.
 - [demos/django-react-native-todolist](./demos/django-react-native-todolist/README.md) A React Native to-do list example app using a Django backend.
 


LucDeCaf:
Is there a way to get some initial data on the client? AFAIK since it's using anonymous sign in, nothing's being synced down by default.

Edit: Got some creds with data, seems to work 👍

demos/react-native-supabase-background-sync/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Run on Android
+
+```sh
+pnpm android
+```
+
+## Supabase and background sync
+
+The `SupabaseConnector` has been configured to override the default `fetch` function that the `@supabase/supabase-js` SDK uses. By default, the `fetch` function used does not work in the background [see here](https://github.com/facebook/react-native/issues/47437), so we need to override it with a something that works in the background, like `expo/fetch`.
```
LucDeCaf: Nit: "...with ~a~ something..."

demos/react-native-supabase-background-sync/README.md
```diff
@@ -0,0 +1,168 @@
...
+adb shell cmd jobscheduler run -f com.anonymous.powersyncreactnativebackgroundsync 0
+```
+
+### iOS
+
+Background tasks are not supported on iOS simulators.
+
+## ⚠️ Please note
+
+The app needs to be in the background for it to run and ultimately, the OS decides when to run background tasks so forcing it to run might not always work immediately. The background task should run on it's own within 20 minutes after the app is put in the background. Testing background sync seems to work better and more consistently on a physical device. If you don't see any logs after starting your background task, you may need to watch all the logs on the device using `adb logcat` (on Android).
```
LucDeCaf: Nit: "~it's~ its"

LucDeCaf (approved):
This is super neat, thanks for this 🙌

LucDeCaf (dismissed):

Chriztiaan (approved):

Chriztiaan:
Minor notes:
1. Please add a reference to this example in the root readme.
2. The tests pass locally for me, not sure where the github action failure is coming from.

Issue #799, opened 2025-12-21
Unable to Build after switching to Capacitor


wills01:
Get the following Build Error:

`unhandledRejection TypeError: Class extends value undefined is not a constructor or null
    at 36115 (/Users/nickwills/Documents/source/CloudGAApp/.next/server/chunks/2818.js:109:17621)
    at t (/Users/nickwills/Documents/source/CloudGAApp/.next/server/webpack-runtime.js:1:127)
    at /Users/nickwills/Documents/source/CloudGAApp/.next/server/chunks/6814.js:3:14717
    at t.a (/Users/nickwills/Documents/source/CloudGAApp/.next/server/webpack-runtime.js:1:866)
    at 25798 (/Users/nickwills/Documents/source/CloudGAApp/.next/server/chunks/6814.js:3:14651)
    at t (/Users/nickwills/Documents/source/CloudGAApp/.next/server/webpack-runtime.js:1:127)
    at /Users/nickwills/Documents/source/CloudGAApp/.next/server/chunks/6814.js:1:437
    at t.a (/Users/nickwills/Documents/source/CloudGAApp/.next/server/webpack-runtime.js:1:866)
    at 6814 (/Users/nickwills/Documents/source/CloudGAApp/.next/server/chunks/6814.js:1:83)
    at t (/Users/nickwills/Documents/source/CloudGAApp/.next/server/webpack-runtime.js:1:127) {
  type: 'TypeError'

stevensJourney:
To help diagnose the problem, I need a bit more information:

**Next.js + Capacitor Setup:**

- Are you using Next.js with Capacitor? If so, how do you have it configured?
- Are you using static export mode (output: 'export' in your next.config.js)?
- Can you share your next.config.js and capacitor.config.ts files?

**PowerSync Package Information:**

- Which PowerSync packages and versions are you using? (e.g., `@powersync/capacitor`, `@powersync/web`, etc.)
- Can you share the relevant parts of your package.json?

**Additional Context:**

- When does this error occur - during next build, at runtime, or when importing specific PowerSync modules?
- Can you share the import statements where you're using PowerSync in your code?
- Are you importing `@powersync/capacitor` in any server-side code, server components, or API routes?

[...]

Issue #542, opened 2025-03-25
SyntaxError: Unexpected token 'export'


mednche:
I am running: 
- react@18.3.1
- next@15.1.7
- @powersync/common@1.25.0
- @powersync/react@1.5.2
- @powersync/web@1.15

I get `SyntaxError: Unexpected token 'export'` when I run `import { usePowerSync } from '@powersync/react';`
The error message says the error is located in `powersync/react/lib/index.js`.

I have added `'use client';` at the top of all the files that deal with powersync. 

Any idea what is going on? Is this an incompatibility with the version of React or Nextjs I have installed?

michaelbarnes:
Hey @mednche  can you share a little more information on how you're compiling the app? Are you using Turbopack for Next.js application?

stevensJourney:
The `unexpected token export` error is usually caused when the `@powersync/web` package is being imported in a Node.js or SSR environment. We should update our `@powersync/web` package to properly support Node.js ESM loading.

The best practise though, is to avoid loading PowerSync in SSR entirely. There are examples for doing this with `next/dynamic` in our `example-nextjs` [Demo](https://github.com/powersync-ja/powersync-js/blob/main/demos/example-nextjs/src/components/providers/DynamicSystemProvider.tsx).

stevensJourney:
This has been fixed in [@powersync/web@1.32.0](https://github.com/powersync-ja/powersync-js/releases/tag/%40powersync%2Fweb%401.32.0)

PR #815, opened 2026-01-09
Fix Web and Capacitor ESM exports

stevensJourney:
Fixes:

- https://github.com/powersync-ja/powersync-js/issues/799
- https://github.com/powersync-ja/powersync-js/issues/542

# Overview

The `@powersync/web` and `@powersync/capacitor` packages don't explicitly declare themselves as a ES Modules, but they use ES syntax such as `import` and `export`.

This is usually fine when building web apps since a bundler typically processes the codebase.

Serverside rendering environments, such as Next.js, will by default, attempt to import and evaluate the PowerSync web SDK in a Node.js (server) environment. The Node.js process will complain when it attempts to run the `import` or `export` statements outside of an ES context. This can result in errors such as:

```
Unexpected token 'export'
```

Which is caused by Node.js trying to import the PowerSync web module which contains ES code without delcaring itself as an ES module.

OR:

```
`unhandledRejection TypeError: Class extends value undefined is not a constructor or null
```

Which is a downstream error which typically manifests if the upstream PowerSync web resources/classes could not be imported.

OR (A Capacitor example):

```
Error: Failed to load external module @powersync/capacitor-83c647fee4bde6c9/dist/esm/index.js: SyntaxError: Unexpected token 'export'
    at <unknown> (src/services/PowerSyncProvider.tsx:3:1)
    at module evaluation (.next/dev/server/chunks/ssr/[root-of-the-server]__28e8abff._.js:1123:30)
    at <unknown> (src/pages/_app.tsx:1:1)
    at module evaluation (.next/dev/server/chunks/ssr/[root-of-the-server]__28e8abff._.js:1283:30)
    at Object.<anonymous> (.next/dev/server/pages/_app.js:7:3)
```

### Declarations 

This updates both packages to explicitly declare ESM:
 
- Modules are declared with `"type": "module"`
- Imports are updated to use the `.js` extension
- `package.json` `exports` are updated/added

### SSR

Being able to import the packages in a SSR environment is a start. If the `PowerSyncDatabase` client is accidentally used in the server environment, it should also not throw any exceptions which could cause rendering issues. We historically provide a SSR version of the SDK which warns about usage and returns empty results for any queries which might have been executed.

This implementation is currently broken due to the PowerSync core version check failing (due to no results being returned) and the offline sync status could not be resolved.

The client has now been updated to work as a placeholder in SSR environments.

## Testing

We have some tests which attempt to `import` our packages in various Node.js environments. These packages are now imported there to ensure that SSR should work correctly

Tests Done:

- Verified that both packages can be successfully imported in a standalone Next.js 16 app without using `next/dynamic`.
- Updated the Are the types wrong tests to reflect ES better.
- Tested React Supabase Todolist App
- Tested React Web Supabase Todolist App
- Tested Example Capacitor App
- Updated `example-nextjs` demo to use Next 16 (currently at 14)

### DIFF
--- a/demos/angular-supabase-todolist/README.md
+++ b/demos/angular-supabase-todolist/README.md
@@ -1,16 +1,15 @@
 # PowerSync + Supabase Angular Demo: Todo List
 
-## Note: Alpha Release
-
-This demo is currently in an alpha release.
-
 ## Overview
 
 Demo app demonstrating use of the [PowerSync SDK for Web](https://www.npmjs.com/package/@powersync/web) together with Supabase.
 
-A step-by-step guide on Supabase<>PowerSync integration is available [here](https://docs.powersync.com/integration-guides/supabase-+-powersync).
+## Run Demo
 
-## Quick Start
+Prerequisites:
+* To run this demo, you need to have properly configured Supabase and PowerSync projects. Follow the instructions in our Supabase<>PowerSync integration guide:
+  * [Configure Supabase](https://docs.powersync.com/integration-guides/supabase-+-powersync#configure-supabase)
+  * [Configure PowerSync](https://docs.powersync.com/integration-guides/supabase-+-powersync#configure-powersync)
 
 1. Run `pnpm install`
 2. Create a `.env` file by copying the template `cp .env.template .env`
--- a/demos/react-native-supabase-todolist/README.md
+++ b/demos/react-native-supabase-todolist/README.md
@@ -4,10 +4,13 @@
 
 Demo app demonstrating use of the [PowerSync SDK for React Native](https://www.npmjs.com/package/@powersync/react-native) together with Supabase.
 
-A step-by-step guide on Supabase<>PowerSync integration is available [here](https://docs.powersync.com/integration-guides/supabase-+-powersync).
-Follow all the steps until, but not including, [Test Everything (Using Our Demo App)](https://docs.powersync.com/integration-guides/supabase-+-powersync#test-everything-using-our-demo-app).
+## Run Demo
 
-## Getting Started
+Prerequisites:
+* To run this demo, you need to have properly configured Supabase and PowerSync projects. Follow the instructions in our Supabase<>PowerSync integration guide:
+  * [Configure Supabase](https://docs.powersync.com/integration-guides/supabase-+-powersync#configure-supabase)
+  * [Configure PowerSync](https://docs.powersync.com/integration-guides/supabase-+-powersync#configure-powersync)
+* Follow all the steps until, but not including, [Test Everything (Using Our Demo App)](https://docs.powersync.com/integration-guides/supabase-+-powersync#test-everything-using-our-demo-app).
 
 Switch into the demo's directory:
 
--- a/demos/react-native-web-supabase-todolist/README.md
+++ b/demos/react-native-web-supabase-todolist/README.md
@@ -6,9 +6,12 @@ This demo app is an extension of the [Supabase Todo List App](../react-native-su
 
 To use PowerSync in your own React Native for Web project, additional config is required. This is detailed in our docs [here](https://docs.powersync.com/client-sdk-references/react-native-and-expo/react-native-web-support).
 
-To run this demo, follow these instructions:
+## Run Demo
 
-## Running this demo
+Prerequisites:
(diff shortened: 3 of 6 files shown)


packages/web/package.json
```diff
@@ -49,7 +53,8 @@
     "build:prod": "pnpm run build:tsc --sourceMap false && pnpm run build:webpack-main && pnpm run build:webpack-workers",
     "clean": "rm -rf lib dist tsconfig.tsbuildinfo",
     "watch": "tsc --build -w",
-    "test": "pnpm build && vitest"
+    "test": "pnpm build && vitest",
+    "test:exports": "attw --pack . --entrypoints . --profile=esm-only "
```
stevensJourney, line 57: While we do export a UMD bundled version of the SDK - we don't bundle WA-SQLite  in that bundle currently. This causes issues if one were to actually `require('@powersync/web)` or `require('@powersync/web/umd')`. Since the commonjs will try and require the ES `wa-sqlite` module.

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/stevenontong/Documents/platform_code/powersync/powersync-react-native-sdk/node_modules/@journeyapps/wa-sqlite/src/sqlite-api.js from /Users/stevenontong/Documents/platform_code/powersync/powersync-react-native-sdk/packages/web/dist/index.umd.cjs not supported.
Instead change the require of sqlite-api.js in /Users/stevenontong/Documents/platform_code/powersync/powersync-react-native-sdk/packages/web/dist/index.umd.js to a dynamic import() which is available in all CommonJS modules.
```

I imagine the CJS SSR requirements for the Web SDK are not very large, so not optimising for this at this stage.

packages/web/tests/src/db/PowersyncDatabase.test.ts
```diff
@@ -9,7 +9,7 @@ describe('PowerSyncDatabase', () => {
   let mockSyncImplementation: any;
 
   beforeEach(() => {
-    const logger = createBaseLogger();
+    const logger = createLogger('test');
```
stevensJourney, line 12: Why this change?

Changing the TypeScript config from `"module": "ESNext"` to `"module": "NodeNext"` (with `"moduleResolution": "nodenext"`) applies stricter ESM interop rules during type-checking.

The `createBaseLogger()` function returns `typeof Logger`. Under the previous looser resolution, TypeScript treated this as compatible with `ILogger`. With `NodeNext`, TypeScript correctly interprets this as the module namespace type rather than the default `Logger` export.

```typescript
import Logger, { type ILogger, type ILogLevel } from 'js-logger';

export function createBaseLogger() {
  return Logger; // Return type is `typeof Logger` (module namespace), not `ILogger`
}
```

Since `PowerSyncDatabaseOptions` expects `logger: ILogger`, passing the result of `createBaseLogger()` directly now causes a type error in these tests.

Using `createLogger('name')` is the intended approach here anyway, as it returns a properly typed `ILogger` instance.

packages/kysely-driver/package.json
```diff
@@ -9,12 +9,12 @@
   "exports": {
     ".": {
       "import": {
-        "default": "./lib/src/index.js",
-        "types": "./lib/src/index.d.ts"
+        "types": "./lib/src/index.d.ts",
```
stevensJourney, line 12: Changing the order of these fixes warnings like

```
  
  Warning: G] The condition "types" here will never be used as it comes after "default" [package.json]
  
      package.json:17:8:
        17 │         "types": "./dist/index.d.cts"
           ╵         ~~~~~~~

```

demos/example-nextjs/next.config.js
```diff
@@ -2,26 +2,5 @@ module.exports = {
...
-          },
-          {
-            test: /\.scss/,
-            use: ['style-loader', 'css-loader', 'sass-loader']
-          }
-        ]
-      }
-    };
-  }
+  turbopack: {}
```
Chriztiaan, line 5: Nice! This means we could consider updating the Next tutorial in our docs. https://docs.powersync.com/tutorials/client/sdks/web/next-js It specifically mentions not using Turbopack.

FYI @bean1352 @benitav

Chriztiaan:
Happy with the changes, tested example-next, example-capacitor, react-supabase-todolist on my side.

Chriztiaan (approved):

PR #816, opened 2026-01-09
More explicit Supabase demo prerequisites

benitav:
Rather than just linking to the docs guide. Extension of https://github.com/powersync-ja/powersync-js/pull/814


benitav:
A test keeps on failing due to a time-out. These are just README updates so I think it should still be safe to merge.

Issue #743, opened 2025-10-22
@powersync/op-sqlite React Native 0.81 compatibility


a11rew:
The current latest version of `@powersync/op-sqlite` doesn't work in React Native 0.81 on Android. It fails at runtime with Turbomodule errors coming from the op-sqlite dep. 

The version of op-sqlite currently used by `@powersync/op-sqlite` (^14.0.2) is outdated, op-sqlite was updated with RN 0.81 support in [15.0.0](https://github.com/OP-Engineering/op-sqlite/releases/tag/15.0.0). Overriding the version of op-sqlite used by `@powersync/op-sqlite` to 15.0.7 seems to work without issues so this should likely be a straightforward version bump to fix? Happy to send a PR in for this

simolus3:
We currently have a peer-dependency on `"@op-engineering/op-sqlite": "^13.0.0 || ^14.0.0"`. Given that we don't seem to be affected by any 15.x changes, changing that to include `|| ^15.0.0` sounds good to me. You could also bump the dev-dependency on `op-sqlite` to ensure we're testing with the latest version. A PR for that would definitely be welcome, thanks!

PR #818, opened 2026-01-09
chore: Bump op-sqlite peer dependency to 13.x.x || 14.x.x || 15.x.x for @powersync/op-sqlite

kobiebotha:
Fixes https://github.com/powersync-ja/powersync-js/issues/743

Tests performed:
- ran `powersynctests` on:
  -  iOS 26.2
  - Android API 34
 - ran detox tests on iOS and android

improved the README while I was here

### DIFF
--- a/.changeset/neat-papayas-refuse.md
+++ b/.changeset/neat-papayas-refuse.md
@@ -1,0 +1,13 @@
+---
+'@powersync/common': minor
+'@powersync/node': minor
+'@powersync/react-native': minor
+'@powersync/web': minor
+---
+
+The [Rust sync client](https://www.powersync.com/blog/speeding-up-powersync-with-a-sqlite-extension-written-in-rust) is now enabled by default.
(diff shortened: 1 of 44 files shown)


packages/powersync-op-sqlite/package.json
```diff
@@ -65,7 +65,7 @@
     "access": "public"
   },
   "peerDependencies": {
-    "@op-engineering/op-sqlite": "^13.0.0 || ^14.0.0",
+    "@op-engineering/op-sqlite": "^13.0.0 || ^15.0.0",
```
simolus3: Why would we exclude version 14.x now? Since we didn't have to change anything, I assume we're compatible with the three major versions:

```suggestion
    "@op-engineering/op-sqlite": "^13.0.0 || ^14.0.0 || ^15.0.0",
```

kobiebotha: syntax snafu on my side, thanks!

PR #819, opened 2026-01-12
Fix: DataStream Race Condition (and flaky unit tests)

stevensJourney:
# Overview

Sync tests were intermittently timing out, particularly the `reports progress > without priorities` test for the Rust client. The tests would hang indefinitely waiting for sync progress updates that never arrived, even though the data was clearly being received from the mock sync service.

**Crucially, these failures were not reproducible locally.** The issue only manifested reliably when running tests in CI within a resource-constrained Docker container with reduced CPU allocation. This made debugging particularly challenging, as the race condition's timing window was too narrow to trigger on developer machines with full CPU resources.

### Initial Observations

When running the tests, interleaved logs were observed suggesting data was flowing through the system:

- The checkpoint was received ✓
- The data operations were received ✓
- The Rust core was returning correct `UpdateSyncStatus` instructions with the right progress values ✓
- But `waitForProgress(database, [10, 10])` never resolved

The sync loop was stuck waiting for data that appeared to already exist in the system.

### Following the Data Flow

The data flow was traced through the system:

1. **Mock service** pushes sync lines to a `ReadableStream` ([tests/utils.ts](packages/node/tests/utils.ts))
2. **`postStreamRaw`** in [`AbstractRemote`](packages/common/src/client/sync/stream/AbstractRemote.ts) reads from this stream and enqueues lines into a [`DataStream`](packages/common/src/utils/DataStream.ts)
3. **Sync loop** in [`AbstractStreamingSyncImplementation`](packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts) calls `DataStream.read()` to consume lines
4. **`read()`** registers a listener and calls `processQueue()` to trigger delivery

The breakdown was happening between steps 2 and 3.

### The Smoking Gun

After adding extensive logging, a critical timing issue was discovered:

```
enqueueData: pushed data, queue length: 2
processQueue: starting new _processQueue
_processQueue: start, queue length: 2 hasDataReader: false   // ← No reader!
_processQueue: early return - closed: false hasDataReader: false
stream waiting   // ← Sync loop calls read() AFTER _processQueue returned
processQueue: returning early, promise exists
processQueue: finally clearing processingPromise
```

**The data was being enqueued before the sync loop had a chance to call `read()` and register a listener.**

The problematic code was in `processQueue()`:

```typescript
protected processQueue() {
  if (this.processingPromise) {
    return;  // ← Silently drops the call!
  }

  const promise = (this.processingPromise = this._processQueue());
  promise.finally(() => {
    this.processingPromise = null;  // ← Runs as microtask AFTER promise resolves
  });
  return promise;
}
```

And in `_processQueue()`:

```typescript
protected async _processQueue() {
  // ...
  if (this.isClosed || !this.hasDataReader()) {
    return;  // ← Returns early if no reader registered yet
  }
  // ...
}
```

When `read()` was called to register a listener, it would call `processQueue()`:

```typescript
async read(): Promise<ParsedData | null> {
  return new Promise((resolve, reject) => {
    const l = this.registerListener({
      data: async (data) => {
        resolve(data);
        l?.();
      },
      // ...
    });

    this.processQueue();  // ← This call was being dropped!
  });
}
```

### Understanding the Race Condition

In the original implementation, reading from the HTTP stream was **reactive**, triggered by a `lowWater` callback when the queue needed more data:

```typescript
// Original: reading was triggered by lowWater callback
const stream = new DataStream({
  lowWater: async () => {
    const { done, value } = await reader.read(); // Only read when queue is low
    // ... decode and enqueue data ...
  }
});
```

Here's what was happening:

1. The `lowWater` callback triggers a read from the HTTP stream, which calls `stream.enqueueData(line)`

2. `enqueueData()` pushes data to the queue and calls `processQueue()`

3. `processQueue()` starts `_processQueue()` and stores the promise in `processingPromise`

4. `_processQueue()` checks `hasDataReader()` - but returns **false** because the sync loop hasn't called `read()` yet!

5. `_processQueue()` returns early (no reader to deliver data to)

6. Meanwhile, the sync loop finishes processing the previous line and calls `read()`

7. `read()` registers a listener and calls `processQueue()`

8. **But `processQueue()` sees `processingPromise` still exists** (the `.finally()` hasn't run yet) and returns early!

9. The `.finally()` callback runs, clearing `processingPromise`

10. **Nobody re-triggers processing.** Data sits in the queue. The sync loop waits forever.

The core issue was that **calls to `processQueue()` were being silently dropped** when processing was already in progress. When the sync loop's `read()` call triggered `processQueue()`, it was ignored because the previous (now-useless) processing was still technically "in progress".

### Why the `.finally()` Timing Matters

The `processingPromise` was cleared using `.finally()` on the promise:

```typescript
promise.finally(() => {
  this.processingPromise = null;
});
```

The `.finally()` callback runs as a **microtask after the promise resolves**. But when `_processQueue()` returns and resolves its promise, other microtasks may already be queued—including the sync loop's continuation from its previous `await read()`.

This meant the sync loop could call `read()` → `processQueue()` **before** the `.finally()` callback cleared `processingPromise`, causing the call to be dropped.

# The Solution

## Await Pending Processing in `read()`

The fix is simple: in `read()`, the code now **waits for any pending processing to complete** before registering a listener:

```typescript
async read(): Promise<ParsedData | null> {
  if (this.closed) {
    return null;
  }

  // Wait for any pending processing to complete first.
  // This ensures the listener is registered before calling processQueue(),
  // avoiding a race where processQueue() sees no reader and returns early.
  if (this.processingPromise) {
    await this.processingPromise;
  }

  return new Promise((resolve, reject) => {
    const l = this.registerListener({
      data: async (data) => {
        resolve(data);
        l?.();
      },
      // ...
    });

    this.processQueue();  // Now guaranteed to start fresh!
  });
}
```

## How the Fix Works

Now when the race condition occurs:

1. `consumeStream()` (or `lowWater` callback) enqueues data, calls `processQueue()`

2. `processQueue()` starts `_processQueue()`, sets `processingPromise`

3. `_processQueue()` finds no reader, returns early

4. Sync loop calls `read()`

5. `read()` sees `processingPromise` exists → **awaits it**

6. The first `_processQueue()` completes, `.finally()` runs, clears `processingPromise`

7. `read()` continues after the await, registers listener

8. `read()` calls `processQueue()` → no `processingPromise` exists, so it starts fresh!

9. `_processQueue()` runs, finds the reader **and** the data, delivers it!

10. Sync loop receives the data and continues ✓

The key insight is that by awaiting the pending promise, the `.finally()` callback is guaranteed to have run and cleared `processingPromise` before the listener is registered and a new processing cycle is triggered. This ensures the `processQueue()` call will actually start processing instead of being silently dropped.

# Additional Changes

## Proactive Stream Consumption

The `postStreamRaw` method was refactored to use a proactive `consumeStream()` loop instead of relying on the `lowWater` backpressure callback to trigger reads. This decouples reading from the network from the DataStream's internal processing state, making the system more predictable:

```typescript
const consumeStream = async () => {
  while (!stream.closed && !abortSignal?.aborted && !readerReleased) {
    const { done, value } = await reader.read();
    // ... process and enqueue data ...

    // Apply backpressure if queue is too full
    if (stream.dataQueue.length > stream.highWatermark) {
      await new Promise((resolve) => {
        stream.registerListener({ lowWater: async () => resolve() });
      });
    }
  }
};

consumeStream().catch((ex) => this.logger.error('Error consuming stream', ex));
```

This approach:

- Reads from the network continuously (as fast as backpressure allows)
- Enqueues data immediately when available
- Only pauses when the queue is full (highWatermark exceeded)
- Is independent of whether there's currently a reader waiting

# Why This Went Undetected

This race condition likely went undetected in production because:

1. **Real-world sync streams are continuous** - In actual usage, the sync service sends a steady stream of events. Even if one `processQueue()` call is dropped, subsequent data arrivals trigger new processing cycles, and the sync loop eventually receives all data.

2. **Unit tests have different timing characteristics** - These tests send isolated events (e.g., a single checkpoint, a single data batch) and immediately inspect state. This creates a narrow window where the race condition manifests: data is enqueued, but the test checks results before the sync loop has processed it.

3. **The race window is extremely small** - It only occurs when `enqueueData()` and `read()` interleave in a specific order relative to microtask scheduling. In continuous streaming, this timing is unlikely to cause permanent data loss.


packages/common/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -573,7 +572,11 @@ export abstract class AbstractRemote {
 
     const stream = new DataStream<T, string>({
       logger: this.logger,
-      mapLine: mapLine
+      mapLine: mapLine,
+      pressure: {
+        highWaterMark: 20,
+        lowWaterMark: 10
```
simolus3, line 579: Why wouldn't the default limits work?

simolus3 (commented):
This feels like a polyfill for async generators at this point, except that we have an internal buffer and lots added internal complexity and state machinery for no apparent reason. I agree with the fix as a quick solution to get tests running again, but I think we should throw out the whole thing sooner than later.

The logic in `DataStream` supports multiple publishers and consumers. We don't have multiple consumers anywhere and only use multiple publishers in the Rust sync client which might as well have its specialized implementation for this. We also don't need to support `close()`, we could just consistently use an `AbortSignal` to keep the flow linear. Using a less generalized interface without support for all the stuff we're not using would probably help a lot here.

This PR is obviously an urgently needed improvement, but I would love to take a look at simplifying this.

stevensJourney:
> This feels like a polyfill for async generators at this point

This has always been the case (or more accurately, a polyfill for ReadableStreams). Put differently, this has attempted to provide similar behaviour - plus extra notifications for stream capacity events (highwater/lowwater, used by the websocket implementation).

This was implemented at a time when we wanted to lessen the need for explicit polyfills in the dev setup of the SDK. These days we do pre-bundle `@powersync/common` etc, so using a standard polyfill could be a good option.

packages/common/src/client/sync/stream/AbstractRemote.ts
stevensJourney: The defaults would work. However, setting lowwater to 10 instead of 0 means we start refilling the buffer before it's completely drained. This helps maintain a constant stream of input without waiting periods where the buffer is empty and we're waiting for the next upstream read.

simolus3 (approved):
Looks good to me, thanks for the investigation and the fix 👍

PR #821, opened 2026-01-13
Remove `DataStream`, support binary sync lines over HTTP

simolus3:
This refactors `AbstractRemote` and `AbstractStreamingSyncImplementation` to have clearer control flow and to support streaming BSON documents over HTTP.

In our SDK, we're using the `DataStream` utility as a polyfill replacing parts of the `ReadableStreams` web API and async iterables in JavaScript (since RN supports neither). However, that abstraction turned out to hurt us quite a bit:

1. For some original sync tests, [race conditions where causing events to get lost](https://github.com/powersync-ja/powersync-js/commit/14c3404887a83d747ed0597fbafc528ee21e9280#diff-3f7d7e2c648a8e32fc8401253692f35a3af64be15bfa136fc7be95bfca113cbc).
2. For the Rust sync client, we had a number of issues where we were trying to post completed token refreshs or crud upload notifications into a closed stream, causing exceptions
3. Most recently, there was another race condition causing CI failures: https://github.com/powersync-ja/powersync-js/pull/819

Clearly, we're having a hard time using `DataStream` correctly. By design, `DataStream` is the equivalent of an asynchronous channel supporting multiple producers and multiple consumers. This makes implementing it quite complicated. However, we don't need most of that in practice!

1. We're only using `DataStream` to iterate over sync lines in an async `while` loop. There is never any concurrency on the consuming side, and there is never more than one consumer.
2. There are technically multiple producers, but there is one clear "main" producer and another one adding secondary events like crud uploads. So we can scope the lifetime of the stream to the main producer.

This PR replaces `DataStream` with a variant of `AsyncIterator` that only has a `next` method. This encodes the expectation that we'll only ever have one listener owning the stream. I have also deliberately hidden the `return` method. Instead, the only way to cancel a stream is to pass an `AbortSignal` to the place creating it and marking that as aborted. This simplifies handling streams in the remote implementation, and makes data flow a bit easier to understand:

1. When reading streams via `fetch`, no translation is necessary because the default response reader already implements the format we need.
2. To translate between the push-based mechanism from RSocket and async iterators, we can use the `event-iterator` package we already depend on and bundle. We just need to rewrite `Symbol.asyncIterator` to a polyfill for RN.
3. Things like parsing lines as BSON or JSON are now a stream transformation instead of being done in `AbstractRemote`, making control flow easier to reason about.

This also allows us to reuse the networking logic between the legacy and the Rust sync client. Finally, I've added support for receiving binary sync lines over HTTP, a feature we support in our Kotlin and Dart SDKs.

## Benchmarks

I ran a very simple benchmark in the node demo app that consists of:

1. Connecting and waiting for the initial sync without any stream subscription.
2. Subscribing to a stream with 1M rows and waiting for that to sync for the first time (this is not a great benchmark since the sync-local step takes around 3 seconds here, but if something was truly horrible here we'd see it).

| Benchmark         | Before        | After  |
|-------------------|---------------|--------|
| RSocket | 9.97s         | 9.52s  |
| HTTP, text        | 10.93s        | 10.77s |
| HTTP, bson        | (unsupported) | 10.74s |


rkistner (commented):
I like the change to remove DataStream - I agree plain AsyncIterators are better here.

I do think we need to test this properly to make sure there are no regressions. I also wonder whether we should stick to JSON for HTTP streams to start with, and make it either opt-in, or part of a major release to switch to BSON? Apart from potential bugs, I've seen cases where BSON is actually _larger_ than the equivalent JSON for the same amount of data.

What is the default for HTTP in other SDKs currently?

packages/common/src/utils/stream_transform.ts
```diff
@@ -0,0 +1,255 @@
...
+
+        const chunk = upstreamEvent.value;
+        for (let i = 0; i < chunk.length; ) {
+          const availableInData = chunk.length - i;
+
+          if (objectBody) {
+            // We're in the middle of reading a BSON document.
+            const bytesToRead = Math.min(availableInData, remainingLength);
+            const copySource = new Uint8Array(chunk.buffer, chunk.byteOffset + i, bytesToRead);
+            objectBody.set(copySource, 4 + copySource.length - remainingLength);
```
rkistner: This offset calculation appears incorrect. Take this test as an example that currently fails:

```ts
  test('splits bson objects (2)', async () => {
    async function* source() {
      yield new Uint8Array([5, 0, 0, 0, 1]);

      yield new Uint8Array([6, 0, 0, 0, 2]);
      yield new Uint8Array([3]);
    }

    const stream = extractBsonObjects(source());
    const a = (await stream.next()).value;
    const b = (await stream.next()).value;
    expect(a).toHaveLength(5);
    expect(b).toHaveLength(6);
    expect(Array.from(a)).toEqual([5, 0, 0, 0, 1]);
    expect(Array.from(b)).toEqual([6, 0, 0, 0, 2, 3]);
    expect(await stream.next()).toStrictEqual(doneResult);
  });
```

simolus3: 🤦 Thanks, I probably got `objectBody` and `copySource` confused.

simolus3:
> What is the default for HTTP in other SDKs currently?

[Dart](https://github.com/powersync-ja/powersync.dart/blob/405687f8064babc9ba2904a88ca6dd47c96c6803/packages/powersync_core/lib/src/sync/streaming_sync.dart#L655), [Kotlin](https://github.com/powersync-ja/powersync-kotlin/blob/1731a344d49efe4842b2d09527182426a4a60c5c/common/src/commonMain/kotlin/com/powersync/sync/StreamingSync.kt#L511) and [native](https://github.com/powersync-ja/powersync-native/blob/ca77f9b6b62541bd6818662fa8b9b7634869bff6/powersync/src/sync/download/http.rs#L33-L36) all unconditionally prefer BSON when the Rust client is used, they'd only use JSON on older service versions.

> I've seen cases where BSON is actually _larger_ than the equivalent JSON for the same amount of data.

To be fair, I would expect the (currently JSON strings) oplog data lines to make up the bulk of traffic for PowerSync. So whether the container around that is BSON or JSON is probably insignificant.

> and make it either opt-in, or part of a major release to switch to BSON

I'm not a fan of making this an option. If using BSON has benefits, we should aim to use it by default. If it's worse, we shouldn't have it in our SDKs at all.
As far as I can remember, the original motivation for BSON-over-HTTP was to support syncing binary data. Having every SDK use BSON by default was a requirement for that, but there has been no further progress since.

So at this point, I'm actually not sure if adding BSON support over HTTP is something worth doing. It being cheaper to decode in the core extension (even with oplog data being JSON, we don't have to scan for escape characters in those large buffers) is the only current benefit I see.

I can remove BSON support from this PR specifically to get async iterators done earlier, but we should come up with a plan for this. We should have a consistent approach to this in our SDKs, so if we don't want BSON we should remove it from the sync service HTTP endpoint and our other SDKs as well.

rkistner:
My biggest concern is that we shouldn't switch from json to bson by default in a patch release. It can make sense to do the switch for the future plan, but then we need to make sure to test the performance properly, and make it a new minor version at minimum.

rkistner (approved):

PR #838, opened 2026-01-29
fix: OP-SQLite writeLock Lock Limits

stevensJourney:
# Overview

We currently use a mix-match of two different NPM packages for Locking/Mutual exclusivity. These packages are the [async-mutex](https://www.npmjs.com/package/async-mutex) and [async-lock](https://www.npmjs.com/package/async-lock) packages.

The Async Lock package in particular supports a maximum limit for pending requests. The default for this value is set to 1000 pending requests. This means that, currently, many concurrent calls to `writeLock` could cause those calls to fail with the error

```
too many pending tasks in queue
```

Having over 1000 pending `writeLock` requests usually indicates some issue in user code, but having these errors only appear in production is also possible. For this reason, IMO, we should remove this limit. 

We also don't really have any reason to be using multiple different libraries for this functionality in our code base. This PR addresses the pending operations limit and multiple dependencies conundrum by dropping `async-lock` in favour of the (slightly more actively maintained) `async-mutex` package.

### DIFF
--- a/packages/capacitor/src/sync/CapacitorSyncImplementation.ts
+++ b/packages/capacitor/src/sync/CapacitorSyncImplementation.ts
@@ -52,7 +52,9 @@ export class CapacitorStreamingSyncImplementation extends AbstractStreamingSyncI
       });
     }
 
-    const mutex = GLOBAL_MUTEX_STORE.get(baseIdentifier)!.locks[lockOptions.type];
+    const mutexRecord = GLOBAL_MUTEX_STORE.get(baseIdentifier)!;
+    mutexRecord.tracking.add(this.instanceId);
+    const mutex = mutexRecord.locks[lockOptions.type];
 
     return mutexRunExclusive(mutex, async () => {
       if (lockOptions.signal?.aborted) {


packages/capacitor/src/sync/CapacitorSyncImplementation.ts
```diff
@@ -1,12 +1,60 @@
...
+      }
+    }
+  }
 
   async obtainLock<T>(lockOptions: LockOptions<T>): Promise<T> {
-    const identifier = `streaming-sync-${lockOptions.type}-${this.options.identifier}`;
-    return CapacitorStreamingSyncImplementation.GLOBAL_LOCK.acquire(identifier, async () => {
+    // If we don't have an identifier for some reason (should not happen), we use a shared Mutex
+    const { identifier: baseIdentifier = 'DEFAULT' } = this.options;
+    if (!GLOBAL_MUTEX_STORE.has(baseIdentifier)) {
```
simolus3, line 45: If we do have `baseIdentifier` in `GLOBAL_MUTEX_STORE`, shouldn't we also add the instance id to the existing tracking set?

stevensJourney: Nice catch! This was difficult to test without unit tests, but I have manually verified this now with the `example-capacitor` demo.

PR #839, opened 2026-02-02
fix: multiple tabs reconnect retry delay

stevensJourney:
## Overview

After [powersync-ja/powersync-js#804](https://github.com/powersync-ja/powersync-js/pull/804), a bug was observed and [reported](https://discord.com/channels/1138230179878154300/1466023819524440097/1466023819524440097): when using multi-tab support, the PowerSync client experiences a delay before reconnecting to the PowerSync service.

**Summary:** With multiple tabs, connect → disconnect → reconnect causes a visible delay (one retry cycle) before the second `connect` completes.

### How to reproduce (multi-tab)

1. Connect a PowerSync client.
2. Disconnect after connecting.
3. Reconnect to the service.

The reconnect path logs:

```
Error: Connection is busy re-opening
```

That error causes `StreamingSyncImplementation` to retry after `retryDelayMs`. The retry succeeds, but the second `connect` is delayed by that interval.

### Root cause

- After the first `connect`, the shared sync implementation connects normally.
- After `disconnect`, the internal DB inside `LockedAsyncDatabaseAdapter` is closed and must be reopened.
- The second `connect` uses the same shared `LockedAsyncDatabaseAdapter`. Using it while the internal DB is still closed triggers “Connection is busy re-opening” and starts the reopen.
- Only on the next iteration of the streaming sync loop does the adapter have a valid internal connection, so sync proceeds after the retry delay.

So the internal DB is only reopened after we discover it’s in that intermediate state, which forces the retry and the delay on the second `connect`.

## The fix

The fix is in `LockedAsyncDatabaseAdapter`:

1. **Reopen before use**  
   Previously, the pattern was:
   - Attempt to use `LockedAsyncDatabaseAdapter` in `AbstractStreamingSyncImplementation` (e.g. via `.writeLock(...)`).
   - The write lock would detect the underlying database is closed and throw `ConnectionClosed`, also triggering a reopen.
   - Subsequent `.writeLock` attempts would await the reopen process.

   The adapter now checks if the internal database is closed on the first call to `.writeLock` and reopens it immediately, so a `ConnectionClosed` error is never exposed.

2. **`reOpenInternalDB()` behavior**
   - When closing or when `reOpenOnConnectionClosed` is disabled, it no-ops (returns) instead of throwing.
   - When a reopen is already in progress, it returns the existing `databaseOpenPromise`.
   - Otherwise it starts and returns the reopen via `_reOpen()`.

3. **Tests**  
   A new test, “should re-open database immediately on reconnect”, asserts that after disconnect → reconnect, a request is sent quickly (within 500 ms, i.e. before any retry delay).

**Result:** The second `connect` no longer incurs the retry delay; callers wait for the adapter’s internal DB to be reopened before proceeding.

### DIFF
--- a/packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
+++ b/packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
@@ -323,8 +323,12 @@ export class LockedAsyncDatabaseAdapter
     if (this.databaseOpenPromise) {
       await this.databaseOpenPromise;
     } else if (!this._db) {
-      // The database is not open anymore, we might need to re-open it
-      // Note: It is safe to re-enter this method multiple times
+      /**
+       * The database is not open anymore, we might need to re-open it.
+       * Typically, _db, can be `null` if we tried to reOpen the database, but failed to succeed in re-opening.
+       * This can happen when disconnecting the client.
+       * Note: It is safe to re-enter this method multiple times.
+       */
       await this.reOpenInternalDB();
     }
 


packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
```diff
@@ -151,13 +151,15 @@ export class LockedAsyncDatabaseAdapter
    * Returns a pending operation if one is already in progress.
    */
   async reOpenInternalDB(): Promise<void> {
-    if (!this.options.reOpenOnConnectionClosed) {
-      throw new Error(`Cannot re-open underlying database, reOpenOnConnectionClosed is not enabled`);
-    }
-    if (this.databaseOpenPromise) {
+    if (this.closing || !this.options.reOpenOnConnectionClosed) {
```
stevensJourney, line 154: The changes made to `reOpenInternalDB` are to make this function re-entrant and safe to `await `in `aquireLock`s initial execution.

packages/web/src/db/adapters/LockedAsyncDatabaseAdapter.ts
```diff
@@ -317,9 +319,17 @@ export class LockedAsyncDatabaseAdapter
   protected async acquireLock(callback: () => Promise<any>, options?: { timeoutMs?: number }): Promise<any> {
     await this.waitForInitialized();
 
-    // The database is being opened in the background. Wait for it here.
+    // The database is being (re)opened in the background. Wait for it here.
     if (this.databaseOpenPromise) {
       await this.databaseOpenPromise;
+    } else if (!this._db) {
```
stevensJourney, line 325: You might ask, How can this become `null`?

We early trigger a [reOpen](https://github.com/powersync-ja/powersync-js/blob/8dee8d772bd2ac9f7198284eb3cdb7db5b564fc1/packages/web/src/worker/sync/SharedSyncImplementation.ts#L264) here when the underlaying connection has closed. This operation will clear the ref to `_db` and attempt to reOpen. The reOpen will fail if there are no other clients connected to the shared worker. The failure to reconnect is fine, we can re-attempt later, but this will leave `_db == null` until then.

Chriztiaan (approved):

Chriztiaan:
Did some manual tests on my side, specifically focusing on refreshing multiple times across multiple tabs.
Works as expected, couldn't reproduce the reported issue on the branch.

PR #840, opened 2026-02-02
Add state tracking to Sync Diagnostics Client

bean1352:
App state (credentials, dynamic schema, client parameters, query history) now lives in a dedicated local PowerSync DB (localStateDb) instead of localStorage. No app state is stored in localStorage.

- **Local DB schema** - Three tables: `query_history`, `app_settings` (credentials + dynamic_schema), `client_parameters`.
- **Two PowerSync databases** - synced database (db) for backend data and dynamic schema; local-only database (localStateDb) for app state (credentials, schema cache, client params, query history). Default context uses db; components that need local state use a nested `PowerSyncContext.Provider value={localStateDb}` so useQuery/writes hit the correct database.
- **Credentials** - Stored in local DB (`app_settings.powersync_credential`).
- **Dynamic schema** - Persisted in `app_settings.dynamic_schema`. Debounced (150 ms) during sync; immediate on connect/clear.
- **Sync overview** - Added `TanStack useQuery` for cached stats (conditional queries + invalidation on DB change). PowerSync useQuery kept for simple reactive SQL elsewhere.
- **SQL console** - Added query history drop down.
- **Rust client implementation** - Removed the option to switch to the JavaScript implementation as the Rust implementation will soon be the default.
- **Client Parameters** - These are now automatically saved as you type, I think this is better UX.

### DIFF
--- a/tools/diagnostics-app/src/app/views/client-params.tsx
+++ b/tools/diagnostics-app/src/app/views/client-params.tsx
@@ -33,30 +33,20 @@ const validateType = (type: ParameterType, value: string): string | undefined =>
   if (!value) return undefined;
   try {
     const converted = CONVERTERS[type](value);
-    if (type === 'number' && isNaN(converted as number)) {
-      return 'Invalid number';
-    }
-    if (type === 'boolean' && value !== 'true' && value !== 'false') {
-      return 'Must be "true" or "false"';
-    }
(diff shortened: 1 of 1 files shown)


bean1352:
Opened a new PR to resolve conflicts: https://github.com/powersync-ja/powersync-js/pull/843

PR #841, opened 2026-02-02
Migrate to trusted publishing

simolus3:
This migrates JS SDK packages to use trusted publishing (so that we don't have to use publishing tokens anymore).

I've made the necessary configuration changes on npm:

- [x] `@powersync/adapter-sql-js`
- [x] `@powersync/attachments`
- [x] `@powersync/attachments-storage-react-native`
- [x] `@powersync/capacitor`
- [x] `@powersync/common`
- [x] `@powersync/drizzle-driver`
- [x] `@powersync/kysely-driver`
- [x] `@powersync/node`
- [x] `@powersync/op-sqlite`
- [x] `@powersync/react`
- [x] `@powersync/react-native`
- [x] `@powersync/tanstack-react-query`
- [x] `@powersync/vue`
- [x] `@powersync/web`

To actually test the release, this PR needs to be merged so that I can trigger the workflow (which is only possible for actions that have been merged to the main branch).

### DIFF
--- a/.changeset/dull-plums-hunt.md
+++ b/.changeset/dull-plums-hunt.md
@@ -1,0 +1,18 @@
+---
+'@powersync/adapter-sql-js': patch
+'@powersync/attachments': patch
+'@powersync/attachments-storage-react-native': patch
+'@powersync/capacitor': patch
+'@powersync/common': patch
+'@powersync/drizzle-driver': patch
+'@powersync/kysely-driver': patch
+'@powersync/node': patch
+'@powersync/op-sqlite': patch
+'@powersync/react': patch
+'@powersync/react-native': patch
+'@powersync/tanstack-react-query': patch
+'@powersync/vue': patch
+'@powersync/web': patch
+---
+
+Enable trusted publishing for the PowerSync SDK.


.github/workflows/release.yml
```diff
@@ -21,16 +27,30 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Dev publish
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        run: |
+          # Previous commands from dev-packages.yml here, e.g.
+          pnpm changeset version --no-git-tag --snapshot dev
+          pnpm changeset publish --tag dev
+        env:
```
stevensJourney, line 55: I think this step, which only does local code changes and then publishes to NPM.js - might not need the `GITHUB_TOKEN`.  Do we need it for some reason?

stevensJourney: Unless if that's the same token used for publishing to NPM.js now? - I forget the specifics.

simolus3: `npm publish` (called somewhere in `pnpm changeset publish`) needs the `GITHUB_TOKEN` environment variable to generate an id token attestating that the publish happened in a specific GitHub actions workflow.

We don't need `NPM_TOKEN` anymore, and that has been removed.

PR #842, opened 2026-02-02
Fix dev publishing

simolus3:
When migrating to trusted publishing, I've accidentally removed the `pnpm build:packages:prod` step our previous script used to run. This caused us to only publish TypeScript sources instead of built packages.

This also removes the redundant `prepublishOnly` script from the capacitor package:

- For release publishes, our top-level release script calls `pnpm build:packages:prod` which does the correct build in all packages (including capacitor).
- For dev publishes, we now invoke that manually.


stevensJourney (approved):
:shipit:

PR #843, opened 2026-02-06
Add state tracking to Sync Diagnostics Client and improve UX

bean1352:
App state (credentials, dynamic schema, client parameters, query history) now lives in a dedicated local PowerSync DB (localStateDb) instead of localStorage. No app state is stored in localStorage.

- **Local DB schema** - Three tables: `query_history`, `app_settings` (credentials + dynamic_schema), `client_parameters`.
- **Two PowerSync databases** - synced database (db) for backend data and dynamic schema; local-only database (localStateDb) for app state (credentials, schema cache, client params, query history). Default context uses db; components that need local state use a nested `PowerSyncContext.Provider value={localStateDb}` so useQuery/writes hit the correct database.
- **Credentials** - Stored in local DB (`app_settings.powersync_credential`).
- **Dynamic schema** - Persisted in `app_settings.dynamic_schema`. Debounced (150 ms) during sync; immediate on connect/clear.
- **Sync overview** - Added `TanStack useQuery` for cached stats (conditional queries + invalidation on DB change). PowerSync useQuery kept for simple reactive SQL elsewhere.
- **SQL console** - Added query history drop down.
- **Rust client implementation** - Removed the option to switch to the JavaScript implementation as the Rust implementation will soon be the default.
- **Client Parameters** - These are now automatically saved as you type, I think this is better UX.

### DIFF
--- a/tools/diagnostics-app/src/components/providers/SystemProvider.tsx
+++ b/tools/diagnostics-app/src/components/providers/SystemProvider.tsx
@@ -8,23 +8,22 @@ import { localStateDb } from '@/library/powersync/LocalStateManager';
 import { queryClient } from '@/lib/queryClient';
 
 export const SystemProvider = ({ children }: { children: React.ReactNode }) => {
-  const [isLocalDbInitialized, setIsLocalDbInitialized] = useState(false);
+  const [isReady, setIsReady] = useState(false);
 
   useEffect(() => {
-    const initializeLocalStateDb = async () => {
+    const initialize = async () => {
       try {
         await localStateDb.init();
-        setIsLocalDbInitialized(true);
         await tryConnectIfCredentials();
       } catch (error) {
-        console.error('Failed to initialize local state database:', error);
-        setIsLocalDbInitialized(true);
+        console.error('Failed to initialize:', error);
       }
+      setIsReady(true);
     };
-    initializeLocalStateDb();
+    initialize();
   }, []);
 
-  if (!isLocalDbInitialized) {
+  if (!isReady) {
     return (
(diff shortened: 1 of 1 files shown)


tools/diagnostics-app/vite.config.mts
```diff
@@ -76,6 +77,7 @@ export default defineConfig({
     })
   ],
   worker: {
-    format: 'es'
+    format: 'es',
+    plugins: () => [wasm(), topLevelAwait()]
```
Chriztiaan: This was recently removed from all our demos. There were also changes for the `optimizeDeps` field.

tools/diagnostics-app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@powersync/diagnostics-app",
-  "version": "0.10.0",
+  "version": "0.9.21",
```
Chriztiaan: This change should be reverted.

tools/diagnostics-app/CHANGELOG.md
```diff
@@ -1,21 +1,5 @@
 # diagnostics-app
```
Chriztiaan, line 2: Revert

Chriztiaan:
After navigating to the `sql-console`, I can run queries and pick them from the history, but if I refresh the page I see this:
image

It doesn't run that query until I change/change back the item or manually type the query. Saw similar behaviour on the client parameters page.

I did notice the client parameters only had this issue in incognito mode.

Chriztiaan (approved):
Happy with the changes. I can't think of any serious edge cases with the tanstack query caching.

PR #848, opened 2026-02-12
(feat): [Vue/Nuxt] Added Sync Streams support and improved overall feature parity with the React SDK.

khawarizmus:
This PR should be merged after as it builds on top of: https://github.com/powersync-ja/powersync-js/pull/797

Summary:

- Added Sync Streams support, exposing `useSyncStream` composable.
- Fix `useStatus` being wrongly marked as deprecated in the SDK reference docs.
- Refactored a few composables to expose the same behaviour as the React SDK.
- Expanded test cases and added tests for Sync Streams


packages/vue/src/composables/powerSync.ts
```diff
@@ -3,7 +3,7 @@ import { App, MaybeRef, Ref, hasInjectionContext, inject, provide, ref } from 'v
 import { setupTopLevelWarningMessage } from './messages.js';
 
 // Create a unique symbol for the PowerSync context
-const PowerSyncKey = Symbol();
+export const POWERSYNC_KEY = Symbol('POWERSYNC_KEY');
```
simolus3: This doesn't seem to be used anywhere outside of this file, can we make it private again?

simolus3 (commented):
The ported sync streams hooks look good to me.

.changeset/friendly-points-ring.md
Chriztiaan, line 1: Potentially two seperate changlog entries. One for Nuxt and one for Vue. A code snippet example would be nice for both - see https://github.com/powersync-ja/powersync-js/blob/main/packages/react/CHANGELOG.md#160.

packages/nuxt/README
Chriztiaan, line 1: This file shouldn't exist, we have a README.md in main for Nuxt now.

Chriztiaan: It would help to have sync stream examples here or in the docs.

packages/vue/src/composables/useQuery.ts
```diff
@@ -45,12 +55,60 @@ export type WatchedQueryResult<T> = {
...
+    active: computed(() => runOnce.value && streamsHaveSynced.value)
+  });
+
+  const watched = useWatchedQuery<T>(query, sqlParameters, {
+    ...toValue(options),
+    active: computed(() => !runOnce.value && streamsHaveSynced.value)
+  });
+
+  const loadingState = createLoadingState<T>();
+  return {
```
Chriztiaan, line 88: Could this be a function? Then we could go back to readable branched code:
```
if (options.runQueryOnce) {
    const single = ...
    return optionMakerhelperfunctionDoodad(single)
}
else {
    const watched = ...
    return optionMakerhelperfunctionDoodad(watcged)
}

File: packages/vue/src/composables/useWatchedQuery.ts [hunk 3]
```diff
@@ -94,5 +101,6 @@ export const useWatchedQuery = <T = any>(
     isLoading,
     isFetching,
     error
+    // figure out how to return a refresh?
```
Chriztiaan: Weird todo comment. Implement or remove I guess

packages/vue/src/composables/powerSync.ts
khawarizmus: The reason I exported this is that at some point (don't remember the context), when I was building the Nuxt module, I needed it and couldn't get it. I figured out a way to go around it, but unless there is a reason to keep it private i would expose it for potentially low-level needs.

Could be useful for module authors, microfrontends, or multi-app in Nuxt.

I don't have a case for any of the above, but that's why I exposed it.

packages/nuxt/README
khawarizmus: I have actually done some modifications to it. But I don't know why the whole file is marked as changed. In any case, I will take this as an opportunity to add sync streams examples.

packages/vue/src/composables/useWatchedQuery.ts
khawarizmus: Couldn't figure out a way to do it. will remove for now

packages/vue/src/composables/useQuery.ts
khawarizmus: Will look into it

packages/vue/src/composables/powerSync.ts
simolus3: I'm generally in favor of not exposing things we don't need to expose: If we later decide to rename or remove the symbol, it's a breaking change.

I searched for `POWERSYNC_KEY` in the repository and found no usages outside of this file. If the nuxt module needs it in the future, we can always start exporting this later. If we conciously want to export it for the use-cases you mentioned, a comment pointing that out would be helpful since it's not clear to me :)

khawarizmus: I think you have a point there. Let's expose it when needed. I also can't clearly see a use case for now and as mentioned the Nuxt module didin't use it.

packages/nuxt/README
Chriztiaan: There is an .md and README file. Get rid of the latter please.

packages/vue/src/composables/useQuery.ts
Chriztiaan: As discussed, branches with composables won't work the way I thought.

PR #855, opened 2026-02-17
Raw table improvements

simolus3:
This adds support for new raw table features for version 0.4.11 of the core extension. For more context, also see the [Kotlin PR](https://github.com/powersync-ja/powersync-kotlin/pull/324).

Raw tables can now be defined in two ways:

1. With the `put` and `delete` statements, like before.
2. By passing a `tableName` describing the name of the table in the local schema, which will make the sync client infer the `put` and `delete` statement (so they're optional, but can still be used to customize them).

This also adds `RawTable.clear`, which has been added in the previous core extension update.

### DIFF
--- a/packages/common/src/db/schema/RawTable.ts
+++ b/packages/common/src/db/schema/RawTable.ts
@@ -33,7 +33,13 @@ interface RawTableTypeWithStatements {
   clear?: string;
 }
 
-interface InferredRawTableType extends Partial<RawTableTypeWithStatements>, TableOrRawTableOptions {
+/**
+ * The schema of a {@link RawTableType} in the local database.
+ *
+ * This information is optional when declaring raw tables. However, providing it allows the sync client to infer `put`
+ * and `delete` statements automatically.
+ */
+interface RawTableSchema extends TableOrRawTableOptions {
   /**
    * The actual name of the raw table in the local schema.
    *
@@ -41,8 +47,10 @@ interface InferredRawTableType extends Partial<RawTableTypeWithStatements>, Tabl
    * name. This is used to infer {@link RawTableType.put} and {@link RawTableType.delete} statements for the sync
    * client. It can also be used to auto-generate triggers forwarding writes on raw tables into the CRUD upload queue
    * (using the `powersync_create_raw_table_crud_trigger` SQL function).
+   *
+   * When absent, defaults to {@link RawTable.name}.
    */
-  tableName: string;
+  tableName?: string;
 
   /**
    * An optional filter of columns that should be synced.
@@ -51,6 +59,10 @@ interface InferredRawTableType extends Partial<RawTableTypeWithStatements>, Tabl
    * unmatched columns as local-only and will not attempt to sync them.
    */
   syncedColumns?: string[];
+}
+
+interface InferredRawTableType extends Partial<RawTableTypeWithStatements> {
+  schema: RawTableSchema;
 }
 
 /**
--- a/packages/common/src/db/schema/Schema.ts
+++ b/packages/common/src/db/schema/Schema.ts
@@ -88,11 +88,12 @@ export class Schema<S extends SchemaType = SchemaType> {
       delete: table.delete,
       clear: table.clear
     };
-    if ('tableName' in table) {
-      // We have schema options
-      serialized.table_name = table.tableName;
-      serialized.synced_columns = table.syncedColumns;
-      Object.assign(serialized, encodeTableOptions(table));
+    if ('schema' in table) {
+      // We have schema options, those are flattened into the outer JSON object for the core extension.
+      const schema = table.schema;
+      serialized.table_name = schema.tableName ?? table.name;
+      serialized.synced_columns = schema.syncedColumns;
+      Object.assign(serialized, encodeTableOptions(table.schema));
     }
 
     return serialized;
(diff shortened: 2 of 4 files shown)


packages/common/src/db/schema/Schema.ts
```diff
@@ -57,7 +58,11 @@ export class Schema<S extends SchemaType = SchemaType> {
    */
   withRawTables(tables: Record<string, RawTableType>) {
     for (const [name, rawTableDefinition] of Object.entries(tables)) {
-      this.rawTables.push(new RawTable(name, rawTableDefinition));
+      if ('name' in rawTableDefinition) {
```
stevensJourney: This is valid and correct, but, it made me do a double-take to see what was going on.. Maybe add a comment mentioning that, although `RawTableType` (accepted in the parameters of this method) doesn't declare a `name` member, `RawTable` - which extends `RawTableType` does declare a `name` field. 

Although after thinking about it, under which circumstances would this be passed in to `withRawTables`?

```typescript
const customSchema = new Schema({});
customSchema.withRawTables({
  ['synced_name_in_map']: {
    name: '[override_to_synced_name_in_map]',
    tableName: '[name_of_sqlite_table]',
    put: {
      sql: 'INSERT OR REPLACE INTO lists (id, name, _rest) VALUES (?, ?, ?)',
      params: ['Id', { Column: 'name' }, 'Rest']
    },
    delete: {
      sql: 'DELETE FROM lists WHERE id = ?',
      params: ['Id']
    }
  }
});
```

Should `RawTable` perhaps be internal?

packages/common/src/db/schema/RawTable.ts
```diff
@@ -11,7 +15,32 @@ export type RawTableType = {
...
+   */
+  tableName: string;
+
+  /**
+   * An optional filter of columns that should be synced.
+   *
+   * By default, all columns in a raw table are considered for sync. If a filter is specified, PowerSync treats
+   * unmatched columns as local-only and will not attempt to sync them.
+   */
+  syncedColumns?: string[];
```
stevensJourney, line 61: Do we usually assume that a table with local-only columns will have more/ or less local-only columns compared to synced columns?

My first thought is that most cases might have significantly more synced columns compared to local-only columns. Which could make using this setting, and specifying a large number of columns a bit tedious.

packages/node/tests/sync.test.ts
```diff
@@ -725,13 +725,83 @@ function defineSyncTests(impl: SyncClientImplementation) {
   });
 
   if (impl == SyncClientImplementation.RUST) {
-    mockSyncServiceTest('raw tables', async ({ syncService }) => {
+    mockSyncServiceTest('raw tables with inferred statements', async ({ syncService }) => {
+      const customSchema = new Schema({});
+      customSchema.withRawTables({
+        lists: {
+          tableName: 'lists'
```
stevensJourney: Is it possible for `tableName` to be optional here? It looks like there might be many cases where the object key and `tableName` would be the same.

packages/common/src/db/schema/RawTable.ts
rkistner: Discussed here: https://github.com/powersync-ja/powersync-sqlite-core/pull/164#discussion_r2803285467

Specifying synced columns better matches the JSON table structure, and are the ones that are actually used in the triggers & prepared statements.

But maybe in the future we could support both? Then you can specify either synced columns, or local/ignored columns if you only have one or two (but not both).

packages/common/src/db/schema/Schema.ts
simolus3: > Although after thinking about it, under which circumstances would this be passed in to `withRawTables`?

The one you've given would be the only one, and I agree that this confusion should probably be prevented on a type level.

> Should `RawTable` perhaps be internal?

I agree, I'll make the type internal.

packages/common/src/db/schema/RawTable.ts
simolus3: The idea is that you'd also have to specify synced columns on regular tables, and so this would be closer to that (see [this comment](https://github.com/powersync-ja/powersync-sqlite-core/pull/164#discussion_r2803285467) on the core extension). It also slightly simplifies some filters in the core extension.

> Which could make using this setting, and specifying a large number of columns a bit tedious.

I assume we could potentially add a (mutually exclusive) `localColumns` field in the future as an alternative if this turns out to be too annoying to manage.

While I would assume that most "mixed" tables have more synced than local columns, the overall count is probably not high enough for this to really matter.

stevensJourney: 👍 I think it makes sense to assess if there are any requests for this behaviour and potentially revisit in the future.

packages/node/tests/sync.test.ts
simolus3: I've considered that as well.

In the end, I wanted to avoid this because existing documentation explicitly states that the key (`lists`) doesn't have to match the name of the local table because it only affects for which tables the custom statement would be used. So changing that now would assign new meaning to that same key which might be confusing.

Additionally, we need some type-safety to express that either:

1. A table schema is given, in which case `put` and `delete` are optional.
2. Otherwise, `put` and `delete` are required.

If `tableName` was optional, `lists: {}` would be a valid definition.

I also want to make passing a schema with the table name an explicit choice since it expresses that the default inferred statements can be used (meaning that the user expects no data transformations in custom statements to be necessary). So this should be syntactically different from the developer just forgetting the `put` and `delete` statements.

simolus3: Tables can now be defined as `withRawTables(users: { schema: {} })` to avoid specifying the name twice.

Still a bit of syntactic overhead there, but that might be fine to point out that `schema` can be customized further.

PR #862, opened 2026-02-23
Added template query buttons to SQL Console in Sync Diagnostics Client

bean1352:
- Added quick-select template query buttons for common PowerSync internal tables (`ps_oplog`, `ps_crud`, `ps_buckets`, `ps_untyped`) with tooltips and docs links to both SQL consoles.
- Implemented a new feature in `SQLConsoleCore` to automatically discover and display database views. Users can now click on these views to run queries directly, enhancing the usability of the SQL console.

**Updated screenshot:**
image


kobiebotha:
I like the concept of template queries (at least for my own use, I'm often digging around in the various system tables when diagnosing issues). For new users, I'm thinking a bit of context about what those buttons are for could be useful. 

Perhaps a heading like this:

**Quick Queries**
[ps_untyped], [ps_oplog], [etc]  buttons

**Manual Query**
`SELECT * FROM ps_buckets`

bean1352:
> I like the concept of template queries (at least for my own use, I'm often digging around in the various system tables when diagnosing issues). For new users, I'm thinking a bit of context about what those buttons are for could be useful.
> 
> Perhaps a heading like this:
> 
> **Quick Queries** [ps_untyped], [ps_oplog], [etc] buttons
> 
> **Manual Query** `SELECT * FROM ps_buckets`

I like this. I added a **Quick Queries** heading and moved the documentation link under the heading instead of repeating it in each tooltip. I think we should keep the term “Queries” because it isn’t really a manual query. Clicking any of the template buttons automatically populates the input and runs it.

Chriztiaan:
Works on my side. What would be really cool is if we had a few template queries that are derived based on the schema/tables we deduce. Like a button for  `Deduced_table_1` would apply  `select * from deduced_table 1`

bean1352:
> Works on my side. What would be really cool is if we had a few template queries that are derived based on the schema/tables we deduce. Like a button for `Deduced_table_1` would apply `select * from deduced_table 1`

I really like this idea.

PR #866, opened 2026-02-24
(feat): Tanstack Query Sync Streams

Chriztiaan:
This PR adds sync stream support to the tanstack react query package, you can now define sync streams when calling useQuery and useQueries.

Examples:
```
useQuery({ queryKey: ['test'], query: 'SELECT 1', streams: [db.syncStream('a')] });

useQueries({
    queries: [
        { queryKey: ['test1'], query: 'SELECT 1', streams: [{ name: 'a' }] },
        { queryKey: ['test2'], query: 'SELECT 2', streams: [{ name: 'b' }] }
    ]
});
```

Verification:
- Added unit tests
- Manually tested these hook changes in the Supabase Todolist Sync Stream demo.

### DIFF
--- a/.changeset/nervous-taxis-explode.md
+++ b/.changeset/nervous-taxis-explode.md
@@ -1,0 +1,5 @@
+---
+'@powersync/react': minor
+---
+
+Exporting `useAllSyncStreamsHaveSynced()` member.
--- a/.changeset/tasty-meals-tap.md
+++ b/.changeset/tasty-meals-tap.md
@@ -1,0 +1,5 @@
+---
+'@powersync/tanstack-react-query': minor
+---
+
+Added sync stream support to the `useQuery()` and `useQueries()` hooks.


stevensJourney (dismissed):
The changes here look good to me

packages/tanstack-react-query/tests/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+    schema: new commonSdk.Schema({
+      lists: new commonSdk.Table({
+        name: commonSdk.column.text
+      })
+    })
+  });
+
+  onTestFinished(async () => {
+    // backport for https://github.com/powersync-ja/powersync-sqlite-core, disconnectAndClear is supposed to do this.
+    await db.execute('DELETE FROM ps_stream_subscriptions');
```
simolus3: Out of interest, can we remove this now that we've upgraded to 0.4.11? If tests still pass without it, we can also remove it from the react tests.

packages/react/src/hooks/streams.ts
```diff
@@ -65,9 +65,6 @@ export function useSyncStream(options: UseSyncStreamOptions): SyncStreamStatus |
   return status.forStream(stream) ?? null;
 }
 
-/**
- * @internal
- */
 export function useAllSyncStreamsHaveSynced(
```
simolus3, line 71: Now that the function is exposed, should we add a small documentation comment like "Returns `true` once all streams in the array have synced at least once?"

simolus3 (dismissed):

packages/tanstack-react-query/tests/utils.ts
Chriztiaan: Removed, didn't affect my tanstack query tests. Note that when I tried removing this for the react package there were issues.

simolus3: Thanks, I'll take a look at why react is failing

PR #879, opened 2026-03-03
fix: sync diagnostics row count

stevensJourney:
# Overview

The Sync diagnostics app currently displays a row count per table in the sync diagnostics page.

image

This row count is currently obtained by counting the ops per table in `ps_oplog`.

```sql
SELECT row_type as name, count() as count, sum(length(data)) as size FROM ps_oplog GROUP BY row_type
```

TLDR: The above approach works if the rows only originate from a single bucket definition in the sync config.

More details:

`ps_oplog` stores ops with the schema:

```sql
CREATE TABLE ps_oplog(
  bucket INTEGER NOT NULL,
  op_id INTEGER NOT NULL,
  row_type TEXT,
  row_id TEXT,
  key TEXT,
  data TEXT,
  hash INTEGER NOT NULL) STRICT;
```

Where the ops are related to a `bucket`.

There will be multiple `ps_oplog` entries for a row if it is referenced from multiple sync bucket definitions. E.g.

```yaml
bucket_definitions:
  # "global" sync rules will sync to all users. Note that 'global' is not a reserved word - buckets can be named anything
  global:
    data:
      # Sync all rows
      - SELECT * FROM organizations

  by_user:
    parameters:
      - SELECT request.user_id() as user_id
    data:
      - SELECT * FROM organizations where user_id = bucket.user_id
```

image

This causes the rowcount to be larger than the actual number of rows (which are deduplicated).

This updates the query logic to query the views directly for a row count.

# Implementation

The `Tables` table now contains a `Row Count` column which should correlate to the number of synced rows present.
image

An additional `Synced Count` column is added to show the previous value as recommended [here](https://github.com/powersync-ja/powersync-js/pull/879#discussion_r2877926620).
image


tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
```diff
@@ -51,10 +51,55 @@ FROM local_bucket_data local
 LEFT JOIN ps_buckets ON ps_buckets.name = local.id
 LEFT JOIN oplog_stats stats ON stats.bucket_id = ps_buckets.id`;
 
-const TABLES_QUERY = `
-SELECT row_type as name, count() as count, sum(length(data)) as size FROM ps_oplog GROUP BY row_type
```
stevensJourney, line 55: Maybe we could alternatively dedupe the ops by the `key`/`row_id` column? I can't recall enough about the specifics to judge if that's valid or not.

tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
```diff
@@ -51,10 +51,55 @@ FROM local_bucket_data local
...
+          count: countRow.count,
+          size: row.size
+        };
+      } catch (ex) {
+        console.warn(`Failed to get row count for table ${row.name}.`, ex);
+        return {
+          name: row.name,
+          // Fallback to count of ops
+          count: row.op_count,
+          size: row.size
```
rkistner: This fallback can be confusing.

If we go with this approach, can we rather explicitly indicate this as null/"unknown"?

tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
rkistner: Counting the number of unique `row_id` values per `row_type` should give you the row count. That may be a more robust approach than counting the individual tables, since it doesn't rely on updating the dynamic schema first.

It could be useful to report both counts:
1. Actual row count (unique row_id).
2. Synced row count (the old `count()`). This is not quite number of ops since they're de-duplicated per key already, but it counts the number of times the same row is synced via different buckets or different keys.

tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
stevensJourney: That's fair. I've gone with alternative approach for counting distinct row_ids.

tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
stevensJourney: I've updated the implementation to report both. I've added tooltips for some descriptions about the differences.

tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
```diff
@@ -218,7 +232,21 @@ export default function SyncDiagnosticsPage() {
 
   const tablesColumns: DataTableColumn<any>[] = [
     { field: 'name', headerName: 'Name', flex: 2 },
-    { field: 'count', headerName: 'Row Count', flex: 1, type: 'number' },
+    {
+      field: 'count',
+      headerName: 'Row Count',
+      flex: 1,
+      type: 'number',
+      tooltip: 'Number of rows synced to this database.'
```
rkistner:
```suggestion
      tooltip: 'Number of unique rows synced to this database.'
```

tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
```diff
@@ -218,7 +232,21 @@ export default function SyncDiagnosticsPage() {
...
+    },
+    {
+      field: 'synced_count',
+      headerName: 'Synced Count',
+      flex: 1,
+      type: 'number',
+      hideOnMobile: true,
+      tooltip: 'Number of row operations synced via different buckets or different replication keys.'
```
rkistner:
```suggestion
      tooltip: 'Total number of rows synced via different buckets or different replication keys.'
```

PR #881, opened 2026-03-05
(feat): Added `manageDestinationExternally` option to diff trigger creation. 

Chriztiaan:
## Problem
The first version of  the TanstackDB `on-demand` sync mode ([temp-tanstack-db#6](https://github.com/powersync-ja/temp-tanstack-db/pull/6)) repeatedly recreates diff triggers on the same destination table as the target `WHERE` clause changes. Each cycle (dispose old trigger and create new trigger) drops and recreates the destination table, this raises a potential issues with pending mutations and creates gaps that would need to be covered with locks.

## Solution
Add a `manageDestinationExternally` option to `createDiffTrigger()`. When enabled:

The SDK no longer creates or drops the destination table and the caller owns the full lifecycle (create before first trigger, drop when done). Since the SDK isn't managing the table, it shouldn't be part of the the `TriggerClaimManager` claim/release flow. Lastly, triggers created with this option use a different naming pattern
so that the periodic `cleanupResources` regex `(`/^__ps_temp_trigger_…/`)`  skips them. This avoids the cleanup process from dropping externally-managed tables it doesn't own.

--------
A `createDiffDestinationTable` helper is also exposed so callers don't need to replicate the create table destination table query themselves.

### DIFF
--- a/.changeset/slimy-jars-compete.md
+++ b/.changeset/slimy-jars-compete.md
@@ -1,0 +1,5 @@
+---
+'@powersync/diagnostics-app': patch
+---
+
+Fixed incorrect table row count when a source table is present in multiple buckets.
--- a/tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
+++ b/tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
@@ -1,19 +1,19 @@
 import { NavigationPage } from '@/components/navigation/NavigationPage';
+import { Alert, AlertDescription } from '@/components/ui/alert';
+import { Button } from '@/components/ui/button';
+import { Card, CardContent } from '@/components/ui/card';
+import { DataTable, DataTableColumn } from '@/components/ui/data-table';
+import { Dialog, DialogContent, DialogDescription, DialogHeader, DialogTitle } from '@/components/ui/dialog';
+import { Spinner } from '@/components/ui/spinner';
+import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from '@/components/ui/table';
 import { NewStreamSubscription } from '@/components/widgets/NewStreamSubscription';
 import { StreamsTable } from '@/components/widgets/StreamsTable';
-import { clearData, connector, db, sync, useSyncStatus } from '@/library/powersync/ConnectionManager';
-import { getTokenUserId, decodeTokenPayload } from '@/library/powersync/TokenConnector';
-import React, { useState } from 'react';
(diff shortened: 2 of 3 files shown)


packages/common/src/client/triggers/TriggerManager.ts
```diff
@@ -223,6 +223,13 @@ export interface CreateDiffTriggerOptions extends BaseCreateDiffTriggerOptions {
    * This table will be dropped once the trigger is removed.
    */
   destination: string;
+
+  /**
+   * When true, the diff trigger will not create or drop the destination table.
+   * The caller is responsible for ensuring the table exists with the correct
+   * schema before creating the trigger and for dropping it when no longer needed.
+   */
+  manageDestinationExternally?: boolean;
```
stevensJourney, line 238: This setting could become confusing when coupled with the `useStorage` option and the disposal logic implemented in `TriggerClaimManager`.

We should make it very clear what the intention of this setting is. Highlighting that it's only for advanced, niche use cases. 

> will not create or drop the destination table.

I haven't verified this, but this might not be 100% true, if you inspect the automatic disposal of resources in `TriggerClaimManager` - it could be that the table will be dropped if the trigger has been disposed - which is probably fine in some use cases, but we need to document that behaviour very well.

Chriztiaan: I believe these items have been covered. Better documentation and having externally managed triggers use a different naming scheme so that they aren't picked up by the trigger cleanup regex.

packages/common/src/client/triggers/TriggerManagerImpl.ts
stevensJourney: You don't need to create a claim for externally managed triggers - since we won't automatically clean those up.

packages/common/src/client/triggers/TriggerManager.ts
```diff
@@ -354,6 +367,18 @@ export interface TrackDiffOptions extends BaseCreateDiffTriggerOptions {
   throttleMs?: number;
 }
 
+/**
+ * @experimental
+ * Options for creating a diff trigger destination table with {@link TriggerManager#createDiffDestinationTable}.
+ */
+export interface CreateDiffDestinationTableOptions {
+  /** If true, the table will be created as a temporary table. Defaults to false. */
```
stevensJourney: Could we default to `temporary: true` instead of `false`? Temporary tables are a lot easier to cleanup - since they automatically get deleted when the connection closes.

Chriztiaan:
Closing this PR in favour of https://github.com/powersync-ja/powersync-js/pull/888.

PR #905, opened 2026-03-22
Fix Capacitor iOS binary SQLite param handling

swhitt:
Fixes #904.

On Capacitor iOS, PowerSync can pass `Uint8Array` payloads into:

```ts
tx.executeRaw('SELECT powersync_control(?, ?)', [op, payload])
```

The Capacitor SQLite iOS bridge doesn't seem to accept that shape directly, and sync fails with:

```text
Run: failed(message: "Error in reading buffer")
```

This normalizes typed-array payloads at the Capacitor/native SQLite boundary before they hit `run(...)`.

I think that keeps the fix in the right place and doesn't break anything unrelated. The sync layer can still use `Uint8Array`, and we only adapt it where the native bridge needs a different shape.

Not sure if a changeset makes sense here, but I think probably yes if this lands as a package fix.

### DIFF
--- a/packages/capacitor/src/adapter/sqliteParams.ts
+++ b/packages/capacitor/src/adapter/sqliteParams.ts
@@ -1,0 +1,33 @@
+export type IOSSqliteBufferParam = Record<string, number>;
+
+export type NormalizedIOSSqliteParam = unknown | IOSSqliteBufferParam;
+
+export function normalizeIOSSqliteParams(params: unknown[]): NormalizedIOSSqliteParam[] {
+  return params.map((param) => normalizeIOSSqliteParam(param));
+}
+
+function normalizeIOSSqliteParam(value: unknown): NormalizedIOSSqliteParam {
+  if (value instanceof Uint8Array) {
+    return uint8ArrayToIOSBuffer(value);
+  }
+
+  if (value instanceof ArrayBuffer) {
+    return uint8ArrayToIOSBuffer(new Uint8Array(value));
+  }
+
+  if (ArrayBuffer.isView(value)) {
+    return uint8ArrayToIOSBuffer(new Uint8Array(value.buffer, value.byteOffset, value.byteLength));
+  }
+
+  return value;
+}
+
+function uint8ArrayToIOSBuffer(array: Uint8Array): IOSSqliteBufferParam {
+  // The Capacitor SQLite iOS bridge expects BLOB params as an index-keyed object
+  // with integer values. It does not accept typed arrays directly.
+  const result: IOSSqliteBufferParam = {};
+  for (let i = 0; i < array.length; i++) {
+    result[String(i)] = array[i];
+  }
+  return result;
+}
--- a/.changeset/wobbly-octopi-juggle.md
+++ b/.changeset/wobbly-octopi-juggle.md
@@ -1,0 +1,5 @@
+---
+'@powersync/capacitor': patch
+---
+
+Normalize binary SQLite parameters for Capacitor iOS so `Uint8Array` sync payloads can be passed through `powersync_control(...)` without hitting `Error in reading buffer`.
--- a/packages/capacitor/src/adapter/CapacitorSQLiteAdapter.ts
+++ b/packages/capacitor/src/adapter/CapacitorSQLiteAdapter.ts
@@ -18,6 +18,7 @@ import {
 import { PowerSyncCore } from '../plugin/PowerSyncCore.js';
 import { messageForErrorCode } from '../plugin/PowerSyncPlugin.js';
 import { CapacitorSQLiteOpenFactoryOptions, DEFAULT_SQLITE_OPTIONS } from './CapacitorSQLiteOpenFactory.js';
+import { normalizeIOSSqliteParams } from './sqliteParams.js';
 /**
  * Monitors the execution time of a query and logs it to the performance timeline.
  */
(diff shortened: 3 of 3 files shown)


packages/capacitor/src/adapter/sqliteParams.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  if (ArrayBuffer.isView(value)) {
+    return uint8ArrayToIOSBuffer(new Uint8Array(value.buffer, value.byteOffset, value.byteLength));
+  }
+
+  return value;
+}
+
+function uint8ArrayToIOSBuffer(array: Uint8Array): IOSSqliteBufferParam {
```
simolus3, line 25: Thanks for the contribution! It's annoying that this workaround is required, but it is what it is.

This feels so absurdly inefficient that I think we should look into not using `line_binary` on Capacitor at all. Or maybe we need a hack to encode the blob as hex and then decode it with `unhex(?)` in SQLite. That's still inefficient and I think we should only do it on platforms where we need it.

Out of curiosity, does this really work on Android or did you just test iOS? Anyway, can we return a regular JSON array here (e.g. `return [...array]`)? That should be _slightly_ more efficient, but it's still not great.

stevensJourney (commented):
Thanks for the contribution!

I believe I've seen other reports of a similar issue before, but, we've not been able to reproduce this error on our end in our `example-capacitor` demo yet. 

In our example, we can see binary payloads being sent to the Capacitor Community SQLite plugin. See the screenshot below from opening the Safari dev tools

Image

These payloads work as expected on our end. 

I believe the `Error in reading buffer` comes from [here](https://github.com/capacitor-community/sqlite/blob/5735ce63ba700ebf7c40a9b29e9805c77d9b49c9/ios/Plugin/CapacitorSQLite.swift#L894). Which would most likely indicate some casting error - as pointed out in your PR description. The fact that the calls work in our demo is still concerning though - which leads me to believe there might be another factor at play.

Could you perhaps share similar logs, which cause the issue mentioned in the PR? I assume the offending payloads should have a different structure. 

One suspect might be if some `Buffer` polyfill is present in your project. If you do see a difference in the logs - could you share any polyfils of bundler config you might be using.

It would be really nice if we could get to the root cause of this issue. Thanks for your efforts!

packages/capacitor/src/adapter/sqliteParams.ts
swhitt: Re: `[...array]`, I think that'd skip the dict-based buffer path in `CapacitorSQLite.swift` entirely. The `Record<string, number>` builds the shape the Swift side expects. Haven't tested Android. Agree on `unhex(?)` longer-term.

swhitt:
No Buffer polyfills or custom bundler config here (Next.js 16, static export with a minimal config). I think it's coming from inside `@powersync/common` itself.

The binary payload reaching the Capacitor bridge isn't a plain `Uint8Array`. RSocket (bundled in `@powersync/common/dist/bundle.mjs`) wraps WebSocket data with a `Buffer` polyfill:

```js
// bundle.mjs line 10809
var buffer = bufferExports.Buffer.from(message.data);

// line 3868 - Buffer is a prototype-swapped Uint8Array
Object.setPrototypeOf(Buffer.prototype, Uint8Array.prototype);

// line 4626 - custom toJSON
Buffer.prototype.toJSON = function toJSON() {
  return { type: "Buffer", data: Array.prototype.slice.call(this._arr || this, 0) };
};
```

Chain: WebSocket `ArrayBuffer` > RSocket `Buffer.from()` > `instanceof Uint8Array` passes (`AbstractStreamingSyncImplementation.ts:986`) > `powersync_control("line_binary", payload)` > Capacitor bridge receives a `Buffer`, not a plain `Uint8Array`.

The iOS bridge (`native-bridge.js:858`) uses `postMessage()` directly, not `JSON.stringify`, so `toJSON()` isn't causing it here. I think WKWebView's internal serialization handles a prototype-swapped `Uint8Array` differently than a plain one (couldn't confirm the exact behavior from source alone).

A log in `CapacitorSQLite.swift` around line 895 would show what's actually arriving:

```swift
print("Buffer read failed: key=\(key), value=\(String(describing: mVal)), type=\(type(of: mVal))")
```

I'll try to add that locally, but probably won't get to it until later this week. Normalizing to a plain `Uint8Array` at the adapter boundary fixes it (at least for me). Re: `example-capacitor`, I'd guess it doesn't use the `line_binary` path, or Vite handles the RSocket Buffer differently.

stevensJourney:
@swhitt after some more recent testing I have been able to reproduce this issue. It seems like there might be a potentially more efficient approach. I've made an updated PR for this approach here https://github.com/powersync-ja/powersync-js/pull/933

stevensJourney:
@swhitt Thanks for the work done here. To summarise the road so far:

- An alternative approach, which theoretically should have been slightly cheaper, was implemented here https://github.com/powersync-ja/powersync-js/pull/933
- A user tested the above and reported that sync performance was very poor [here](https://github.com/powersync-ja/powersync-js/issues/904#issuecomment-4325968470) 
- We then did some benchmarks for sync performance in https://github.com/powersync-ja/powersync-js/pull/940, see [this](https://github.com/powersync-ja/powersync-js/pull/940#issuecomment-4377442627) comment for benchmarks. The work done here lead us to make NDJSON HTTP connections the default for native Capacitor apps.

I included the approach done in this PR as a test in the benchmarks of https://github.com/powersync-ja/powersync-js/pull/940. The performance was quite similar to the first alternative approach mentioned in the list. Given the vast difference between HTTP NDJSON and binary payloads - I think the only feasible solution is to change the default connection type used.

I'll close this PR to end the loop. Thanks again for raising the original issue and all the work done on your part to help close this.

PR #919, opened 2026-03-31
JourneyApps -> PowerSync and links update

cahofmeyr:


packages/adapter-sql-js/package.json
```diff
@@ -5,12 +5,12 @@
...
     "access": "public"
   },
-  "description": "A development db adapter based on SQL.js for JourneyApps PowerSync",
+  "description": "A development DB adapter based on SQL.js for PowerSync",
   "type": "module",
   "main": "dist/bundle.mjs",
   "module": "dist/bundle.mjs",
   "types": "lib/index.d.ts",
-  "author": "JOURNEYAPPS",
+  "author": "PowerSync",
```
simolus3, line 13: For those changes to get released, we'd have to add a changesets entry (`pnpm -w changeset add`) to affected packages. But we can probably also leave it as-is, they'll get picked up by some future change anyway.

simolus3 (approved):

packages/adapter-sql-js/package.json
kobiebotha: Yeah I think changes like this are non-urgent so it's fine for them to slow roll their way into the next release.

PR #923, opened 2026-04-04
fix(web): replace unload event with pagehide for Chrome 146+ compatibility

JexanJoel:
Fixes #912

## Problem
Chrome 146+ is rolling out deprecation of the `unload` event, causing a Permissions Policy violation for all users with the default `enableMultiTabs: true` flag:

This affects 1% of page loads in Chrome 146, reaching 100% by Chrome 153 (September 2026).

## Fix
Replaced both `unload` references in `PowerSyncDatabase.ts` with `pagehide`, which is the recommended alternative per the [Chrome deprecation guide](https://developer.chrome.com/docs/web-platform/deprecating-unload).

## Changes
- `packages/web/src/db/PowerSyncDatabase.ts`
  - `window.addEventListener('unload', ...)` → `window.addEventListener('pagehide', ...)`
  - `window.removeEventListener('unload', ...)` → `window.removeEventListener('pagehide', ...)`

### DIFF
--- a/packages/web/src/db/PowerSyncDatabase.ts
+++ b/packages/web/src/db/PowerSyncDatabase.ts
@@ -71,13 +71,13 @@ export interface WebEncryptionOptions {
 
 type WithWebEncryptionOptions<Base> = Base & WebEncryptionOptions;
 
-export type WebPowerSyncDatabaseOptionsWithAdapter = WithWebSyncOptions<
+export type WebPowerSyncDatabaseOptionsWithAdapter = WithWebSyncOptions
   WithWebFlags<PowerSyncDatabaseOptionsWithDBAdapter>
 >;
-export type WebPowerSyncDatabaseOptionsWithOpenFactory = WithWebSyncOptions<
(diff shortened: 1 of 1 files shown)


simolus3 (changes requested):
The migration guide also says this:

> If you're cleaning up any resources in this event, then they may have to be restored on page restore

We're not restoring anything though, so using the `pagehide` event is not appropriate.

We could probably remove those listeners without a replacement since workers are supposed to detect this anyway (https://github.com/powersync-ja/powersync-js/issues/912#issuecomment-4160486385). But we would have to test that carefully.

JexanJoel:
Thanks for the feedback! I missed that part of the migration guide. Should I close this PR and wait for consensus on the correct approach in #912, or would you prefer I update it once a direction is decided?

JexanJoel:
Closing this for now since the correct fix needs more discussion in #912 first. Happy to reopen or submit a new PR once the approach is decided. Thanks for the review!

PR #924, opened 2026-04-04
fix(web): remove unload event listener to fix Chrome 146+ deprecation

JexanJoel:
Fixes #912

Based on maintainer feedback in #912, the `unload` event listeners have been removed entirely rather than replaced with `pagehide`.

`pagehide` is not appropriate here because `close()` is a terminal action - if the tab is restored from bfcache, the database would remain closed.

Since navigator locks now reliably handle tab detection, the `unload` listeners are no longer needed.

Changes in `packages/web/src/db/PowerSyncDatabase.ts`:
- Removed `unloadListener` property
- Removed `window.addEventListener('unload', ...)` block from constructor
- Removed `window.removeEventListener('unload', ...)` block from `close()`

### DIFF
--- a/packages/web/src/db/PowerSyncDatabase.ts
+++ b/packages/web/src/db/PowerSyncDatabase.ts
@@ -38,6 +38,7 @@ import { AsyncDbAdapter } from './adapters/AsyncWebAdapter.js';
 
 export interface WebPowerSyncFlags extends WebSQLFlags {
   /**
+   * @deprecated This flag is no longer used. Navigator locks now handle tab detection automatically.
    * Externally unload open PowerSync database instances when the window closes.
    * Setting this to `true` requires calling `close` on all open PowerSyncDatabase
    * instances before the window unloads
--- a/packages/web/src/db/PowerSyncDatabase.ts
+++ b/packages/web/src/db/PowerSyncDatabase.ts
@@ -72,13 +72,13 @@ export interface WebEncryptionOptions {
 
 type WithWebEncryptionOptions<Base> = Base & WebEncryptionOptions;
 
-export type WebPowerSyncDatabaseOptionsWithAdapter = WithWebSyncOptions
+export type WebPowerSyncDatabaseOptionsWithAdapter = WithWebSyncOptions<
   WithWebFlags<PowerSyncDatabaseOptionsWithDBAdapter>
 >;
-export type WebPowerSyncDatabaseOptionsWithOpenFactory = WithWebSyncOptions
+export type WebPowerSyncDatabaseOptionsWithOpenFactory = WithWebSyncOptions<
   WithWebFlags<PowerSyncDatabaseOptionsWithOpenFactory>
 >;
-export type WebPowerSyncDatabaseOptionsWithSettings = WithWebSyncOptions
+export type WebPowerSyncDatabaseOptionsWithSettings = WithWebSyncOptions<
   WithWebFlags<WithWebEncryptionOptions<PowerSyncDatabaseOptionsWithSettings>>
 >;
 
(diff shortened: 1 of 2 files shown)


JexanJoel:
Based on the discussion in #912, I've removed the unload listeners entirely since navigator locks now handle tab detection. Happy to revert or adjust if more testing is needed before this is merged.

packages/web/src/db/PowerSyncDatabase.ts
```diff
@@ -71,13 +71,13 @@ export interface WebEncryptionOptions {
...
+export type WebPowerSyncDatabaseOptionsWithAdapter = WithWebSyncOptions
   WithWebFlags<PowerSyncDatabaseOptionsWithDBAdapter>
 >;
-export type WebPowerSyncDatabaseOptionsWithOpenFactory = WithWebSyncOptions<
+export type WebPowerSyncDatabaseOptionsWithOpenFactory = WithWebSyncOptions
   WithWebFlags<PowerSyncDatabaseOptionsWithOpenFactory>
 >;
-export type WebPowerSyncDatabaseOptionsWithSettings = WithWebSyncOptions<
+export type WebPowerSyncDatabaseOptionsWithSettings = WithWebSyncOptions
   WithWebFlags<WithWebEncryptionOptions<PowerSyncDatabaseOptionsWithSettings>>
```
simolus3: Why did these change, isn't this just a syntax error now?

packages/web/src/db/PowerSyncDatabase.ts
simolus3, line 46: We should also mark this field as deprecated since it's no longer used.

packages/web/src/db/PowerSyncDatabase.ts
```diff
@@ -263,4 +254,4 @@ export class PowerSyncDatabase extends AbstractPowerSyncDatabase {
         return new WebStreamingSyncImplementation(syncOptions);
     }
   }
-}
+}
```
simolus3: Please run this file through prettier.

simolus3 (commented):
I ran a quick manual test of these changes in the react supabase todolist demo app. Closing a tab providing the database for the sync worker still makes it retry with a different tab, so it looks like the unload listeners indeed aren't necessary anymore.

I'd like to wait for Steven to also take a look once he's back, in case he finds additional issues to look out for.

packages/web/src/db/PowerSyncDatabase.ts
JexanJoel: Sorry about that, it got mangled during editing. fixed now!

packages/web/src/db/PowerSyncDatabase.ts
JexanJoel: Good point, I've added a @deprecated JSDoc note to it.

packages/web/src/db/PowerSyncDatabase.ts
JexanJoel: Done, ran it through, should be clean now. And thanks for testing it manually, appreciate it! Happy to wait for Steven's review.

packages/web/src/db/PowerSyncDatabase.ts
stevensJourney: On my side, it looks like this is still a syntax issue. I'd be happy to approve this PR once this is updated.

stevensJourney (dismissed):

stevensJourney (approved):

simolus3:
Thanks for your contribution!

JexanJoel:
Thank you! Happy to contribute to PowerSync 🙌

PR #925, opened 2026-04-12
(fix): pass dbName instead of key to setSyncEnabled in getSyncEnabled…

johnnysedh3lllo:
## Bug

In `getSyncEnabled`, when no value exists in localStorage for the given key, the fallback call to `setSyncEnabled` incorrectly passes `key` (the already-prefixed string e.g. `syncEnabled-example.db`) instead of `dbName`. This causes `setSyncEnabled` to construct a double-prefixed key (`syncEnabled-syncEnabled-example.db`), creating a separate localStorage entry that is never read by `getSyncEnabled`.

## Fix

Pass `dbName` instead of `key` to `setSyncEnabled` in the fallback branch.

## Evidence

Two keys are created in localStorage instead of one when `getSyncEnabled` is called for the first time:

image

- `syncEnabled-example.db` (correct)
- `syncEnabled-syncEnabled-example.db` (erroneous duplicate)

### DIFF
--- a/demos/react-supabase-todolist-optional-sync/src/library/powersync/SyncMode.ts
+++ b/demos/react-supabase-todolist-optional-sync/src/library/powersync/SyncMode.ts
@@ -5,7 +5,7 @@ export function getSyncEnabled(dbName: string) {
   const value = localStorage.getItem(key);
 
   if (!value) {
-    setSyncEnabled(key, false);
+    setSyncEnabled(dbName, false);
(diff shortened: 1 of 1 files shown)


simolus3 (approved):

simolus3:
Thank you!

johnnysedh3lllo:
Thank God and you're welcome. I'm glad i could.

PR #926, opened 2026-04-13
feat: add total buckets tracking

joshuabrink:
Enhances the Sync Diagnostics page by making the server-side parameter query result limit more visible to users and providing clearer warnings as the bucket count approaches this limit.

### DIFF
--- a/tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
+++ b/tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
@@ -8,8 +8,9 @@ import { Spinner } from '@/components/ui/spinner';
 import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from '@/components/ui/table';
 import { NewStreamSubscription } from '@/components/widgets/NewStreamSubscription';
 import { StreamsTable } from '@/components/widgets/StreamsTable';
-import { formatBytes } from '@/lib/utils';
+import { cn, formatBytes } from '@/lib/utils';
 import { clearData, connector, db, sync, useSyncStatus } from '@/library/powersync/ConnectionManager';
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@/components/ui/tooltip';
 import { decodeTokenPayload, getTokenUserId } from '@/library/powersync/TokenConnector';
 import { useQueryClient, useQuery as useTanstackQuery } from '@tanstack/react-query';
 import { ChevronDown, ChevronUp, Eye, Info } from 'lucide-react';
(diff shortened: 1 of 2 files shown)


rkistner (commented):
I think it's great to track & report this, but should be clear on the wording.

Note that there are actually two separate limits, both with a default of 1000, and both reported with code PSYNC_S2305:
1. "Too many buckets". This is what the PR here measures.
2. "Too many parameter query results".

These are often the same, but is different in some important cases:
1. If the parameter query or sync stream subquery has duplicate results, those are counted individually for "parameter query results", but de-duplicated in the bucket count.
2. Queries that don't result in partitioning don't count towards "parameter query results", but do count 1 towards the number of buckets.

Another point is that the limit of 1000 is the _default_. While it's not common to change this, the text here shouldn't present that as always the case.

tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
```diff
@@ -420,12 +485,35 @@ export default function SyncDiagnosticsPage() {
...
+          {totals.buckets >= 800 && (
+            <Alert className={cn('mt-4', totals.buckets >= 900 ? 'border-destructive/50' : 'border-amber-500/50')}>
+              <Info className={cn('h-4 w-4', totals.buckets >= 900 ? 'text-destructive' : 'text-amber-600')} />
+              <AlertDescription>
+                <span className={cn('font-medium', totals.buckets >= 900 ? 'text-destructive' : 'text-amber-600')}>
+                  {totals.buckets >= 900 ? 'Critical: ' : 'Warning: '}
+                </span>
+                {totals.buckets.toLocaleString()} of {DEFAULT_SYNC_LIMIT.toLocaleString()} buckets used (PSYNC_S2305,
+                default limit). {totals.parameterized_buckets.toLocaleString()} are parameterized - at least that many
+                parameter query results on the server. Review your sync rules to reduce buckets and parameter query
```
bean1352: I think we should replace "sync rules" with either "Sync Config" or "Sync Streams/Sync Rules"

bean1352 (dismissed):

tools/diagnostics-app/src/app/views/sync-diagnostics.tsx
```diff
@@ -420,12 +484,35 @@ export default function SyncDiagnosticsPage() {
...
+              <AlertDescription>
+                <span className={cn('font-medium', totals.buckets >= 900 ? 'text-destructive' : 'text-amber-600')}>
+                  {totals.buckets >= 900 ? 'Critical: ' : 'Warning: '}
+                </span>
+                {totals.buckets.toLocaleString()} of {DEFAULT_SYNC_LIMIT.toLocaleString()} buckets used (PSYNC_S2305,
+                default limit). {totals.parameterized_buckets.toLocaleString()} are parameterized — at least that many
+                parameter query results on the server, which has a separate limit of the same default. Review your Sync
+                Config to reduce buckets and parameter query results for this user.{' '}
+                <a
+                  href="https://docs.powersync.com/sync/rules/parameter-queries"
```
benitav: Is there a better place that we can link to that also covers the "parameter query" equivalent for sync streams? Or look at this https://docs.powersync.com/debugging/troubleshooting#too-many-buckets-psync_s2305

joshuabrink: I think link both sync rules and streams in the tooltip, then include the troubleshooting link in the Alert card. Seems the most appropriate.

bean1352 (approved):

PR #930, opened 2026-04-14
Support `OPFSWriteAheadVFS`, concurrent reads

simolus3:
This adds an option to use the [new `OPFSWriteAheadVFS` VFS](https://github.com/rhashimoto/wa-sqlite/discussions/315) under `WASQLiteVFS`.

Unlike the existing VFS implementations, this one supports concurrent reads through a write-ahead log. To make use of that in the PowerSync SDK, we use a simple concurrency setup:

1. When the VFS is enabled, we spawn two dedicated workers by default instead of one. On the second worker, we open the database in read-only mode. On `WASQLiteOpenFactoryOptions`, the `additionalReaders` option can be used to increase the amount of dedicated readers.
2. In `AsyncConnectionPool`, we may use a read worker for read locks if available. Since the read pool is typically fairly small and we expect most queries to be reads, we also use the write connection for reads (giving us concurrent queries even with only one additional worker).

### DIFF
--- a/.changeset/thick-items-train.md
+++ b/.changeset/thick-items-train.md
@@ -1,0 +1,5 @@
+---
+'@powersync/web': minor
+---
+
+Add `WASQLiteVFS.OPFSWriteAheadVFS`, which also supports concurrent reads.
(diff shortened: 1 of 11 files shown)


Chriztiaan (approved):

Chriztiaan:
This looks awesome!

PR #937, opened 2026-04-28
Shared worker: Immediate reconnect on connection interrupt

simolus3:
When the sync process is managed by the Rust client as part of the SQLite extension, the client's state machine is coupled to the physical SQLite connection.

On the web, the shared sync worker use experience different connections over time as tabs are opened and closed. With the Rust sync client, attempting to handle a sync event on a connection without an active client throws. We recover from these (and all other sync errors) eventually, but it's not ideal:

1. The cause of the error is a tab being closed, but we only realize this on the next sync event (e.g. a keepalive line sent roughly every 20 seconds).
2. After each error, we wait for a configurable retry delay (5 seconds by default).

Issue 1 isn't problematic on its own (we don't really have anything to do without an event so we might as well use a broken connection), but issue 2 also means that we introduce a retry delay on the next checkpoint if it is sent before the next keepalive. This behavior is problematic and we should reconnect immediately to avoid this latency.

This PR fixes both issues: A field stores whether a connection may have been interrupted. If we get an "no iteration active" error from Rust while that field is set, we retry immediately instead of waiting for the retry delay (fixing issue 2). Also, we emit a bogus sync event as soon as the shared worker detects that the database may have been closed (fixing issue 1). This makes the Rust client much more usable on the web.

### DIFF
--- a/.changeset/purple-comics-explain.md
+++ b/.changeset/purple-comics-explain.md
@@ -1,0 +1,6 @@
+---
+'@powersync/common': patch
+'@powersync/web': patch
+---
+
+Using the Rust sync client on the web, immediately reconnect when the underlying SQLite connection is changed.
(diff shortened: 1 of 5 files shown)


stevensJourney (approved):
This looks like a good improvement to me.

PR #938, opened 2026-04-29
Remove JavaScript sync client

simolus3:
This removes the JavaScript client implementation, leaving the existing default (Rust) as the only option. That removes about 2k lines we no longer have to maintain, and should slightly decrease the bundle size for our users. 

The changes to `AbstractStreamingSyncImplementation.ts` should be uncontroversial, they just involve deleting the old client. The rest of this PRs applies simplifications to build scripts and reduces our dependency footprint. That deserves some explanations:

- To delete unused code from `@powersync/common`, I've removed TypeScript definitions for sync protocol lines and bucket storage methods used to implement the old sync client. We still want to be able to use those types in tests though. So I've added a types-only export to `@powersync/common` for that purpose.
- This also removes the `bson` dependency from our SDKs. There's nothing really wrong with that package anymore (as in, it doesn't require top-level async or polyfills since version 6), but it's evectively unused and fewer dependencies are always better. The only remaining place involving BSON was to encode the sync _request_ body for RSocket. Old versions (1.11.3 and older) of the PowerSync service  didn't accept JSON payloads there, so this drops compatibility with them. Due to an oversight on my part, the Rust client never passed a BSON encoder to that method though. So adding the Rust client implicitly dropped support for those old versions, and I'm not aware of any complaints since we've made it the default. At this point, I think the simplification of not having the BSON dependency around anymore outweighs backwards-compatibility.


packages/common/package.json
```diff
@@ -30,14 +30,18 @@
         "types": "./dist/index.d.cts",
         "require": "./dist/bundle.cjs"
       }
+    },
+    "./sync_protocol": {
```
stevensJourney: Could we make this `./internal/sync_protocol`?

stevensJourney (dismissed):
Happy with these changes :)

stevensJourney (approved):

PR #940, opened 2026-05-03
fix: [Capacitor] use HTTP NDJSON streaming as default

stevensJourney:
closes: https://github.com/powersync-ja/powersync-js/issues/904. 

The Capacitor SDK currently extends the Web SDK - where we provide some additional functionality tailored to Capacitor ecosystem, such as a SQLite driver based of Capacitor Community SQLite.

Currently: In the Web SDK, and by extension Capacitor SDK, we default to connecting to the PowerSync service via WebSockets which provide binary data payloads. These payloads are passed to our Rust Core via a SQLite query parameter. 

The Capacitor Community SQLite library has quite poor performance for handling binary query parameters. See the comment below for a basic benchmark comparison.

The results from benchmarking indicate that connecting to the PowerSync service via an NDJSON HTTP stream provides much better performance. The benefit is that we pass string JSON values to SQLite via Capacitor Community SQLite instead of binary Buffers. 

This PR updates the defaults for Capacitor to use NDJSON based HTTP connections by default when using Capacitor Community SQLite. Capacitor apps are also web apps: in order to keep things consistent with the Web SDK, we default to Web Sockets on web. This may bring more confusion than it's worth - the change is currently focussed to only affect the use-case with poor performance issues.


packages/capacitor/src/sync/CapacitorRemote.ts
```diff
@@ -0,0 +1,12 @@
+import { Capacitor } from '@capacitor/core';
+import { WebRemote } from '@powersync/web';
+
+export class CapacitorRemote extends WebRemote {
+  protected get supportsStreamingBinaryResponses(): boolean {
+    /**
+     * We'd like to avoid passing Binary buffers to SQLite when using
+     * iOS for now. This is due to inefficient binary processing.
```
simolus3: We could perhaps expand this comment a bit, e.g. by linking to https://github.com/capacitor-community/sqlite/blob/f507a1e779688ea72b9d7e8744c647f7b688c568/ios/Plugin/CapacitorSQLite.swift#L888-L912 and mentioning that buffers are serialized to JSON objects which causes the slowdown?

simolus3 (dismissed):

stevensJourney:
In order to get some more context. I did a small benchmark which compared the various connection options' sync time on iOS and Android. The test setup involved measuring the time taken to do an initial sync of 10_000 basic rows.

[Update] The initial test result posted in this comment was incorrect. This comment has been updated with valid values.

The general sync performance for Capacitor on iOS is terrible when using Buffers. Android also has a substantial performance hit when using Buffers - however the impact in iOS is orders of magnitude worse than in Android.

Legend:

`HTTP Default` refferes to the new default of NDJSON over HTTP

## Capacitor benchmark results

The differences between Android and IOS are too large to display in the same graph.

Android
image

iOS
image

OPFS seems to have issues on iOS due to opauge origins and security concerns. 

Note that the `HTTP binary indexed object` test uses the method from https://github.com/powersync-ja/powersync-js/pull/905 to pass binary payloads to SQLite.

## Kotlin SDK

For a comparison, the same data was synced using the Kotlin SDK. The time observed is in the realm of the NDJSON HTTP results seen above.

image


## Summary 

While there clearly are general issues when using Buffers, I think for now - using HTTP is the better option.

packages/capacitor/src/sync/CapacitorRemote.ts
stevensJourney: 👍 I added a more detailed note and link.

packages/capacitor/src/adapter/CapacitorSQLiteAdapter.ts
```diff
@@ -134,31 +177,35 @@ class CapacitorConnectionPool extends BaseObserver<DBAdapterListener> implements
...
+        // We cannot use `run` here for both cases.
+        (platform == 'android' && query.toLowerCase().trim().startsWith('select'))
+      ) {
         return _query(query, params);
       }
 
+      const mappedParams = mapSQLiteParameterValues({
+        platform,
+        values: params
+      });
```
simolus3, line 192: Why don't we need this for the Android query path, shouldn't we map parameters there as well?

packages/capacitor/src/adapter/CapacitorSQLiteAdapter.ts
```diff
@@ -134,31 +177,35 @@ class CapacitorConnectionPool extends BaseObserver<DBAdapterListener> implements
     const _execute = async (query: string, params: any[] = []): Promise<QueryResult> => {
       const platform = Capacitor.getPlatform();
 
-      if (db.getConnectionReadOnly()) {
+      if (
+        db.getConnectionReadOnly() ||
+        // Android: use query for SELECT and executeSet for mutations
+        // We cannot use `run` here for both cases.
+        (platform == 'android' && query.toLowerCase().trim().startsWith('select'))
```
simolus3, line 184: We don't have to fix this in this PR since the logic is unchanged, but does this break with `RETURNING` clauses?

simolus3 (approved):
I'm happy with these changes. There's nothing we can do about binary parameters being so expensive, so switching to text makes sense.

packages/capacitor/src/adapter/CapacitorSQLiteAdapter.ts
stevensJourney: The Android path above calls `_query`, where the mapping is done there instead. This avoids mapping the params multiple times.

packages/capacitor/src/adapter/CapacitorSQLiteAdapter.ts
stevensJourney: Yeah, `RETURNING` queries don't work on Android currently. It's one of the known limitations in the Readme currently.

PR #941, opened 2026-05-04
Fix `readTransaction` errors on node and op-sqlite drivers

simolus3:
Currently, we run a `BEGIN IMMEDIATE` statement to start transactions (both readonly and readwrite). Technically, we're not really supposed to use `BEGIN IMMEDIATE` for readonly transactions since that statement is documented as taking a write lock, but for readonly connections in WAL mode it's still very convenient because there it uses the current state of the database as a snapshot. This has benefits over a regular `BEGIN` because writes between the start of the readonly transaction and its first statement would not be seen (ensuring the entire `readTransaction` block forms a logical transaction).

If we use readwrite connections with `pragma query_only = TRUE` however, attempting to use `BEGIN IMMEDIATE` throws and we shouldn't do it. So, this PR:

1. Changes the Node SDK to use readonly connections instead of the `pragma query_only` workaround.
2. Changes the common package to avoid `BEGIN IMMEDIATE` for connections using the `pragma query_only = TRUE` workaround (which is the case for op-sqlite which can't open actual readonly connections).
3. Adds tests for both behaviors to the node, op-sqlite and web pacakges.

### DIFF
--- a/packages/common/src/db/DBAdapter.ts
+++ b/packages/common/src/db/DBAdapter.ts
@@ -69,7 +69,17 @@ export interface SqlExecutor {
   executeBatch: (query: string, params?: any[][]) => Promise<QueryResult>;
 }
 
-export interface LockContext extends SqlExecutor, DBGetUtils {}
+export interface LockContext extends SqlExecutor, DBGetUtils {
+  /**
+   * How the connection has been opened.
+   *
+   * `writer` indicates that the lock context is capable of writing to the database.
+   * `queryOnly` indicates that the lock context has been opened in a readwrite mode, but a `PRAGMA query_only = TRUE`
+   * disabled writes.
+   * `readOnly` indicates that the lock context has been opened by passing `SQLITE_OPEN_READONLY` to `sqlite3_open_v2`.
+   */
+  connectionType?: 'writer' | 'queryOnly' | 'readOnly';
+}
 
 /**
  * Implements {@link DBGetUtils} on a {@link SqlRunner}.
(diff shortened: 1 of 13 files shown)


packages/common/src/db/DBAdapter.ts
```diff
@@ -69,7 +69,17 @@ export interface SqlExecutor {
...
+export interface LockContext extends SqlExecutor, DBGetUtils {
+  /**
+   * How the connection has been opened.
+   *
+   * `writer` indicates that the lock context is capable of writing to the database.
+   * `queryOnly` indicates that the lock context has been opened in a readwrite mode, but a `PRAGMA query_only = TRUE`
+   * disabled writes.
+   * `readOnly` indicates that the lock context has been opened by passing `SQLITE_OPEN_READONLY` to `sqlite3_open_v2`.
+   */
+  connectionType?: 'writer' | 'queryOnly' | 'readOnly';
```
stevensJourney, line 81: Just for my understanding: why would this be optional here?

stevensJourney (approved):
These changes look good to me

packages/common/src/db/DBAdapter.ts
simolus3: Basically, we only really need to know the connection type if it's the `queryOnly` hack for op-sqlite.

Making this required would technically be a breaking change, and I want to guard against cases where `@powersync/common` is upgraded without the respective SDK being upgraded too (since we use carret dependencies from SDK packages to `@powersync/common`).

PR #942, opened 2026-05-04
Make `js-logger` a regular dependency

### LINKED ISSUE #939: js-logger should be a dependency instead of devDependency
The types are broken because `import { createBaseLogger } from '@powersync/web'` types `createBaseLogger` as `any` unless `js-logger` is installed 

`@powersync-common` has `js-logger` as a devDependency which means it's not present on the system without installing it manually so it should be a dependency or peerDependency

simolus3:
`@powersync/common` bundles `js-logger`, but the package can't be a dev dependency because the public `createBaseLogger()` function is declared to return a type from that package, in `common/index.d.cts`:

```TypeScript
import Logger, { ILogger, ILogLevel } from 'js-logger';
export { GlobalLogger, ILogHandler, ILogLevel, ILogger, ILoggerOpts } from 'js-logger';

// ...

declare function createBaseLogger(): typeof Logger;
declare function createLogger(name: string, options?: /* ... */): ILogger;
```

So if users call that method, their TypeScript compiler wouldn't be able to resolve `typeof Logger` and `ILogger` if `js-logger` is nowhere to be found. Making this a regular dependency avoids the issue.

Closes https://github.com/powersync-ja/powersync-js/issues/939.

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -59,7 +59,8 @@
     "test:exports": "attw --pack . --exclude-entrypoints internal/sync_protocol"
   },
   "dependencies": {
-    "event-iterator": "^2.0.0"
+    "event-iterator": "^2.0.0",
+    "js-logger": "catalog:"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "catalog:",
@@ -71,7 +72,6 @@
     "buffer": "^6.0.3",
     "cross-fetch": "^4.1.0",
     "estree-walker": "^3.0.3",
-    "js-logger": "catalog:",
     "magic-string": "^0.30.21",
     "rollup": "catalog:",
     "rollup-plugin-dts": "catalog:",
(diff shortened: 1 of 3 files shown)


packages/common/package.json
```diff
@@ -59,7 +59,8 @@
     "test:exports": "attw --pack . --exclude-entrypoints internal/sync_protocol"
   },
   "dependencies": {
-    "event-iterator": "^2.0.0"
+    "event-iterator": "^2.0.0",
+    "js-logger": "catalog:"
```
stevensJourney, line 63: Adding this as a regular dependency, to fix the type errors, is a valid fix. Just wanted to check if the intention is to keep bundling `js-logger` in `@powersync/common` after making this change? 
Bundling might be unnecessary in this case.

stevensJourney (approved):

packages/common/package.json
simolus3: >  Bundling might be unnecessary in this case.

That's true, but I want to avoid additional build changes (which are temporary anyway because we're about to migrate away from the package).

PR #959, opened 2026-05-13
Upgrade TypeScript, unified watcher setup

simolus3:
When working on internal SDK changes in `common`, it's convenient to be able to see what needs to be changed in the individual SDK packages consuming that. Right now though, that process requires rebuilding every package in this repository, which is very slow.

In the PowerSync service, one can simply run `pnpm watch:ts` to run tsc everytime something changes, which is much faster than having to invoke `rollup` on every package. Packages still have their complex build logic, but running TypeScript is the first step of that. In the future, I hope we get to make it the _only_ step for most packages, but we're not really there yet.

This also upgrades to TypeScript version 6. I mainly had to change a few `catch` blocks which correctly use an `unknown` type by default (but we can change that to `any`). We also a few packages opting in to strict type checks without really having save code - I've added `!` assertions to keep existing semantics. Finally, resolving types in `vue` and `nuxt` is beyond broken for reasons I don't understand (it looks like types are resolved against `@powersync/common/dist` while symbols are resolved against `@powersync/common/lib`??). A few `as any` fix this, when we split `@powersync/common` into interfaces and implementation we should be able to fix this properly.


Chriztiaan (commented):
Cool change, generally looks good to me.
Could you test a react-native demo using this change on a fresh clone/device? Just for my own sanity.

Chriztiaan (dismissed):

simolus3:
I can't get any of our RN demos to work even on `main`, they just render a black screen for me... I'll see if I can upgrade a few demos to a more recent RN version, but that seems to break a lot.

However, I've set up a new expo app, added PowerSync dependencies via the pnpmfile and got that to work. I also found a TypeScript issue in the `demos:inject` script which is fixed now.

Chriztiaan:
> I can't get any of our RN demos to work even on `main`, they just render a black screen for me... I'll see if I can upgrade a few demos to a more recent RN version, but that seems to break a lot.
> 
> However, I've set up a new expo app, added PowerSync dependencies via the pnpmfile and got that to work. I also found a TypeScript issue in the `demos:inject` script which is fixed now.

Thanks for checking, I'll do a test on my side too.

Chriztiaan:
RN demos do work on my side :)

Chriztiaan (approved):

Issue #962, opened 2026-05-13
[tauri-plugin-powersync] Re-entrant mutex deadlock between background tokio task and main thread when emitting events


mandrade2:
Package version: tauri-plugin-powersync v0.0.3  
Tauri: v2.10.3  
Platform: macOS

My desktop app becomes permanently unresponsive shortly after PowerSync connects. Sampling the process with spindump revealed a re-entrant mutex deadlock between a tokio worker thread (running PowerSync background tasks) and the main thread (handling a frontend IPC request).

I am not a tokio or Rust expert, so I traced this as far as I could. I'd be happy to test any patch or provide more diagnostics. I did use llms for this initially, then tried to understand and make sense of it. After reading more and more I do believe this is a real issue. See below for more info. If you want the whole spindump file I can provide it via a private channel.

#### Deadlock chain
```
Thread A — tokio-rt-worker (background task)  
TauriDatabaseState::new spawns two tokio::spawn tasks that watch for DB updates and sync status changes. When an event arrives, they call handle.emit(...) directly from the tokio worker:
tauri_plugin_powersync::database::TauriDatabaseState::new
  → tokio::spawn (forward_updates / forward_sync_status)
    → handle.emit("table-updates:...", &update)
      → AppManager::emit
        → listeners.emit_js(self.webview.webviews_lock().values(), &emit_args)
          → webview.eval()  // inside emit_js_filter iteration
            → WryWebviewDispatcher::eval_script
              → __psynch_mutexwait / semaphore_wait_trap  ← BLOCKED waiting for main thread
Thread B — main thread (handling frontend IPC)  
Meanwhile, the main thread receives a Tauri IPC request and tries to look up the webview:
tauri::ipc::protocol::get
  → tauri::manager::AppManager::get_webview
    → WebviewManager::webviews_lock
      → std::sync::Mutex::lock
        → __psynch_mutexwait  ← BLOCKED waiting for Thread A to release webviews_lock
```
The spindump explicitly notes:  
> (blocked by turnstile with priority 47 waiting for gitscrolls-desktop thread 0xa8efda8)

#### Why it deadlocks according to AI
1. `emit()` acquires `webviews_lock` and iterates over all webviews
2. For each webview, `emit_js_filter` calls `eval()` to inject JS into the frontend
3. `eval_script` in `tauri-runtime-wry` sends a message to the main thread and **blocks waiting for a response** (via a oneshot channel)
4. While it's blocked, it still holds `webviews_lock`
[...]

PR #965, opened 2026-05-15
Tauri: Fix binding integer values, emit events on main thread

simolus3:
This fixes two issues with the Tauri SDK:

1. Emitting events from an async Tokio task could lead to deadlocks inside of Tauri, emitting them from the main thread seems to improve things (https://github.com/powersync-ja/powersync-js/issues/962).
2. Non-negative integers are parsed as `u64` values by `serde_json`, which we didn't pick up correctly (fix originally contributed in https://github.com/powersync-ja/powersync-js/pull/955, added here so that we can ideally release this today).

### DIFF
--- a/.changeset/lovely-needles-beam.md
+++ b/.changeset/lovely-needles-beam.md
@@ -1,0 +1,5 @@
+---
+'tauri-plugin-powersync': patch
+---
+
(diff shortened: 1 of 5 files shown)


stevensJourney (approved):
Looks solid to me :)

PR #966, opened 2026-05-15
Remove `js-logger` dependency

simolus3:
As proposed in https://github.com/powersync-ja/powersync-js/discussions/948, this removes the `js-logger` dependency in favor of a custom logging interface. The idea is that:

1. The interface should be as simple as possible, and be easy to implement for users with custom logging needs.
2. This is only used for logs from the SDK, not a general-purpose logging library. We can afford making logging inside the SDK a bit less ergonomic (passing `level` as a named parameter instead of one method per level) to keep the interface smaller.
3. Outside of the message, which is always a string, we have additional fields which are well-typed and named (credits to Luc for the idea).
4. There is no global hierarchy of loggers, and no global `createBaseLogger`. If you want a PowerSync database to log, you pass a logger in. By default, we use an implementation forwarding logs to `console.log`.

For the most part, users can keep using the default option and continue to get a default out-of-the-box experience where messages with level `info` and higher are logged to the console with a `[PowerSync]` prefix. To customize this:

1. Implement the `PowerSyncLogger` interface, and pass an instance of that to `PowerSyncDatabase`. Or, if you want the default logger but with another minimum severity, use `createPowerSyncLogger`.
2. When instantiating a `WASQLiteOpenFactory` directly, one also needs to pass a logger. The web SDK would forward the one from the database by default.
3. For log messages forwarded from workers, the way to configure the minimum severity has changed. Before, we'd use the log level of the database logger. Because the logger interface is essentially a sink now and doesn't expose that, we can't do that anymore. Instead,
   - For database workers, a `logLevel` can be passed to `WASQLiteOpenFactory`, and messages with that level or higher are forwarded to the main tab.
   - For sync workers, the minimum log level can be configured as an option when constructing the database.

This is a breaking change, which is why this PR has the `v2` branch as a target. The idea is to bundle up changes which we can then release at once.


packages/common/src/utils/Logger.ts
```diff
@@ -1,47 +1,97 @@
...
- * their own logging levels and behaviors. These loggers inherit configuration
- * from the base logger by default but can override settings independently.
+ * @param options Options to configure a minimum severity of the logger or a prefix to make messages more recognizable.
  */
-export function createLogger(name: string, options: CreateLoggerOptions = {}): ILogger {
-  const logger = Logger.get(name);
-  if (options.logLevel) {
-    logger.setLevel(options.logLevel);
-  }
+export function createPowerSyncLogger(options?: Partial<CreateLoggerOptions>): PowerSyncLogger & CreateLoggerOptions {
```
rkistner: Would `createConsoleLogger` perhaps be a better description here?

packages/attachments/src/AbstractAttachmentQueue.ts
```diff
@@ -304,7 +310,10 @@ export abstract class AbstractAttachmentQueue<T extends AttachmentQueueOptions =
           return true;
         }
       }
-      this.logger.error(`UploadAttachment error for record ${JSON.stringify(record, null, 2)}`);
+      this.logger.log({
+        level: LogLevels.error,
+        message: `UploadAttachment error for record ${JSON.stringify(record, null, 2)}`
```
LucDeCaf: We can probably assign the log's `error` field to the caught error object.

packages/attachments/src/AbstractAttachmentQueue.ts
```diff
@@ -478,9 +494,9 @@ export abstract class AbstractAttachmentQueue<T extends AttachmentQueueOptions =
         }
         await this.downloadRecord(record);
       }
-      this.logger.debug('Finished downloading attachments');
-    } catch (e) {
-      this.logger.error('Downloads failed:', e);
+      this.logger.log({ level: LogLevels.debug, message: 'Finished downloading attachments' });
+    } catch (error) {
+      this.logger.log({ level: LogLevels.error, message: 'Downloads failed:', error });
```
LucDeCaf: Nit: ":" after "Downloads failed" doesn't really make sense anymore :)

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -190,7 +190,7 @@ export class AttachmentQueue {
         if (status.connected) {
           // Device came online, process attachments immediately
           this.syncStorage().catch((error) => {
-            this.logger.error('Error syncing storage on connection:', error);
+            this.logger.log({ level: LogLevels.error, message: 'Error syncing storage on connection:', error });
```
LucDeCaf: Nit: same as above

packages/common/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -545,7 +546,10 @@ export abstract class AbstractRemote {
 
       if (!res.ok || !res.body) {
         const text = await res.text();
-        this.logger.error(`Could not POST streaming to ${path} - ${res.status} - ${res.statusText}: ${text}`);
+        this.logger.log({
+          level: LogLevels.error,
+          message: `Could not POST streaming to ${path} - ${res.status} - ${res.statusText}: ${text}`
+        });
```
LucDeCaf, line 556: I think we can assign the log's `error` to the error created below it before throwing.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -387,9 +390,10 @@ The next upload iteration will be delayed.`);
               // Exit the upload loop if the sync stream is no longer connected
               break;
             }
-            this.logger.debug(
-              `Caught exception when uploading. Upload will retry after a delay. Exception: ${(ex as Error).message}`
-            );
+            this.logger.log({
+              level: LogLevels.debug,
+              message: `Caught exception when uploading. Upload will retry after a delay. Exception: ${(ex as Error).message}`
```
LucDeCaf: This is debatable because it's a debug log, but we might want to pass in the error anyways. No strong opinion.

packages/node/tests/sync.test.ts
```diff
@@ -937,11 +937,12 @@ function defineSyncTests(bson: boolean) {
 
   mockSyncServiceTest('can reconnect based on query changes', async ({ syncService }) => {
     // Test for https://discord.com/channels/1138230179878154300/1399340612435710034/1399340612435710034
-    const logger = createLogger('test', { logLevel: Logger.TRACE });
     const logMessages: string[] = [];
-    (logger as any).invoke = (level, args) => {
-      console.log(...args);
-      logMessages.push(util.format(...args));
+    const logger: PowerSyncLogger = {
+      log(_level, ...message) {
```
LucDeCaf: This looks like it's still using the initial `PowerSyncLogger.log(level: number, ...message: any[])` syntax.

packages/vue/src/composables/useSingleQuery.ts
```diff
@@ -129,9 +130,9 @@ export const useSingleQuery = <T = any>(
     const queryValue = toValue(query);
     try {
       parsedQuery = parseQuery(queryValue, toValue(sqlParameters).map(toValue));
-    } catch (e) {
-      logger.error('Failed to parse query:', e);
-      handleError(e);
+    } catch (error) {
+      logger?.log({ level: LogLevels.error, message: 'Failed to parse query:', error });
```
LucDeCaf: Nit: Same as prior

packages/vue/src/composables/useSingleQuery.ts
```diff
@@ -145,9 +146,9 @@ export const useSingleQuery = <T = any>(
       try {
         const result = await executor();
         handleResult(result);
-      } catch (e) {
-        logger.error('Failed to fetch data:', e);
-        handleError(e);
+      } catch (error) {
+        logger?.log({ level: LogLevels.error, message: 'Failed to fetch data:', error });
```
LucDeCaf: Nit: Same as prior

packages/vue/src/composables/useWatchedQuery.ts
```diff
@@ -46,9 +46,9 @@ export const useWatchedQuery = <T = any>(
     const queryValue = toValue(query);
     try {
       parsedQuery = parseQuery(queryValue, toValue(sqlParameters).map(toValue));
-    } catch (e) {
-      logger.error('Failed to parse query:', e);
-      handleError(e);
+    } catch (error) {
+      logger.log({ level: LogLevels.error, message: 'Failed to parse query:', error });
```
LucDeCaf: Nit: Same as above

LucDeCaf (commented):
I like the changes outlined here. There's just a few places which seem to be leftovers from an earlier draft, or which didn't get translated as cleanly.

packages/common/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
simolus3: I think it's good to include it 👍

packages/nuxt/tests/composables/useDiagnosticsLogger.test.ts
Chriztiaan: Dated comment I guess.

Chriztiaan (approved):

LucDeCaf (approved):

PR #970, opened 2026-05-19
fix: use text response path for React Native fetches

### LINKED ISSUE #969: [Bug]: @powersync/react-native — non-streaming HTTP fetches default to responseType: "blob", which RN 0.85 (Expo SDK 56) drops, leaving sync stuck in "downloading"
### Summary

On React Native 0.85 (Expo SDK 56 preview), every non-streaming HTTP request made by `@powersync/react-native` (e.g. `/write-checkpoint2.json`, credential fetches) emits `Invalid responseType: blob` and the response body is silently dropped. The HTTP request reports `200`, but `res.json()` never resolves on the JS side. From PowerSync's perspective the call is in-flight forever, so `SyncStatus.dataFlowStatus.downloading` stays `true` even after `hasSynced` flips and the server has acknowledged every write.

This is **separate from** the well-known streaming path — that one was already routed around via `reactNative: { textStreaming: true }` in `fetchStreamRaw`. Setting `connectionMethod: SyncStreamConnectionMethod.WEB_SOCKET` does not fix it, because the regression is in the regular `get`/`post` calls, not the streaming POST.

### Environment

- `@powersync/react-native@1.35.1`
- `@powersync/op-sqlite@0.9.8`
- `react-native@0.85.3` (Expo SDK 56 preview, new architecture enabled)
- iOS 26 simulator, Hermes

### Symptoms

- Metro logs spam `WARN Invalid responseType: blob` every ~5 seconds (matching the write-checkpoint poll cadence).
- `useStatus()` reports `connected: true, hasSynced: true, dataFlowStatus.downloading: true` indefinitely.
- Server-side: PowerSync service logs show `GET /write-checkpoint2.json → 200` repeatedly and `Updated checkpoint: N | updated: [] | removed: []`, confirming the server is healthy and the client just isn't recognising the responses.
- Local writes (`collection.insert`/`update`/`delete`) DO land in Postgres — uploads work — but the client's `downloading` flag never clears.
- Downgrading to RN 0.83 / Expo 55 makes it disappear, confirming this is an RN-version regression.

### Root cause

In `packages/react-native/src/sync/stream/ReactNativeRemote.ts` (built form: `dist/index.js`), the bundled `Fetch` class defaults to:

```ts
class Fetch {
  _nativeResponseType = 'blob';
  // ...
  __setNativeResponseType({ reactNative }) {
    if (reactNative?.textStreaming) {
      this._nativeResponseType = 'text';
      return;
    }
    this._nativeResponseType =
      reactNative?.__nativeResponseType ?? this._nativeResponseType;
  }
}
```

`fetchStreamRaw` opts into `reactNative.textStreaming = true` explicitly, so the streaming sync path is unaffected. But `AbstractRemote#get` and `#post` call `this.fetch(url, { method, headers, body })` without that option, so all non-streaming HTTP traffic falls back to `responseType: "blob"`.

React Native 0.85's `RCTNetworking.mm` / `WorkletsNetworking.mm` regressed handling of `responseType: "blob"` — they now only accept `"text"` and `"base64"` and emit `Invalid responseType: blob` while dropping the body. On RN 0.83 and earlier the blob path still worked, which is why this is a silent break for users upgrading to Expo 56.

### Suggested fix

Two equivalent shapes, in order of minimality:

1. **In `ReactNativeRemote`, inject `reactNative: { textStreaming: true }` for non-streaming calls too** — either by overriding `get`/`post`, or by overriding the `fetch` getter from `AbstractRemote` to wrap every options object. This mirrors what `fetchStreamRaw` already does and keeps the `blob` default for any external consumer of the bundled `Fetch` class.

2. **Or default `_nativeResponseType` to `"text"` directly in the `Fetch` class.** The `"text"` code path is already the one used by the streaming opt-in, so the change is just making the safe path the default. Backward compatibility is preserved via the existing `__nativeResponseType` escape hatch.

Option 2 confirmed locally as a `bun patch` against `1.35.1` for ~1h with no regressions — `dataFlowStatus.downloading` clears as expected, writes round-trip cleanly, and the Metro spam stops. Patch:

```diff
diff --git a/dist/index.js b/dist/index.js
@@ -8168,7 +8168,7 @@ function createStream(cancel) {
 
 class Fetch {
     _nativeNetworkSubscriptions = new Set();
-    _nativeResponseType = "blob";
+    _nativeResponseType = "text";
     _nativeRequestHeaders = {};
```

VIVAAN-DHAWAN:
Fixes #969

## Summary
- Wrap `ReactNativeRemote.fetch` so every React Native HTTP request passes `reactNative: { textStreaming: true }`.
- Add a small React Native vitest harness with `react-native` / `react-native-fetch-api` test aliases.
- Cover non-streaming `get()` and `post()` calls so checkpoint and credential-style requests stay on the text response path in RN 0.85.
- Add a patch changeset for `@powersync/react-native`.

## Reproduction before fix
- `pnpm --filter @powersync/react-native test --run` failed with 2 tests showing `GET` and `POST` fetch options did not include `reactNative.textStreaming`.

## Validation after fix
- `pnpm --filter @powersync/react-native test --run` -> 2 tests passed.
- `pnpm --filter @powersync/react-native build` -> passed.
- `pnpm --filter @powersync/react-native test:exports` -> no problems found.
- `pnpm exec prettier --check ...changed files...` -> all matched files use Prettier style.
- `git diff --check` -> passed.

### DIFF
--- a/.changeset/fresh-phones-fix.md
+++ b/.changeset/fresh-phones-fix.md
@@ -1,0 +1,5 @@
+---
+'@powersync/react-native': patch
+---
+
+Force React Native remote HTTP requests onto the text response path so RN 0.85 does not drop non-streaming response bodies.
--- a/packages/react-native/package.json
+++ b/packages/react-native/package.json
@@ -18,6 +18,7 @@
     "build": "tsc -b && rollup -c rollup.config.mjs",
     "build:prod": "tsc -b && rollup -c rollup.config.mjs",
     "clean": "rm -rf lib dist tsconfig.tsbuildinfo",
+    "test": "vitest --config vitest.config.ts",
     "watch": "tsc -b -w",
     "test:exports": "attw --pack ."
   },
--- a/packages/react-native/src/sync/stream/ReactNativeRemote.ts
+++ b/packages/react-native/src/sync/stream/ReactNativeRemote.ts
@@ -15,6 +15,10 @@ import { fetch } from 'react-native-fetch-api';
 
 export const STREAMING_POST_TIMEOUT_MS = 30_000;
 
+type ReactNativeFetchOptions = RequestInit & {
+  reactNative?: Record<string, unknown>;
+};
+
 /**
  * Directly imports fetch implementation from react-native-fetch-api.
  * This removes the requirement for the global `fetch` to be overridden by
@@ -36,6 +40,21 @@ export class ReactNativeRemote extends AbstractRemote {
       ...(options ?? {}),
       fetchImplementation: options?.fetchImplementation ?? new ReactNativeFetchProvider()
     });
+  }
+
+  get fetch(): FetchImplementation {
+    const fetchImplementation = super.fetch;
+    return ((input, init) => {
+      const options = (init ?? {}) as ReactNativeFetchOptions;
+
+      return fetchImplementation(input, {
+        ...options,
+        reactNative: {
+          ...(options.reactNative ?? {}),
+          textStreaming: true
+        }
+      } as RequestInit);
+    }) as FetchImplementation;
   }
 
   getUserAgent(): string {
--- a/packages/react-native/tests/mocks/react-native-fetch-api.ts
+++ b/packages/react-native/tests/mocks/react-native-fetch-api.ts
@@ -1,0 +1,3 @@
+export const fetch = async () => {
+  throw new Error('react-native-fetch-api should be injected in tests');
+};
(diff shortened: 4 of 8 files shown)


stevensJourney (approved):
Thanks for the fix!

I could reproduce the issue and confirmed this fixed it on my end.

As mentioned in the issue, we should still consider other alternatives to this implementation of `fetch` in the future. For now, I think this is the quickest/simplest win with what seems to be no negative impact.

minhnhat165:
I hope this fix can be released soon. Expo SDK 56 is now stable, and I currently can’t test because I’m running into this bug.

Chriztiaan:
Just following up @minhnhat165, this has been released in version `@powersync/react-native@1.35.2`

PR #974, opened 2026-05-21
Release mutex per attachment, persist incrementally, allow stopSync to interrupt mid-batch

### LINKED ISSUE #971: AttachmentQueue: foreground `saveFile` blocks for the full duration of an in-flight upload (shared mutex held across network I/O)
### Summary

In `@powersync/common`, `AttachmentService` exposes a single shared `Mutex` (via `withContext`) that is acquired by **both** the foreground `AttachmentQueue.saveFile` path and the background `SyncingService.processAttachments` loop. `processAttachments` keeps that mutex held across the per-attachment `remoteStorage.uploadFile` / `downloadFile` calls. As a result, a slow or stalled network request from the background sync loop blocks any foreground `saveFile` for the full duration of the upload.

### Where

- `packages/common/src/client/attachments/AttachmentService.ts` — `mutex = new Mutex()`, `withContext()` wraps the entire callback in `runExclusive`.
- `packages/common/src/client/attachments/SyncingService.ts` — `processAttachments()` calls `service.withContext(async (context) => { for (...) await this.uploadAttachment(att, context); ... })`. The `uploadAttachment` body awaits `this.remoteStorage.uploadFile(blob, attachment)`, which is user-provided `fetch`-based I/O.
- `packages/common/src/client/attachments/AttachmentQueue.ts` — `saveFile()` ultimately also goes through `service.withContext(...)` to insert the new attachment row, so it queues behind the in-flight upload.

### Observed behaviour

A reproducible hang in a Tauri 2 / WebKitGTK 2.52 desktop app on Linux:

1. User registers a new person with a photo. UI calls `queue.saveFile(...)`.
2. At the same time, `processAttachments` is mid-iteration on a previously queued attachment whose upload has stalled (flaky network, sleeping mobile NAT, etc.).
3. WebKitGTK's `fetch` has an implicit ~96 s timeout (Linux TCP keepalive defaults are 2 h, far longer); the stalled upload sits inside the mutex until WebKit finally aborts it.
4. During that window the foreground `saveFile` is pending — the user sees a silent infinite spinner; no error is thrown.
5. Once the upload eventually errors out and the mutex releases, the foreground `saveFile` proceeds and the file lands on disk normally.

### Reproduction

On Linux, applying degraded network with `tc netem`:

```
tc qdisc add dev <iface> root netem loss random 30% 25% \
    delay 800ms 400ms distribution normal reorder 25% 50% rate 100kbit
```

then triggering a foreground `saveFile` while a previous attachment is mid-upload reliably produces ~96 s delays between `[attachments] arrayBuffer ready` and `[attachments] Photo saved`. The delay matches the WebKit implicit fetch timeout almost exactly. Removing the netem rule (or running on a stable network) makes the symptom disappear.

### Hypothesis

The mutex correctly protects the in-memory `AttachmentContext` cache and the atomic state transitions in the `attachments` table, but its scope in `processAttachments` is wider than it needs to be — the per-attachment network call has no shared state to protect. Holding the lock across the network I/O means any foreground writer is gated on the slowest in-flight upload.

### Environment

- `@powersync/common` 1.51.0 (also observed on 1.53.1)
- `@powersync/tauri-plugin` 0.0.4
- Tauri 2.x, WebKitGTK 2.52 on Linux
- Supabase Storage as `RemoteStorageAdapter`

Happy to provide additional logs / a minimal repro if useful.

khawarizmus:
## Summary

Ports powersync-ja/powersync.dart#415 to the JS SDK. Closes #971.

## The issue

The `AttachmentQueue` mutex was held across the entire sync batch — including every per-attachment network upload/download. Foreground `saveFile` / `deleteFile` / watched-attachment processing all contend for the same mutex, so a slow or stalled upload hung foreground operations for the full duration. `stopSync()` also couldn't interrupt mid-batch.

## The fix

- `SyncingService.processAttachments` now runs each attachment's I/O outside the mutex and persists state per-row via a brief `withContext` re-acquire.
- `stopSync()` aborts a new `AbortController`; the sync loop polls `isActive()` between attachments and exits within one attachment's processing time.
- Watchers now observe per-attachment progress instead of one atomic commit at end of batch.

## `syncLoopMutex`

The Dart version gets serialization of sync triggers for free from `asyncMap`. JS doesn't have that — `setInterval`, `watchActiveAttachments.onDiff`, and `statusChanged` all call `syncStorage()` independently, and the old wide mutex was implicitly serializing them. Added a dedicated `syncLoopMutex` to preserve that single-batch-at-a-time semantics without re-introducing the blocking. Foreground callers don't touch this lock, so #971 still goes away.

### AI Disclosure

This PR was created with the help of Opus 4.7 and Claude Code. I’ve reviewed the resulting code changes and made sure tests are passing locally.


packages/common/src/attachments/SyncingService.ts
```diff
@@ -35,35 +35,62 @@ export class SyncingService {
...
+   *                           the loop exits early. Used by `stopSync` to interrupt
+   *                           a running batch within one attachment's processing
+   *                           time.
    */
-  async processAttachments(attachments: AttachmentRecord[], context: AttachmentContext): Promise<void> {
-    const updatedAttachments: AttachmentRecord[] = [];
+  async processAttachments(
+    attachments: AttachmentRecord[],
+    options: {
+      withContext: <T>(callback: (context: AttachmentContext) => Promise<T>) => Promise<T>;
```
stevensJourney: Since `SyncingService` already receives `AttachmentService` in its constructor, do we need to pass `withContext` into `processAttachments` as an option? It seems like `SyncingService` could call `this.attachmentService.withContext(...)` directly when it needs an `AttachmentContext`.

packages/common/src/attachments/SyncingService.ts
```diff
@@ -35,35 +35,58 @@ export class SyncingService {
...
+   *                           the loop exits early. Used by `stopSync` to interrupt
+   *                           a running batch within one attachment's processing
+   *                           time.
    */
-  async processAttachments(attachments: AttachmentRecord[], context: AttachmentContext): Promise<void> {
-    const updatedAttachments: AttachmentRecord[] = [];
+  async processAttachments(
+    attachments: AttachmentRecord[],
+    options?: {
+      isActive?: () => boolean;
```
simolus3: A callback is kind of an odd way to express this IMO, couldn't we pass the `AbortSignal` instead?

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -325,24 +343,41 @@ export class AttachmentQueue {
    *
    * This is called automatically at regular intervals when sync is started,
    * but can also be called manually to trigger an immediate sync.
+   *
+   * Concurrent invocations are serialized via `syncLoopMutex`.
    */
   async syncStorage(): Promise<void> {
-    await this.attachmentService.withContext(async (ctx) => {
-      const activeAttachments = await ctx.getActiveAttachments();
+    await this.syncLoopMutex.runExclusive(async () => {
```
simolus3: Note that you can pass an abort signal to `runExclusive` too, which may be helpful in case there's an aborted call blocking on `syncStorage`.

packages/common/src/attachments/SyncingService.ts
khawarizmus: Fixed this by using the internal `attachmentService`'s context

packages/common/src/attachments/SyncingService.ts
khawarizmus: Fixed this by passing teh signal directly

packages/common/src/attachments/AttachmentQueue.ts
khawarizmus: This is a good catch, as it could have potentially blocked a pending batch. I have included a test for that case and fixed the implementation to also pass the signal to `runExclusive`

Chriztiaan (approved):

PR #975, opened 2026-05-24
feat: support mediaType in watched attachments

### LINKED ISSUE #922: watchAttachments: support mediaType in WatchedAttachmentItem
## Problem

When an attachment is saved via `saveFile()`, the `mediaType` is stored in the local attachment record. However, when another device syncs and the `watchAttachments` callback creates a new `QUEUED_DOWNLOAD` record, `mediaType` is never set — only `id`, `filename`, `state`, `metaData`, and `timestamp` are populated.

This means `attachment.mediaType` is always `null` on devices that download attachments from remote storage, even though the uploading device had the correct MIME type.

### Relevant code (in `AttachmentQueue`)

```js
// When watchAttachments detects a new attachment:
attachmentUpdates.push({
  id: watchedAttachment.id,
  filename,
  state: AttachmentState.QUEUED_DOWNLOAD,
  hasSynced: false,
  metaData: watchedAttachment.metaData,
  timestamp: new Date().getTime()
  // ← no mediaType
});
```

### Current `WatchedAttachmentItem` shape

The items passed to `onUpdate()` only support:
- `id`
- `fileExtension`
- `filename`
- `metaData`

There's no `mediaType` field.

## Proposal

Add an optional `mediaType` field to `WatchedAttachmentItem`:

```ts
interface WatchedAttachmentItem {
  id: string;
  fileExtension?: string;
  filename?: string;
  metaData?: string;
  mediaType?: string; // ← new
}
```

When present, pass it through to the attachment record created for `QUEUED_DOWNLOAD`:

```js
attachmentUpdates.push({
  id: watchedAttachment.id,
  filename,
  state: AttachmentState.QUEUED_DOWNLOAD,
  hasSynced: false,
  metaData: watchedAttachment.metaData,
  mediaType: watchedAttachment.mediaType, // ← new
  timestamp: new Date().getTime()
});
```

This would allow users to derive the MIME type from their data model (e.g., from the file extension in the attachment ID, or a dedicated column) and pass it through `watchAttachments`, so downloaded attachments have the correct `mediaType`.

## Workaround

Currently working around this by inferring MIME type from the file extension embedded in the attachment ID (e.g., `uuid.webp` → `image/webp`).

## Environment

- `@powersync/web` v1.37.0

VIVAAN-DHAWAN:
Closes #922

## Summary
- Add optional mediaType to WatchedAttachmentItem for filename and fileExtension variants.
- Preserve watched mediaType when creating QUEUED_DOWNLOAD attachment records.
- Document the new field and add a changeset for @powersync/common.

## Verification
- Reproduced the bug with the new regression: mediaType was null before the fix.
- pnpm --filter @powersync/common build
- pnpm --filter @powersync/node build
- pnpm exec vitest run packages/node/tests/attachments.test.ts
- pnpm exec prettier --check .changeset/fuzzy-cameras-sip.md packages/common/src/attachments/AttachmentQueue.ts packages/common/src/attachments/WatchedAttachmentItem.ts packages/node/tests/attachments.test.ts
- git diff --check


.changeset/fuzzy-cameras-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/common': patch
```
Chriztiaan: This change affects the public API surface of WatchedAttachmentItem (even though non-breaking). I think this should be a minor change instead.

Chriztiaan (commented):
Happy with the changes. Just update the changeset entry's level please.

LucDeCaf (approved):

PR #976, opened 2026-05-25
chore: capacitor integration tests

stevensJourney:
The work here is inspired and based off the Tauri integration tests added by @simolus3 . [Reference](https://powersync.com/blog/tauri-integration-tests)

This PR adds integration tests for the Capacitor SDK. 

For context, most Capacitor plugins include an `example-app` folder which is a basic Capacitor App used to verify the plugin compiles correctly for Android and iOS.

This PR uses the `example-app` as a code runner which loads the Vitest Browser-mode server's URL as the source of the Capacitor web content. 

We use a small trick where the `example-app`'s Capacitor config is a TypeScript (could also be JavaScript) file which sets the app's web content source to the value of an environment variable. This means that the app can be started with a dynamic web contents source.

The Vitest browser mode integration then builds the Capacitor app and starts it by providing the appropriate URL for the Vitest test suite. 

This PR ports the React Native tests from `tools/powersynctests` for Capacitor. The test suite essentially only handles basic query operations and Capacitor specific logic - like ensuring concurrency. AI usage disclaimer - the tests were ported using Codex GPT 5.5.

## Bug fix

These unit tests actually exposed a bug in the current code! The `executeBatch` method was broken on iOS. This adds a fix.


.github/workflows/test-simulators.yaml
```diff
@@ -106,18 +81,84 @@ jobs:
...
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: '.nvmrc'
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Build
+        run: pnpm build:packages
```
simolus3: nit: We could use `pnpm run -r --filter "@powersync/capacitor..." build` to only build used packages, but I doubt that saves much compared to the time it takes to start Gradle / emulators.

.github/actions/setup-android-emulator/action.yaml
```diff
@@ -0,0 +1,70 @@
...
+      uses: actions/cache@v3
+      id: avd-cache
+      with:
+        path: |
+          ~/.android/avd/*
+          ~/.android/adb*
+        key: ${{ inputs.cache-key }}
+
+    - name: Set up JDK
+      uses: actions/setup-java@v3
```
simolus3: I see this was copied over, but still:

```suggestion
      uses: actions/setup-java@v5
```

packages/capacitor/example-app/android/app/src/androidTest/java/com/getcapacitor/myapp/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,26 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * Instrumented test, which will execute on an Android device.
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+@RunWith(AndroidJUnit4.class)
+public class ExampleInstrumentedTest {
```
simolus3, line 17: I assume this is unused since we don't use AGP connected tests, right?

packages/capacitor/example-app/android/app/src/test/java/com/getcapacitor/myapp/ExampleUnitTest.java
```diff
@@ -0,0 +1,18 @@
+package com.getcapacitor.myapp;
```
simolus3, line 1: I guess the same here, but now I realize this is from a template so we could also keep them

packages/capacitor/tests/setup/capacitor.ts
```diff
@@ -0,0 +1,19 @@
+async function installCapacitorInTestFrame() {
+  if (window.top == null || window.top === window) {
+    return;
+  }
+
+  const root = window.top as Window & typeof globalThis;
+  const current = window as Window & typeof globalThis;
+
+  // await waitForCapacitorBridge(root);
```
simolus3: I assume this reliably adds the globals before we run? In that case, I suppose we can drop the `async` and remove this comment.

simolus3 (dismissed):
I ran the integration tests in an iOS simulator, it's definitely cool to see the vitest UI there 👍

Perhaps also update `package/capacitor/example-app/README.md` to mention that the app serves as an integration test host?

packages/capacitor/example-app/android/app/src/androidTest/java/com/getcapacitor/myapp/ExampleInstrumentedTest.java
stevensJourney: Jip, this was a boiler plate template file. I think it's easier to just keep them.

packages/capacitor/example-app/android/app/src/test/java/com/getcapacitor/myapp/ExampleUnitTest.java
stevensJourney: Jip, this was a boiler plate template file. I think it's easier to just keep them.

packages/capacitor/tests/setup/capacitor.ts
stevensJourney: Nice catch. Indeed. I moved the comment to the function declaration and made it a sync function.

simolus3 (approved):

PR #977, opened 2026-05-25
Align crud uploads with Dart implementation

simolus3:
The Dart SDK invokes the CRUD upload method of the configured backend connector at least once per `connect()` call, even if we can't connect to the streaming sync endpoint. We want the same behavior in the JS SDK, and this PR adds that.

To make the implementation easier to reason about, this also aligns the implementation with the Dart SDK:

- We call `crudUploadLoop` in `connect()`, it takes a scoped abort signal now to abort along with the corresponding `streamingSync()` call after a disconnect.
- Instead of invoking uploads directly in an event handler, we create something resembling a stream controller to dispatch and throttle events on the consuming side.

Note that there are still some fishy things going on here, e.g. we break out of the `_uploadAllCrud()` loop if there's an error and we're not connected. The Dart SDK does the same, and this PR is still an improvement, but we should probably spec out how we actually want the upload flow to behave before applying this to other SDKs.

### DIFF
--- a/packages/common/src/utils/async.ts
+++ b/packages/common/src/utils/async.ts
@@ -18,33 +18,3 @@ export function throttleTrailing(func: () => void, wait: number) {
     }
   };
 }
-
-/**
- * Throttle a function to be called at most once every "wait" milliseconds,
- * on the leading and trailing edge.
- *
- * Roughly equivalent to lodash/throttle with {leading: true, trailing: true}
- */
-export function throttleLeadingTrailing(func: () => void, wait: number) {
-  let timeoutId: ReturnType<typeof setTimeout> | null = null;
-  let lastCallTime: number = 0;
-
-  const invokeFunction = () => {
-    func();
-    lastCallTime = Date.now();
-    timeoutId = null;
-  };
-
-  return function () {
-    const now = Date.now();
-    const timeToWait = wait - (now - lastCallTime);
-
-    if (timeToWait <= 0) {
-      // Leading edge: Call the function immediately if enough time has passed
-      invokeFunction();
-    } else if (!timeoutId) {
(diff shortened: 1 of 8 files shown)


packages/common/src/utils/async.ts
```diff
@@ -19,32 +19,57 @@ export function throttleTrailing(func: () => void, wait: number) {
...
-      // Set a timeout for the trailing edge if not already set
-      timeoutId = setTimeout(invokeFunction, timeToWait);
+export function asyncNotifier(): AsyncNotifier {
+  let waitingConsumers: (() => void)[] = [];
+  let hasPendingNotification = false;
+
+  return {
+    notify() {
+      if (waitingConsumers.length > 0) {
+        waitingConsumers.splice(0, 1)[0]();
```
stevensJourney: It looks like `waitingConsumers` could contain multiple waiters, but `notify()` only resolves the first one. Is that intentional queue semantics? The interface/docs read like a notification should wake all current waiters. If we only expect one consumer, maybe this should assert/document that; otherwise `notify()` should drain and resolve all waiting consumers.

simolus3: That is a good point. I originally had an assert here to only support a single consumer, but an AI code review pointed out that there might be multiple listeners from distinct `connect` calls (since `crudUploadLoop` isn't awaited).

I agree that only supporting a single consumer is cleaner though, we can use structured concurrency to ensure we don't have concurrent `crudUploadLoop` calls.

stevensJourney (approved):

simolus3:
It looks like the Tauri integration tests are broken (timeouts without a message). They pass locally though, so I'll merge this and investigate those failures on another branch.

PR #985, opened 2026-06-09
feat:  expose AttachmentQueue's AttachmentContext

stevensJourney:
Some `AttachmentQueue` users have a use case for custom `saveFile` methods. The `attachmentService` exposes a context to mutate `attachment` records in a safe manner. Unfortunately, the `attachmentService` is  currently marked as `private` which means extending classes cannot access it. 

The `AttachmentsService` contains functionality to watch active attachments and obtain an `AttachmentContext` via a mutex. Users won't likely need to watch the attachment state, but they could need an `AttachmentContext` - which exposes methods for reading and manipulating `AttachmentRecord`s.

This makes `AttachmentContext` a public API and adds a small method to `AttachmentQueue` to access an `AttachmentContext`.

An example use case for requiring access is available at https://github.com/powersync-ja/powersync-js/pull/983

### DIFF
--- a/.changeset/rude-schools-mix.md
+++ b/.changeset/rude-schools-mix.md
@@ -1,0 +1,5 @@
+---
+'@powersync/common': patch
+---
+
+[Attachments] Make `AttachmentQueue.attachmentService` available for classes extending the base class. This allows for custom saveFile implementations.
--- a/.changeset/rude-schools-mix.md
+++ b/.changeset/rude-schools-mix.md
@@ -1,5 +1,5 @@
 ---
-'@powersync/common': patch
+'@powersync/common': minor
 ---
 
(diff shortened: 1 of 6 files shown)


simolus3 (commented):
We need to run `pnpm exec api-extractor run --local` to update the API snapshot. Also `AttachmentService` should be marked as `@alpha` instead of `@internal` now since this makes it part of the public API.

stevensJourney:
> We need to run `pnpm exec api-extractor run --local` to update the API snapshot. Also `AttachmentService` should be marked as `@alpha` instead of `@internal` now since this makes it part of the public API.

It's nice to see the API Extractor in action! After properly examining the current state of the various components. I updated this PR to not expose `AttachmentService` directly, but rather add a small method to `AttachmentQueue` to expose a now public `AttachmentContext`.

simolus3 (approved):

PR #986, opened 2026-06-09
Node: Remove CommonJS build

simolus3:
In a `package.json`'s `export` fields, Node.js supports the `module-sync` type, which

> matches no matter the package is loaded via `import`, `import()` or `require()`. The format is expected to be ES modules that does not contain top-level await in its module graph - if it does, `ERR_REQUIRE_ASYNC_MODULE` will be thrown when the module is `require()`-ed.

Luckily, none of our transitive imports contain a top-level await (if they did, we would have a lot of trouble with other bundlers too). So, we can remove the CommonJS build without dropping support for CommonJS! This allows us to ship `tsc` output without any additional bundlers or build logic. I've verified that things are still usable from CommonJS files by running this `test.cjs` file in `demos/example-node`:

```JavaScript
const { PowerSyncDatabase, Schema } = require('@powersync/node');

const db = new PowerSyncDatabase({ schema: new Schema({}), database: { dbFilename: 'test.db' } });

async function useDatabase() {
  await db.get('SELECT 1');
  await db.close();
}

useDatabase();
```

Getting rid of build tools and source transformations reduces complexity and makes it easier to reason about shipped code, so I strongly believe we should do this wherever possible. Node is a first step, but I think we can get away with that on almost all of our main packages:

1. We can assume modern web toolchains support ESM modules directly.
2. For node, this trick allows us to remove CommonJS modules.
3. React Native supports ESM modules since 0.79, so the only reason we still need a bundling step in `@powersync/common` is to rewrite `Symbol.asyncIterator` use in the `event-iterator` package. IMO, we should vendor that package into our sources and remove build steps there too.

I want to do that on the `v2` branch, but fixing Node is a relatively low-risk item we can do first.


Chriztiaan:
For
> 1. We can assume modern web toolchains support ESM modules directly.

and

> React Native supports ESM modules since 0.79

It would be good to know if this means we can drop UDM support React Native Web, not 100% sure whether this holds for Angular either where we have to copy worker assets over.

Chriztiaan (approved):

simolus3:
I'll keep those in mind for tests before removing any additional CommonJS builds 👍

PR #987, opened 2026-06-09
Refactor open and sync option types

simolus3:
This refactors types passed to different `PowerSyncDatabase` constructors and to `PowerSyncDatabase.connect`. Above all else, the goal of the refactoring is simplification: We have way too many different option types which it very hard to understand where options are actually coming from.

This PR changes three things:

## 1. Remove PowerSync open factories

In addition to `SQLOpenFactory`, which opens a `DBAdapter`, we also had deprecated classes to open `PowerSyncDatabase` instances. We should just remove them.

## 2. Refactor sync options

Compared to all our other SDKs, the complexity of sync options types in the JavaScript SDK is absurd. There's `BaseConnectionOptions` (marked as `@internal` despite being the one interface containing the options users are actually supposed to set), `AdditionalConnectionOptions`, `RequiredAdditionalConnectionOptions` (which is not the same as `Required<AdditionalConnectionOptions>`), `InternalConnectionOptions` and finally `PowerSyncConnectionOptions`. And then there are probably a bunch more for the web that I already forgot about.

Compare this to the Dart SDK, which has a single public class for this: [`SyncOptions`](https://pub.dev/documentation/powersync/latest/powersync/SyncOptions-class.html). So, this cleans up options a bit to be more explicit:

1. The only type users care about is `SyncOptions`, which is mirrored from the Dart SDK.
2. An internal function can apply defaults to turn `SyncOptions` into `type ResolvedSyncOptions = Required<SyncOptions>`.
3. Then there's a single additional interface representing named parameters on the constructor for streaming sync implementations (the existing `AbstractStreamingSyncImplementationOptions`). `serializedSchema` was moved to this.

Additionally, some SDKs supported additional sync options. For example, the Node SDK supported a custom `Dispatcher` for undici. In an effort to make the option type consistent across SDKs, setting SDK-specific sync options now happens when creating the database instead of when calling `connect`. For Node, one can set `remoteOptions` on the database constructor which allows installing a custom dispatcher.

## 3. Refactor open options

Different JavaScript SDKs support slightly different options to open databases. While this is unfortunate, it's also unavoidable given how different the JavaScript SDKs are. Even things that feel like they should be supported on all SDKs, like wrapping a `DBAdapter` into a `PowerSyncDatabase` are not supported on the Tauri SDK.

So I think having option types for each SDK is the way to go. However, there's still a lot we can improve here:

1. All three existing ways to open databases were called `database` in the old options type. This worsens the quality of auto-complete. We now have three fields out of which exactly one must be set: `opened` to use an existing `DBAdapter`, `factory` to use a custom `SQLOpenFactory` or `database` with common open options (filename, location).
2. On some SDKs, notably the web, additional options are available for the `database` option. However, these are _only_ available on the `database` option. For example, it's no longer possible to combine `encryptionKey` with an existing `DBAdapter` - a check that previously required a runtime warning is now prevented on a type level.
3. SDKs have additional options, which are now part of the top-level options field. Node allows customizing the undici dispatcher, the web has options like `broadcastLogs` (which was previously part of `WebSQLFlags` despite not actually being used for db workers).

For the web SDK, I've merged different options previously used to open database workers:

1. Both `WASqliteOpenFactory` and `PowerSyncDatabase` with the `database` key now take the exact same option: `WebSQLOpenOptions`. When constructing a `WASqliteOpenFactory` directly, one also needs to pass a `PowerSyncLogger` (when opened through `PowerSyncDatabase`, it inherits the logger from the database). All other types (like `WebSQLFlags`) have been removed and merged into this.
2. There are additional web-specific options: `broadcastLogs` (to configure whether shared worker should emit broadcast logs) and `WebSyncOptions` (which can be used to configure a custom sync worker).

### Web change summary

The most common case of opening web database works without changes: `new PowerSyncDatabase({database: {dbFilename: 'test.db'}})`. Because options are shared, it's also possible to include a VFS in there without requiring an explicit WA-sqlite open factory: `new PowerSyncDatabase({database: {dbFilename: ..., vfs: WASQLiteVFS.OPFSWriteAheadVFS}})`. Options that previously required a `flags` object have also been merged into that (`database: {dbFilename: ..., useWebWorker: false}`).

When constructing a `WASQLiteOpenFactoryOptions`, one would use the same options + a logger: `new WASQLiteOpenFactory({logger: createConsoleLogger('database'), open: { dbFilename: ..., vfs: WASQLiteVFS.OPFSWriteAheadVFS }})`.


.changeset/afraid-carpets-sin.md
```diff
@@ -0,0 +1,27 @@
...
+- Refactor open options: Each SDK has its own option type (but most use `PowerSyncDatabaseOptions`).
+  Using a custom open factory or an existing `DBAdapter` instance now requires the `factory` and
+  `opened` key, respectively (previously these options shadowed the `database` field).
+  On the web, all options that previously required a `WASQLiteOpenFactory` are now available on
+  `database` too.
+  Also on the web, all database open options are represented via `WebSQLOpenOptions`. Additional
+  options related to the sync process have moved to the `WebSpecificOptions` type to separate those
+  concerns better.
+- Aligning the implementation with the public documentation, the default connection method has changed
+  from WebSockets to HTTP.
```
simolus3, line 27: I want to revert this default for React Native in a follow-up PR, since WebSockets is a better default for that platform. There's another v2 proposal to make WebSockets opt-in for other SDKs which requires an additional refactor, I'll use that opportunity to clean this up.

Chriztiaan, line 27: This might be something for `v2` in general, but I think a small upgrade guide would be valuable. Thinking of things like the database creation options being restructured/moved. Maybe just examples?

Chriztiaan (approved):
Generally I am happy with the changes and the simplification.

.changeset/afraid-carpets-sin.md
simolus3: We will definitely have a lot of documentation updates to do for this :D But to give us a good overview for later, I've added a few common examples at the end of this changeset entry.

Chriztiaan (approved):

.changeset/afraid-carpets-sin.md
Chriztiaan: Awesome!

PR #991, opened 2026-06-10
Fix integration test browser issues with ArrayBuffers

mandrade2:
Hey there! Recently read [this](https://powersync.com/blog/tauri-integration-tests) great post from @simolus3 on vitest integration tests for Tauri. As soon as I read it I tried it out. My app makes extensive usage of `tauri-plugin-fs` so I started testing and noticed that reading files was broken on the integration tests. 

When Vitest runs Tauri integration tests inside an iframe, `tauri-plugin-fs` and other Tauri IPC internals use `instanceof ArrayBuffer` to detect binary responses from Rust commands.

However, the `ArrayBuffer` returned by `invoke()` is created in the **top-level window's realm** (where Tauri injects its globals), while the test code runs in the **iframe's realm**. Because of JavaScript's cross-realm semantics, `instanceof ArrayBuffer` returns `false` when the constructor and the instance come from different realms. This breaks any binary data transfer — for example, `readFile` from `@tauri-apps/plugin-fs` will silently misbehave or throw because the returned `ArrayBuffer` fails the `instanceof` check.

I've attached a fix in this PR. It's just a small extension of the iframe workaround. Also searched for more instanceof usages across Tauri plugins and did not find any other. So it seems we'll not hit an endless stream of cases like this (hope so)

### DIFF
--- a/packages/tauri/tests/integration.test.ts
+++ b/packages/tauri/tests/integration.test.ts
@@ -15,6 +15,10 @@ async function installTauriInTestFrameHack() {
 
   current.__TAURI_INTERNALS__ = root.__TAURI_INTERNALS__;
   current.__TAURI_EVENT_PLUGIN_INTERNALS__ = root.__TAURI_EVENT_PLUGIN_INTERNALS__;
+
+  // Fix cross-realm instanceof ArrayBuffer for tauri-plugin-fs binary responses.
+  current.ArrayBuffer = root.ArrayBuffer;
+
 }
 
 const schema = new Schema({
--- a/packages/tauri/tests/integration.test.ts
+++ b/packages/tauri/tests/integration.test.ts
@@ -16,7 +16,10 @@ async function installTauriInTestFrameHack() {
   current.__TAURI_INTERNALS__ = root.__TAURI_INTERNALS__;
   current.__TAURI_EVENT_PLUGIN_INTERNALS__ = root.__TAURI_EVENT_PLUGIN_INTERNALS__;
 
-  // Fix cross-realm instanceof ArrayBuffer for tauri-plugin-fs binary responses.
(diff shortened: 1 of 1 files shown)


packages/tauri/tests/integration.test.ts
```diff
@@ -15,6 +15,10 @@ async function installTauriInTestFrameHack() {
 
   current.__TAURI_INTERNALS__ = root.__TAURI_INTERNALS__;
   current.__TAURI_EVENT_PLUGIN_INTERNALS__ = root.__TAURI_EVENT_PLUGIN_INTERNALS__;
+
+  // Fix cross-realm instanceof ArrayBuffer for tauri-plugin-fs binary responses.
```
simolus3: Could you expand this comment a bit? Maybe say something like this:
```suggestion
  // When binary responses are used, Tauri injects top-level code matching against
  // ArrayBuffers. Ensure this iframe realm uses the same class for compatibility. We
  // don't currently use binary responses in the PowerSync plugin, but it serves as an
  // example for general integration tests with Tauri.
```

simolus3 (commented):
Thanks for the contribution! This change makes sense to me, I'll also update the post to include this snippet. I think we should point out that we don't actually rely on this for our tests (I'm still happy to include it as a good practice though).

simolus3 (approved):
Thanks! We've also updated the blog post by the way :)

mandrade2:
My project uses Yjs and I wrote some integration tests. Had to also add `Uint8Array` in the same way we did so for `ArrayBuffer`. Don't think we should have this upstream but leaving it written down here in case anyone hits the same issue. Overall i think it's a small price to pay to be able to have integration tests like these. Have been very valuable to me!

PR #995, opened 2026-06-16
Simplify `DBAdapter` interface and query result types

simolus3:
`DBAdapter` is the interface we implement to give a PowerSync SDK access to SQLite. It's a public interface, which means that users can implement it too if they want a custom SDK. However, it's also very annoying to implement, since one needs to write all `get()`, `getOptional()`, `getAll()`, `execute()`, ... methods by hand. We introduced mixins to help with those classes, but that's still kind of awkward to use since it splits the type hierarchy (one class + another one applying the mixin).

Really, what we want is for users to implement a minimal amount of methods and then get all the others by default. Now that we can afford a breaking change, we can use the right tool for the job: This changes `DBAdapter` and `LockContext` to be abstract classes, meaning that users only really need to implement a single method (`executeRaw`) and get the rest for free. `execute` can still be implemented in cases where that's more efficient than reconstructing rows from raw objects.

Specifically, this PR makes three changes:

## Result set restructuring

Our current `QueryResult` interface is heavily inspired from RNQS, which I don't think is the most convenient representation: Rows are represented as JS objects, and accessed on results through a field called `_array`. There's also a `length` getter and an `item()` method to get the nth row (a convenient API for those not knowing arrays exist /s).

This keeps that interface for backwards compatibility, but also adds an `array` getter directly to `QueryResult` to obtain mapped rows. Additionally, `QueryResult` (and `execute`) are generic now, which accessing rows in a  rows in a similar way to `get()`, `getAll` and `getOptional`.

`executeRaw` also exposes the last insert id and changes now, which may be useful in some cases and allows implementing `execute` by delegating to `executeRaw` and mapping results.

## Table update notifications

On a type level, we had three different representations of update notifications and helpers to convert between them. This normalizes them into a single format, `changedTables: string[]`. Information about the type of updates (insert, update, delete) was only provided by the OP-SQLite implementation and isn't used anywhere in the SDK.

## Migration to abstract classes

The previous mixins used to implement `DBAdapter` were introduced to share implementation details without a breaking change. Using abstract classes instead gives us a much simpler implementation and avoids having to split everything into two classes to apply the mixin.

One minor caveat is that `TriggerDiffHandlerContext` (an interface) can no longer implement `LockContext`, it now exports the context as a field instead.


.changeset/sixty-peas-grow.md
LucDeCaf, line 1: I noticed `@powersync/common` isn't in the changeset. That could very well be due to v2 restructurings I'm not fully aware of, flagging anyways just-in-case.

packages/adapter-sql-js/src/SQLJSAdapter.ts
```diff
@@ -173,60 +167,43 @@ class SqlJsConnectionPool extends BaseObserver<DBAdapterListener> implements Con
...
 
-class SqlJsExecutor implements SqlExecutor {
-  constructor(readonly db: SQLJs.Database) {}
+class SqlJsLockContext extends LockContext {
+  constructor(readonly db: SQLJs.Database) {
+    super();
+  }
+
+  get connectionType() {
+    return undefined;
```
LucDeCaf: I'm curious as to when we would or wouldn't want to return an actual value from `connectionType`. I'm guessing it's because we don't want to make the interface too restrictive on external implementors, but I also think that we probably want to implement this for our first-party LockContext implemetations.

simolus3:
Marking this as WIP, as op-sqlite doesn't seem to return the information we need (contrary to its typescript definitions). I've opened https://github.com/OP-Engineering/op-sqlite/issues/419, let's see what comes out of that.

packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts
```diff
@@ -232,11 +223,3 @@ class OPSQLiteConnectionPool extends BaseObserver<DBAdapterListener> implements
     }
   }
 }
-
-export class OPSQLiteDBAdapter extends DBAdapterDefaultMixin(OPSQLiteConnectionPool) implements DBAdapter {
-  async executeBatch(query: string, params: any[][] = []): Promise<QueryResult> {
```
LucDeCaf, line 237: It's possible this is reimplemented elsewhere, but it looks to me like we've lost the native OPSQLite executeBatch call in this file's refactor.

Edit: I see later on that we used to type `OPSQLiteExecutor` as `Omit<DBAdapter, 'executeBatch'>`, so removing executeBatch could have been intentional.

Edit edit: It looks like `executeBatch`'s native implementation is re-added in the actual `OPSQLiteConnection` object; not sure what the drawbacks would be to having the native implementation in the LockContext as well.

packages/shared-internals/src/client/sync/bucket/SqliteBucketStorage.ts
```diff
@@ -101,13 +101,10 @@ export class SqliteBucketStorage extends BaseObserver<BucketStorageListener> imp
         return false;
       }
 
-      const rs = await tx.execute("SELECT seq FROM main.sqlite_sequence WHERE name = 'ps_crud'");
-      if (!rs.rows?.length) {
```
LucDeCaf, line 105: Just "check"-ing that this "check" is no longer needed (haha), I don't have a lot of context on SQLite bucket storage.

packages/web/src/db/adapters/wa-sqlite/DatabaseClient.ts
```diff
@@ -190,15 +183,20 @@ export class DatabaseClient<Config extends SQLOpenOptions = WebDBAdapterConfigur
  * While an instance is active, it has exclusive access to the underlying database connection (as represented by its
  * token).
  */
-class ClientSqlExecutor implements SqlExecutor {
+class ClientSqlExecutor extends LockContext {
```
LucDeCaf: Nit: Probably more accurate now to call this `ClientLockContext` or similar. (would also need to update the doc-comment above it)

LucDeCaf (commented):
Nice stuff, this will make first- and third-party DBAdapters so much easier!

LucDeCaf:
> Marking this as WIP, as op-sqlite doesn't seem to return the information we need (contrary to its typescript definitions). I've opened [OP-Engineering/op-sqlite#419](https://github.com/OP-Engineering/op-sqlite/issues/419), let's see what comes out of that.

Ah, just saw this now - that's frustrating. Hopefully it gets resolved quickly.

rkistner:
Some general comments on the API structuring:

On the adapter level, while we can covert from raw rows to the object format without loss, doing that in JS is often slower than letting the driver do that conversion. This matters when working with 10-100k+ results with many small columns. Since accessing the results as objects is more common, we should optimize for that. Unfortunately that means in user-side API, the user will still have to specify upfront whether the results should be in object or raw array format, instead of just different accessors on the results.

Then on the end-user API, can we simplify the results for the `execute` function further?

While it's not feasible to just return an array like `getAll()` does, we can perhaps remove one level of nesting, even if it's just as additional helpers? For example:

```ts
const rs = await db.execute('insert into my_table(id) values(...) returning *');
// Before
const rows = rs.rows?._array;
// This PR
const rows = rs.rows?.mapped;
// Maybe just an additional getter like this?
const rows = rs.mapped;
```

It would be even nicer to just use `rs.rows` as an array of objects, but that might be too big a breaking change?

And for a top-level array on the results (`rs.mapped` above), perhaps we can have it always be set to a (potentially-empty) array, instead of an optional field? That could help for the common case of accessing the results, without needing the optional chaining operator. If the user does need to distinguish between no results vs 0 rows, we can still have the more advanced APIs available.

packages/shared-internals/src/client/sync/bucket/SqliteBucketStorage.ts
simolus3: That check is part of `tx.get` (which throws if no row is returned). So it's still there, but I don't think there a realistic way for the sequence to be empty here.

simolus3:
> user-side API, the user will still have to specify upfront whether the results should be in object or raw array format

I agree, the API I originally had in mind seems to be impossible to implement efficiently. We have `RawQueryResult` and `QueryResult` as distinct types now. OP-sqlite version 17.0.0 should expose column names, the last insert id and affected rows from `executeRaw`, so it makes sense to share parts in a common superinterface.

> It would be even nicer to just use `rs.rows` as an array of objects, but that might be too big a breaking change?

I agree, but we can make `QueryResult` generic over the result type and have it implement `Iterable<T>`:

```TypeScript
// Before
const rs = await db.execute('insert into my_table(id) values(...) returning *');
const rows = rs.rows?._array;

// After
const rs = await db.execute<{id: string, name: string}>('insert into ... returning *');
for (const { id, name } in rs) {
  // ...
}

// Or instead:
const rows = rs.array;
```

This feels reasonably simple to me without requiring another field. Let me know what you think.

packages/powersync-op-sqlite/src/db/OPSqliteAdapter.ts
simolus3: The way we had it before was actually correct. I thought we could benefit from batches by implementing them on the connection level, but since op-sqlite starts a transaction for batches we can't use that method in transactions.

So the current setup adopts native batches when called as a top-level method, but falls back to the slower implementation when that method is called in transactions. I'll revert the changes to that.

rkistner:
> I agree, but we can make `QueryResult` generic over the result type and have it implement `Iterable<T>`:
> 
> ```ts
> // Before
> const rs = await db.execute('insert into my_table(id) values(...) returning *');
> const rows = rs.rows?._array;
> 
> // After
> const rs = await db.execute<{id: string, name: string}>('insert into ... returning *');
> for (const { id, name } in rs) {
>   // ...
> }
> 
> // Or instead:
> const rows = rs.array;
> ```
> 
> This feels reasonably simple to me without requiring another field. Let me know what you think.

I like that approach - I didn't even think about the possibility of supporting `Iterable<T>` directly on `QueryResult`. And I also that the additional `rs.array` option is good to have for the cases where you want an actual array.

packages/adapter-sql-js/src/SQLJSAdapter.ts
simolus3: We only use this to avoid a `BEGIN IMMEDIATE` for op-sqlite adapters (because we use readwrite connections with `pragma query_only`, where `BEGIN IMMEDIATE` for read transactions would not work).

I've removed `undefined` as an option from the type, implementations can return the actual connection variant.

rkistner: Side note - should we do a feature request for actual read-only connections with op-sqlite? It could help to eventually unify the implementation to use BEGIN IMMEDIATE everywhere.

simolus3: I have opened https://github.com/OP-Engineering/op-sqlite/issues/421, if that's accepted we can remove this field later.

LucDeCaf (approved):

PR #996, opened 2026-06-17
Remove `event-iterator` dependency

simolus3:
I'm looking at ways to reduce build complexity in the JS SDK, and getting rid of rollup to make `tsc` output the only thing we ship would help a lot. Currently, `@powersync/common` has two good reasons to apply rollup:

1. We want to bundle RSocket dependencies.
2. For compatibility with older React Native / Hermes versions, we want to replace `Symbol.asyncIterator` references with a ponyfill.

I'll look at 1 in another PR to the `v2` branch, but 2 is something we can fix without a breaking change. `event-iterator` is the only dependency using `Symbol.asyncIterator`, so we can rewrite our usages with a custom implementation of it. The package is essentially a bridge between `onEvent` callbacks and async iterators, and we already had most of that in the existing `asyncNotifier` utility.

So, this:

1. Expands `asyncNotifier` into a `EventQueue` implementation allowing more than a single outstanding event.
2. Refactors `onChangeWithAsyncGenerator` and `watchWithAsyncGenerator` to be based on the `EventQueue` utility. This actually fixes a bug, breaking out of an async iterator would not have cleaned up the inner listener before.
3. Refactors the RSocket client to use that as well. The interface is slightly simpler here, we don't need to worry about `AsyncIterator.return()`.

Disclosure: Most added tests were generated with Claude Code, I manually tweaked and verified them. I smoke-tested the RSocket client with the Node.js demo since it's not covered by automated tests.

### DIFF
--- a/.changeset/slimy-penguins-cheer.md
+++ b/.changeset/slimy-penguins-cheer.md
@@ -1,0 +1,5 @@
+---
+'@powersync/common': patch
+---
+
+Remove `event-iterator` dependency.
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -59,7 +59,6 @@
     "test:exports": "attw --pack . --exclude-entrypoints internal/sync_protocol"
   },
   "dependencies": {
-    "event-iterator": "^2.0.0",
     "js-logger": "catalog:"
   },
   "devDependencies": {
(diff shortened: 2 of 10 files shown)


Chriztiaan (approved):

simolus3:
I tested this in a fresh expo project (all our demos just render a black screen for me) with a copied `.pnpmfile.cjs`.

PR #997, opened 2026-06-17
Fix drizzle and `updateSchema` using transactions not compatible with WAL OPFS

simolus3:
When using `OPFSWriteAheadVFS`, it's important to use `BEGIN IMMEDIATE` for transactions so that the VFS can properly transition to an isolated write state, a plain `BEGIN` is not supported.

This fixes two issues forwarded to me by @bean1352:

1. When called outside of a transaction, `powersync_replace_schema` internally sets up its own transaction using a plain `BEGIN` statement. We'll [change that to throw instead](https://github.com/powersync-ja/powersync-sqlite-core/issues/190), this PR fixes the issue by adding an explicit transaction around it.
2. The Drizzle driver uses plain `BEGIN` and `COMMIT` statements. It does that because `drizzle` supports a transaction hint describing whether `BEGIN EXCLUSIVE` should be used. IMO, ignoring that hint and letting the DB adapter use its preferred way of opening transactions is a better thing to do though. We don't really have a choice if we want to support drizzle with the OPFS WAL VFS.


Chriztiaan (dismissed):

packages/drizzle-driver/tests/sqlite/lock-usage.test.ts
```diff
@@ -370,47 +384,50 @@ describe('Lock Usage Tests', () => {
 
   describe('Transaction operations', () => {
     lockUsageTest(
-      'should use writeLock for read-write transaction (default)',
-      async ({ db, readLockSpy, writeLockSpy }) => {
+      'should use writeTransaction for read-write transaction (default)',
+      async ({ db, readLockSpy, readTransactionSpy, writeTransactionSpy }) => {
```
LucDeCaf: `readLockSpy` is now unused.

LucDeCaf (dismissed):
LGTM, just one unused variable in a test file.

LucDeCaf (approved):

PR #998, opened 2026-06-17
chore: Minor changes to demos to accommodate recent open option changes

Chriztiaan:
Noticed a few demos had `tsc` and runtime issues related to the refactor from https://github.com/powersync-ja/powersync-js/pull/987.

This is a quick pass to fix the low hanging fruit.


simolus3 (approved):
Thanks! I'm kind of surprised the CI check for demos didn't catch this.

(I also want to test most demos before releasing v2, but some of them are so outdated that I can't even get them to run due to old React Native / Electron versions).

PR #999, opened 2026-06-18
Remove data flow status to align `SyncStatus` interface with other SDKs

simolus3:
The `dataFlowStatus` getter on `SyncStatus` only exists in the JavaScript SDKs. On every other PowerSync SDK, its fields (`uploading`, `downloading`, `uploadError` and `downloadError`) are part of the main `SyncStatus` class.

In https://github.com/powersync-ja/powersync-js/pull/993, `downloading` was moved from the data flow status to the main `SyncStatus` class. Christiaan pointed out that having `downloading` and `uploading` on separate interfaces isn't great, so this PR completes the migration by moving everything to the main `SyncStatus` interface.

To not make upgrading to v2 harder than it needs to be, I've kept `SyncDataFlowStatus` getters around as deprecated members. Internally, we also need to track these fields separately because they're not managed by the sync status from the core extension. This uses `JavaScriptSyncState` to refer to these values, but that type is only relevant internally.

I patched the demos manually by grepping for `dataFlowStatus`.  I have manually tested the `react-supabase-todolist` demo.

### DIFF
--- a/packages/shared-internals/src/db/crud/SyncStatus.ts
+++ b/packages/shared-internals/src/db/crud/SyncStatus.ts
@@ -164,7 +164,7 @@ export class SyncStatusSnapshot implements SyncStatus {
    * Not all errors are serializable over a MessagePort. E.g. some `DomExceptions` fail to be passed across workers.
    * This explicitly serializes errors in the SyncStatus.
    */
-  protected serializeError(error?: Error) {
+  serializeError(error?: Error) {
     if (typeof error == 'undefined') {
       return undefined;
     }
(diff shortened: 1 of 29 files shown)


packages/shared-internals/src/db/crud/SyncStatus.ts
```diff
@@ -140,7 +164,7 @@ export class SyncStatusSnapshot implements SyncStatus {
    * Not all errors are serializable over a MessagePort. E.g. some `DomExceptions` fail to be passed across workers.
    * This explicitly serializes errors in the SyncStatus.
    */
-  protected serializeError(error?: Error) {
+  serializeError(error?: Error) {
```
simolus3, line 167: Making this `protected` while adding it to the internal-but-shared-across-packages interface `AbstractStreamingSyncImplementationListener` causes TypeScript errors for the nuxt package.

I believe this is because tsc thinks the web SDK imports types from `lib/`, while nuxt imports types from `dist/`. When a `protected` member is involved, these actually have to point to the same definition or tsc is going to complain. This wouldn't affect users, but it's kind of annoying and yet another reason to get rid of bundlers when we can.

Chriztiaan (approved):
Looks good!

PR #1000, opened 2026-06-22
[Fix] Preserve Error.cause on SyncStatus upload/download errors

bean1352:
`SyncStatus` flattens `uploadError` and `downloadError` into plain objects before they reach `statusChanged` listeners. The flattening only kept `name`, `message`, and `stack`, so `Error.cause` was dropped. 

## Change

- `serializeError` now includes `cause`. It recurses for `Error` causes so the entire chain is preserved, and passes non-`Error` causes through unchanged (`Error.cause` can be any value).

There is no API change, and the existing `name`, `message`, and `stack` behaviour is unchanged.

## Testing

Reproduced in a React Native app by throwing `new Error('Test Error', { cause: new Error('Cause Error', { cause: new Error('Deep Error') }) })` from the connector's `uploadData`, then logging `uploadError` from a `statusChanged` listener.

- Before the fix, the listener received only `Test Error` and its `cause` was `undefined`.
- After the fix, the listener received the full chain: `Test Error`, caused by `Cause Error`, caused by `Deep Error`.

## Note

Claude Opus 4.8 helped research where the cause was being lost. I reproduced the issue and tested the fix myself.

### DIFF
--- a/packages/common/src/db/crud/SyncStatus.ts
+++ b/packages/common/src/db/crud/SyncStatus.ts
@@ -277,21 +277,20 @@ export class SyncStatus {
    * Not all errors are serializable over a MessagePort. E.g. some `DomExceptions` fail to be passed across workers.
    * This explicitly serializes errors in the SyncStatus.
    */
-  protected serializeError(error?: Error) {
+  protected serializeError(error?: Error): Error | undefined {
     if (typeof error == 'undefined') {
       return undefined;
     }
-    const serialized: { name: string; message: string; stack?: string; cause?: unknown } = {
+    const serialized: Error = {
       name: error.name,
       message: error.message,
       stack: error.stack
     };
     // `Error.cause` can be any value (the spec types it as unknown). Preserve it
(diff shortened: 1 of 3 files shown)


packages/common/src/db/crud/SyncStatus.ts
```diff
@@ -285,11 +281,19 @@ export class SyncStatus {
     if (typeof error == 'undefined') {
       return undefined;
     }
-    return {
+    const serialized: { name: string; message: string; stack?: string; cause?: unknown } = {
```
Chriztiaan: Can we use the built-in Error type? Or is there a reason we can't? That would also clean the typing on line 292.
Image

Chriztiaan:
Looks good, just the changeset entry beyond the typing question.
This might be an API change that we need to include according to the build errors, will find the steps you need to follow.

Edit:
```
cd packages/common
pnpm api-extractor run --local 
cp temp/common.api.md etc 
*git check-in file*
```

Chriztiaan (approved):

PR #1002, opened 2026-06-23
Load RSocket libraries dynamically

simolus3:
RSocket is a pain to import for users, and the only remaining reason we apply a bundler to our SDK. We can't avoid that entirely, but we can apply some tricks to only bundle code directly importing RSocket.

Before this PR, all of `@powersync/shared-internals` and dependencies like RSocket were bundled into a single file. Now, we:

1. Use plain tsc JavaScript modules for almost everything, except:
2. A single file providing RSocket support, written in a way to avoid any other imports.

This means that most of the SDK is now much easier to work on (more packages that can be built with `tsc -w` alone), and smaller too: The minified `WebSocketSupport` file with dependencies is 26 kB (111 kB uncompressed) that no longer need to be loaded and parsed unless WebSockets are actually used. 

This also simplifies `AbstractRemote a bit:

- Remove `AbstractRemoteOptions`! It was an implementation detail that shouldn't leak into public APIs.
  - Make `fetch` an abstract method, allowing implementations to call the global `fetch` directly (avoiding `bind(globalThis)`).
  - Avoid passing `fetchOptions` through different calls, if they are needed then an implementation should just add them before calling `fetch`.
- For the user agent, avoid importing `package.json` to get the version. Using a constant updated by changesets avoids having to bundle the entire `package.json`.

Manual testing:

- [x] Node with HTTP
- [x] Node with WebSockets
- [x] Vite-bundled app with HTTP
- [x] Vite-bundled app with WebSockets
- [x] Webpack-bundled app with HTTP
- [x] Webpack-bundled app with WebSockets
- [x] ReactNative with HTTP
- [x] ReactNative with WebSockets

### DIFF
--- a/packages/shared-internals/src/client/sync/stream/AbstractRemote.ts
+++ b/packages/shared-internals/src/client/sync/stream/AbstractRemote.ts
@@ -1,5 +1,4 @@
 import { type fetch } from 'cross-fetch';
-import PACKAGE from '../../../../package.json' with { type: 'json' };
 import { FetchStrategy, PowerSyncCredentials, LogLevels, PowerSyncLogger } from '@powersync/common';
 
 import { AbortOperation } from '../../../utils/AbortOperation.js';
@@ -7,10 +6,10 @@ import {
   doneResult,
   extractBsonObjects,
   extractJsonLines,
-  SimpleAsyncIterator,
-  valueResult
+  SimpleAsyncIterator
 } from '../../../utils/stream_transform.js';
-import { webSocketSyncStream } from './WebSocketSupport.js';
+import { WebSocketSupport, WebSocketSyncStreamPlatform } from './WebSocketSupport.js';
+import { EventQueue } from '../../../utils/async.js';
 
 /**
  * @internal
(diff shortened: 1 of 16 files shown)


packages/shared-internals/src/client/sync/stream/AbstractRemote.ts
```diff
@@ -23,78 +20,43 @@ export type RemoteConnector = {
 };
 
 const POWERSYNC_TRAILING_SLASH_MATCH = /\/+$/;
-const POWERSYNC_JS_VERSION = PACKAGE.version;
+// Note: A postversion script will make Changesets keep this constant up to date with the version in package.json.
+const POWERSYNC_JS_VERSION = 'unset before initial release';
```
simolus3: I'll create a pre-release checklist to make sure the changesets PR updates this before a release.

rkistner: Could it be cleaner to have this constant in a dedicated file?

Chriztiaan (approved):

PR #1004, opened 2026-06-23
Remove old schema v1 syntax

simolus3:
The `Table` constructor supports two syntax variants:

1. A deprecated v1 variant where columns and indices are passed as arrays.
2. The v2 variant we use by default everywhere, using JS objects to declare column names and types.

Since the v1 variant is deprecated and a comment explains that it'll be removed in the next major release, now is the time for that. This PR:

- Removes explicit v2 names (like `TableV2Options`), replacing them with plain names (`TableOptions`).
- Removes the v1 syntax from the `Table` constructor.

The ability to construct a table from an array of columns is still useful, however:

- In the diagnostics client, creating a schema through table arrays is simpler than having to construct records.
- We have to represent tables as columns arrays at some point, as that's the format expected by the core extension.

So, this adds `ResolvedTable` as an advanced class representing columns the old way. This kind of weakens the idea of removing v1 variants, but I think being explicit about the way tables are constructed (instead of allowing both styles on the same constructor) is a decent simplification already.

_AI use_: I used Claude Code to help migrate a few helper types and demos.


LucDeCaf (approved):
LGTM; there's a few unused imports, but I'm assuming those will be optimised away by the bundler.

PR #1008, opened 2026-06-25
Web: Merge workers, simplify custom build

simolus3:
This makes two changes to the web SDK: A minor change to merge the sync and database workers into one, and a larger / breaking change to remove the UMD build and workers.

## Merging workers

Since most apps use both a sync and a database worker, merging them into one can be slightly more efficient since code duplicated in them doesn't have to be loaded twice.

This introduces `web/src/worker/worker.ts` as an entrypoint for both database and sync workers. When it detects that it's running as a dedicated worker, it behaves exactly like the old database worker. Shared workers can take both roles, so clients will set up a message channel and send a port to the worker along with the tag indicating what the channel is for (database or sync).

## Build simplifications

In decent setups, bundlers for apps would detect the `new SharedWorker(new URL('./worker.js', import.meta.url))` pattern we use to compile and bundle `worker.js`. Webpack and vite both do this, and it's the best way to use the web SDK. This pattern works the exact same way as it did before.

Some systems, notably Metro (React Native Web), Nuxt and Angular, don't work that way. For those, we've previously shipped a custom worker build under `@powersync/web/umd/worker/{sync|db}` using webpack. Then there's a script users need to run to copy those compiled workers into their app. This is okay (similar to what we have on the Dart SDK), but there are a few possible simplifications:

1. That build uses webpack, while most of our other tooling is based on rollup. I think we should stick to a single tool, so I migrated the build to use rollup.
2. The worker was compiled as a UMD module. This seems like an odd choice to me since workers are standalone programs, IIFE would have probably been more appropriate? But since we load the default worker with `type: module`, the best way to ensure consistency between our setup and custom worker builds would be to bundle the worker as an ESM module too.
3. We also shipped an UMD module of `@powersync/web` that users can import. This is used for React Native Web, but again I think we should stick to ESM modules wherever possible (since Metro supports those). So I've replaced that with an ESM bundle instead. We can't remove this bundle entirely, because:
4. For React Native Web, the builtin way of loading workers in the web SDK is broken without bundlers. The reason is that Metro bundles files into something that isn't an ESM module, but also doesn't replace `meta.import.url` (which we use to construct workers). With the previous UMD build, webpack replaced `meta.import.url` with a value that wouldn't have worked at runtime. In the new build, I've replaced the only function using that syntax with a stub that throws. Users need to use custom workers anyway, this gives them an actionable error message. A conditional export means decent setups get TS outputs as before, React Native Web get the bundle.

## What does this mean for users

For apps built with webpack and vite using the default worker, nothing changes! For apps using a custom worker, they should replace `/@powersync/worker/WASQLiteDB.umd.js` and `/@powersync/worker/SharedSyncImplementation.umd.js` with `/@powersync/worker.js`.

React Native Web users need to stop importing `@powersync/web/umd`. Instead, they should import `@powersync/web` directly. This works with plain React Native, unfortunately expo removes RN-specific conditional imports for the web so one needs to add `config.resolver.unstable_conditionsByPlatform.web.push('react-native-web');` to `metro.config.js`. It would have been really nice if this worked out of the box, but the new setup is still simpler than the old one.

## Tests

I have manually tested the Angular, Nuxt and React Native Web demos on the web since they need build changes. I have also tested the webpack and react-supabase-todolist (vite) demos to make sure things are still working.

For the RNW demo, I also tested a release build (from `pnpm build:web`).

AI use disclosure: Claude Code helped with custom rollup plugins for the web build.


demos/example-nextjs/src/library/powersync/powersync-provider.tsx
```diff
@@ -18,13 +18,13 @@ function getDB(): PowerSyncDatabase {
       logger,
       open: {
         dbFilename: 'powersync-nextjs.db',
-        worker: '/@powersync/worker/WASQLiteDB.umd.js',
+        worker: '/@powersync/worker.js',
```
Chriztiaan: Is custom workers still needed for Next?

Chriztiaan (approved):

demos/example-nextjs/src/library/powersync/powersync-provider.tsx
simolus3: Interestingly they aren't! I tried removing the worker options and turbopack correctly recognized `new SharedWorker` and `new Worker` invocations in the SDK and bundled the worker. I've removed the option and tested both dev and release builds.

PR #1010, opened 2026-06-26
Remove react native polyfills and RNQS support

simolus3:
The React Native SDK applies rollup to bundle a bunch of polyfills. Some of those are related to BSON which we no longer use in JavaScript and could have likely been removed a while ago, others are breaking changes.

The changes in this PR are:

1. Inline `@powersync/op-sqlite` into `@powersync/react-native` and remove the separate package, making `op-sqlite` the only option.
2. Remove `react-native-fetch-api` as a fetch polyfill. It was responsible for a lot of workarounds in the RN package, like not supporting binary responses and requiring additional fields to support streaming. Instead, we now:
    - use `expo/fetch`, which actually implements the fetch API and supports streaming responses.
    - when that's unavailable, fall back to the default `fetch` implementation from RN. This doesn't support streaming responses, we throw in that case. We can still use it to request write checkpoints though.
    - The default connection method is now SDK and configuration-specific: We use http when possible, and fall back to WebSocket on React Native if no compatible fetch client is available.
3. Remove text decoding polyfills: `TextDecoder` is available in recent RN versions, _and_ we only use it to decode ndjson responses (both the expo client and web sockets would give us a stream of BSON objects which we don't decode).

This allows removing rollup from `@powersync/react-native` entirely, we only ship the TypeScript outputs now.

## Tests

Most React Native demos in this repository use an old RN version, and I can't get them to work. Instead, I tested these changes with:

1. The `tools/powersynctests` demo app, which I've also updated to React Native 0.86.
2. The `react-native-barebones-opsqlite` demo, which I've also updated to React Native 0.86 (using RSocket for downloads and global fetch to request write checkoints).
3. A standalone expo-based app, which was still on RN 0.85 but things seemed to work using expo fetch.

IMO, we should either take some time to modernize or RN demos or remove them. As per our plan to only support platform versions that are still supported, `0.85.x` would be the oldest RN version we support and yet we still have demos on `0.79.x`.


packages/react-native/package.json
```diff
@@ -33,38 +35,22 @@
...
+    "@op-engineering/op-sqlite": "catalog:",
+    "expo": "catalog:",
     "@types/react": "^19.1.1",
     "react": "^19.2.0",
-    "react-native": "0.83.1",
-    "react-native-fetch-api": "^3.0.0",
-    "rollup": "catalog:",
-    "text-encoding": "^0.7.0",
-    "web-streams-polyfill": "3.2.1"
+    "react-native": "^0.86.0"
```
Chriztiaan: Since Expo 56 (latest) only supports 0.85, it may be better for us to drop this too.

demos/react-native-supabase-todolist/package.json
```diff
@@ -12,7 +12,7 @@
     "@azure/core-asynciterator-polyfill": "^1.0.2",
     "@dr.pogodin/react-native-fs": "2.37.0",
     "@expo/vector-icons": "^15.1.1",
-    "@journeyapps/react-native-quick-sqlite": "^2.5.2",
+    "@op-engineering/op-sqlite": "^17.0.0",
```
Chriztiaan, line 15: If I recall correctly, switching from RNQS to Op-sqlite like this (which upgrade path wise is actually quite nice imo) should be seamless and uses the same sqlite file right (and doesn't cause a full resync)?

Chriztiaan (approved):
Cool diff!

Chriztiaan (approved):

demos/react-native-supabase-todolist/package.json
simolus3: The path is the same on iOS but not on Android. As discussed offline, I've added a native helper method in 543ee74684f4e4f0f12221c7c2675d920fd8d19f that will make us use the old database location if a file exists at that path.

I've tested this in a fresh demo app by first opening it with RNQS and then applying overrides to use the v2 sources. I couldn't get RNQS to run on iOS to test that, but I see the code is exactly the same ([old](https://github.com/powersync-ja/react-native-quick-sqlite/blob/5fbbe50013bc0600f42547f76c61986bc420b5fc/ios/QuickSQLite.mm#L51-L53), [new](https://github.com/OP-Engineering/op-sqlite/blob/ca6e573cebbe6934618bf697b3feb4d70f09ed9d/ios/OPSQLite.mm#L70-L72)).

PR #1012, opened 2026-06-29
Use proper read-only connections with op-sqlite

simolus3:
To emulate read-only connections with the OP-SQLite adapter, we currently rely on `pragma query_only = true`. This mostly works, but it breaks `BEGIN IMMEDIATE` statements to start transactions. Starting with `@op-engineering/op-sqlite` version `17.1.0`, it's possible to open connections with `SQLITE_READONLY`.

This PR updates `op-sqlite` dependencies, adopts read-only connections, and removes `query_only` workarounds.

### DIFF
--- a/demos/react-native-barebones-opsqlite/package.json
+++ b/demos/react-native-barebones-opsqlite/package.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@azure/core-asynciterator-polyfill": "^1.0.2",
-    "@op-engineering/op-sqlite": "^17.0.0",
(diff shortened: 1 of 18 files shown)


LucDeCaf (approved):
Looks good - just to confirm, was `getConnectionType()` only ever used to help with mimicing readonly in `op-sqlite`?

simolus3:
> just to confirm, was `getConnectionType()` only ever used to help with mimicing readonly in `op-sqlite`?

Exactly, that was the only internal use. It's essentially an implementation detail, removing it is still technically a breaking change but fine for v2.

PR #1015, opened 2026-06-30
fix(ts): some types polyshing, shorten filenames in web package dist/worker folder

delagen:
Fixed errors with skipLibCheck:false

Multiple errors as
```
TS2503: Cannot find namespace 'Logger'
    node_modules/@powersync/common/lib/utils/Logger.d.ts:11:10:
      11 │     WARN: Logger.ILogLevel;
         ╵           ~~~~~~
```

Allow to use HEX key for encryption as stated in https://utelle.github.io/SQLite3MultipleCiphers/docs/configuration/config_sql_pragmas/#pragma-key--hexkey
by replacing quotes and forcefully set ChaCha20 as cipher, because under hood it's seems not default as stated in docs

### DIFF
--- a/packages/web/src/db/adapters/wa-sqlite/RawSqliteConnection.ts
+++ b/packages/web/src/db/adapters/wa-sqlite/RawSqliteConnection.ts
@@ -45,8 +45,8 @@ export class RawSqliteConnection {
     );
     await this.executeRaw(`PRAGMA temp_store = ${this.options.temporaryStorage};`);
     if (this.options.encryptionKey) {
-      const escapedKey = this.options.encryptionKey.replace("'", "''");
-      await this.executeRaw(`PRAGMA key = '${escapedKey}'`);
+      const escapedKey = this.options.encryptionKey.replaceAll("'", "''");
+      await this.executeRaw(`PRAGMA key = '${escapedKey}';`);
     }
     await this.executeRaw(`PRAGMA cache_size = -${this.options.cacheSizeKb};`);
 
--- a/packages/common/src/client/sync/stream/AbstractRemote.ts
+++ b/packages/common/src/client/sync/stream/AbstractRemote.ts
@@ -42,7 +42,7 @@ const KEEP_ALIVE_LIFETIME_MS = 90_000;
 /**
  * @internal
  */
-export const DEFAULT_REMOTE_LOGGER = Logger.get('PowerSyncRemote');
+export const DEFAULT_REMOTE_LOGGER: ILogger = Logger.get('PowerSyncRemote');
 
 /**
  * @internal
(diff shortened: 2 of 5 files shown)


packages/web/src/db/adapters/wa-sqlite/RawSqliteConnection.ts
```diff
@@ -45,8 +45,8 @@ export class RawSqliteConnection {
     );
     await this.executeRaw(`PRAGMA temp_store = ${this.options.temporaryStorage};`);
     if (this.options.encryptionKey) {
-      const escapedKey = this.options.encryptionKey.replace("'", "''");
-      await this.executeRaw(`PRAGMA key = '${escapedKey}'`);
+      const escapedKey = this.options.encryptionKey.replace(/"/g, `""`);
+      await this.executeRaw(`PRAGMA key = "${escapedKey}"`);
```
simolus3: Why would we use double quotes here? It should be a string literal, which uses single quotes. We don't currently support hexkeys, doing that would also require a new option to not break existing users.

simolus3 (commented):
While the logger type changes look reasonable to me, we're also working on a new major version of the SDK on the `v2` branch that removes js-logger entirely. We're aiming to at least have a dev release for that out soon. So I'm not sure if it still makes sense to merge this.

packages/web/src/db/adapters/wa-sqlite/RawSqliteConnection.ts
rkistner: See the ~SQLCipher docs https://www.zetetic.net/sqlcipher/sqlcipher-api/#PRAGMA_key~ docs https://utelle.github.io/SQLite3MultipleCiphers/docs/configuration/config_sql_pragmas/#pragma-key--hexkey:

```
-- Example of a raw key for the SQLCipher scheme
PRAGMA key = "x'54686973206973206D792076657279207365637265742070617373776F72642E'";
```

Note that there is a difference between providing a raw key like that without key derivation, and providing a "passphrase" in hex to support binary data, using `PRAGMA hexkey`.

I'm not sure it's a good idea to let the user specify it in the above formats directly, but it could be nice to accept a Uint8Array key and/or raw key and do this conversion automatically.

Regardless, the current `replace("'", "''")` is a bug, since it would only replace the first quote.

delagen: Yes, current is bugged, and doesn't accept other key types. 

I thought about UInt8Array, but I didn't decide to change API, cause it doesn't solve other types, as Raw for example.

Introducing additional option may require additional validation to prevent provide both option at once.

delagen: Reverted quotes back to literal, currently prefer using 'raw:hex' value. Also forced to using chacha20 cipher stated in docs, because seems it's not default

delagen: Created MR to v2 #1016

PR #1016, opened 2026-06-30
fix(key): fix escaping of encryptionKey

delagen:


packages/web/src/db/adapters/wa-sqlite/RawSqliteConnection.ts
```diff
@@ -55,8 +55,8 @@ export class RawSqliteConnection {
     );
     await this.executeRaw(`PRAGMA temp_store = ${this.options.temporaryStorage};`);
     if (this.options.encryptionKey) {
-      const escapedKey = this.options.encryptionKey.replace("'", "''");
-      await this.executeRaw(`PRAGMA key = '${escapedKey}'`);
+      const escapedKey = this.options.encryptionKey.replace(/'/g, "''");
```
simolus3: Thanks! Could we use `replaceAll()` though? IMO that makes the intention a bit clearer:

```suggestion
      const escapedKey = this.options.encryptionKey.replaceAll("'", "''");
```

And then there are similar issues in `packages/node/README.md` and `demos/example-node/src/main.ts`, could you fix those as well?

packages/web/src/db/adapters/wa-sqlite/RawSqliteConnection.ts
```diff
@@ -55,8 +55,8 @@ export class RawSqliteConnection {
...
+      await this.executeRaw(`PRAGMA cipher = 'chacha20'; PRAGMA key = '${escapedKey}';`);
```
simolus3: What's the reason for selecting chacha20 explicitly? It's the default scheme for sqlite3mc.

simolus3 (approved):

packages/web/src/db/adapters/wa-sqlite/RawSqliteConnection.ts
delagen: Seems I incorrectly used https://github.com/Willena/sqlite-jdbc-crypt, I cannot open database via it.
Or may be cipher incorrectly applied.
Removed ChaCha20 forced pragma.

PR #1021, opened 2026-07-01
chore(v2): Updating Expo demos to Expo 57/React Native 0.86

Chriztiaan:
## Description
With v2 release looming, it was a good opportunity to update our React Native demos.

Updated and tested the following Expo demos:
- react-native-supabase-todolist
- react-native-supabase-group-chat
- react-native-supabase-background-sync
- react-native-web-supabase-todolist

Drive-by changes:
- Removed Django demo from ps-js monorepo, moved to community repo [here](https://github.com/powersync-community/powersync-django-react-native-todolist).
- `react-native-prompt-android` doesn't seem to play well with some of the newer versions of Expo/RN, so I opted to drop the dependency.

## Testing
To manually test:
1. `pnpm install`
2. `pnpm build:packages`
3. Link demos with monorepo packages with `pnpm demos:inject`
4. Navigate to demo, run `pnpm install`
5. Clear any stale build files with `npx expo prebuild --clean `
6. Run `pnpm ios`, `pnpm android`, `pnpm web`(for RNW).

PR #1025, opened 2026-07-02
Accumulated changes from v2 branch

simolus3:
The `v2` branch contains breaking changes for the JavaScript SDK that have already been reviewed. Commits in this come from the following PRs:

1. Remove `js-logger` in favor of a custom logging implementation (#966).
2. Refactor open and sync options to avoid having to use factories in most cases (#987).
   - Follow-up in #998 to adopt a few demos.
3. Encapsulate implementation details by moving them to `@powersync/shared-internals` (#993).
4. Align `SyncStatus` interface with other SDKs (#999).
5. Simplify `DBAdapter` interface and query result types (#995).
6. Load RSocket libraries dynamically (#1002).
7. Remove old schema v1 syntax (#1004).
8. Merge web workers and simplify the `@powersync/web` build (#1008).
9. Remove CommonJS builds and avoid bundlers in most of our packages (#1011).
10. React-Native improvements: Remove most polyfills and RNQS support (#1010).
    - Use proper read-only connections with `op-sqlite` to avoid the `pragma query_only` workaround (#1012).
    - Restore relevant parts of deleted readme (#1014).
11. Remove members annotated as `@deprecated` since before the `v2` fork (#1013).
12. Fix escaping `encryptionKey` (#1016).
13. Update demos to the latest Expo and React Native version (#1021).

Additionally, #1017 fixed dev publishing (already broken on `main`) and #1026 fixed a few test failures from new features on `main` I missed when rebasing  `v2`.

While the plan is to release this in around 2 weeks, we can already merge these changes to `main` to use as dev releases, targets for other PRs and to update external demos and integrations (I believe the initial logger changes broke `@tanstack/powersync-db-collection`, for example).


Chriztiaan (approved):
I have double checked the list of mentioned PRs which form part of this accumulated PR. I am happy with the changeset entries and the upgrade paths that are mentioned in them. 

I think it's realistic that we might have an above normal issue/query rate for a while after this releases, but I don't expect anything drastic.

PR #1026, opened 2026-07-02
Fix tests failing on `v2` branch

simolus3:
This fixes a few tests that started failing after rebasing the `v2` branch onto the latest changes from `main`. The three issues fixed are:

1. A new test in `tanstack-react-query` using internal APIs that have been hidden on `v2`, requiring changed imports.
2. Some attachment tests opening queues on which `stopSync()` is not called before the database is closed, which can lead to hard to debug asynchronous failures as they try to scan for attachments on a closed database. It also looks like there's a chance for `syncStorage()` to ignore abort signals when called after `stopSync()`, but this doesn't occur in our tests.
3. A web test setup requiring a bit more time to run more reliably.

AI use: I used Claude in the investigation on attachment failures, the changes here are written by hand.

### DIFF
--- a/packages/common/src/attachments/AttachmentQueue.ts
+++ b/packages/common/src/attachments/AttachmentQueue.ts
@@ -348,6 +348,7 @@ export class AttachmentQueue {
    */
   async syncStorage(): Promise<void> {
     const signal = this.syncAbortController?.signal;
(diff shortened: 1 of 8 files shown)


khawarizmus (approved):
LGTM

Chriztiaan (approved):

PR #1029, opened 2026-07-07
Web: Support in-memory databases

simolus3:
This adds an option to use the [example in-memory VFS](https://github.com/powersync-ja/wa-sqlite/blob/master/src/examples/MemoryVFS.js) from WA-SQLite with web databases.

For some online-first workloads, not using persistence at all can be an interesting option. It avoids OPFS and navigator locks (which have known issues on Safari), and is a lot faster for queries. So when there isn't a lot of data to sync, re-downloading it every time the app is opened can be more reliable and faster than using a proper file system implementation.

Without any further code changes, this VFS can be used in three configurations depending on passed flags:

1. In a shared worker (the default): Recommended, all tabs share the same database.
2. In a dedicated worker (when `enableMultiTab: false`). It's not really clear why one would want that, but it's supported.
3. Without any workers at all (`enableMultiTab: false` and `enableWebWorker: false`): By not using any message port roundtrips, this is _much_ faster than any other single-threaded VFS but runs queries synchronously which could block the UI for computationally-expensive queries.

### Multi-threaded in-memory VFS?

This uses a very simple in-memory VFS that can't be shared across workers, so it can't be used concurrently. In a way, that is a good thing because it makes concurrency control much simpler (we don't need navigator locks), but it also means that for workloads with a large amount of concurrent queries, those can't run in parallel.

There are ways to improve this setup:

1. We could use a write-ahead log to provide concurrent reads and writes, e.g. with [this implementation](https://github.com/whygee-dev/memory-write-ahead-vfs). The WAL implementation uses navigator locks though, which is something we want to avoid where possible.
2. I have ideas for a WAL VFS that doesn't use navigator locks at all, but wouldn't work across multiple tabs. But it's not clear whether this use case is common enough to optimize for it.

Fundamentally, multi-tab access to the same in-memory database will require navigator locks (either to prevent concurrent access or to coordinate readers/checkpoints). Since we want to avoid some of the pitfalls of navigator locks with this, exposing the simple in-memory VFS might be as good as it gets. If it turns out that we need parallel access after all, we can land improvements in follow-up PRs.

At this point though, I think the VFS implementation is no longer the lowest-hanging fruit for optimization. Looking into the worker protocol to reduce roundtrips would likely speed things up more, for example.

### DIFF
--- a/.changeset/brown-suits-push.md
+++ b/.changeset/brown-suits-push.md
@@ -1,0 +1,5 @@
+---
+'@powersync/web': minor
+---
+
+Allow using in-memory databases with `WASQLiteVFS.InMemoryVFS`.
--- a/packages/web/src/db/adapters/wa-sqlite/RawSqliteConnection.ts
+++ b/packages/web/src/db/adapters/wa-sqlite/RawSqliteConnection.ts
@@ -1,6 +1,6 @@
 import { Factory as WaSqliteFactory, SQLITE_ROW } from '@journeyapps/wa-sqlite';
 
-import { DEFAULT_MODULE_FACTORIES, WASQLiteModuleFactory, WASQLiteVFS } from './vfs.js';
+import { loadModuleAndVfs, WASQLiteVFS } from './vfs.js';
 import { TemporaryStorageOption } from '../options.js';
 import { RawQueryResult, SqliteValue } from '@powersync/common';
 
@@ -37,11 +37,8 @@ export class RawSqliteConnection {
    * The `sqlite3*` connection pointer.
    */
   private db: number = 0;
-  private _moduleFactory: WASQLiteModuleFactory;
 
-  constructor(readonly options: RawWaSqliteDatabaseOptions) {
-    this._moduleFactory = DEFAULT_MODULE_FACTORIES[this.options.vfs];
-  }
+  constructor(readonly options: RawWaSqliteDatabaseOptions) {}
 
   get isOpen(): boolean {
     return this.db != 0;
@@ -64,10 +61,7 @@ export class RawSqliteConnection {
   }
 
   private async openSQLiteAPI(): Promise<SQLiteAPI> {
-    const { module, vfs } = await this._moduleFactory({
-      dbFileName: this.options.filename,
-      encryptionKey: this.options.encryptionKey
-    });
+    const { module, vfs } = await loadModuleAndVfs(this.options);
     const sqlite3 = WaSqliteFactory(module);
     sqlite3.vfs_register(vfs, true);
     /**
--- a/packages/web/src/db/adapters/wa-sqlite/vfs.ts
+++ b/packages/web/src/db/adapters/wa-sqlite/vfs.ts
@@ -71,10 +71,10 @@ export async function loadModuleAndVfs({
   encryptionKey
 }: RawWaSqliteDatabaseOptions): Promise<{ module: SQLiteModule; vfs: SQLiteVFS }> {
   let moduleFactory = syncModuleFactory;
-  let resolveVfs: (module: any) => Promise<any>;
+  let resolveVfs: (module: any) => Promise<SQLiteVFS>;
 
   switch (vfs) {
-    case WASQLiteVFS.IDBBatchAtomicVFS:
+    case WASQLiteVFS.IDBBatchAtomicVFS: {
       moduleFactory = asyncModuleFactory;
       const { IDBBatchAtomicVFS } = await import('@journeyapps/wa-sqlite/src/examples/IDBBatchAtomicVFS.js');
       resolveVfs = (module) => {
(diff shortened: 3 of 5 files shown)


whygee-dev:
>I have ideas for a WAL VFS that doesn't use navigator locks at all, but wouldn't work across multiple tabs. But it's not clear whether this use case is common enough to optimize for it.

That made me curious, care to share it ?

rhashimoto:
> We could use a write-ahead log to provide concurrent reads and writes, e.g. with [this implementation](https://github.com/whygee-dev/memory-write-ahead-vfs). The WAL implementation uses navigator locks though, which is something we want to avoid where possible.

If you use a lazy write lock (only release the navigator lock on BroadcastChannel request), you would only pay for lock acquisition when a different connection wants to write. That's what OPFSWriteAheadVFS does by default.

rhashimoto:
> This adds an option to use the [example in-memory VFS](https://github.com/powersync-ja/wa-sqlite/blob/master/src/examples/MemoryVFS.js) from WA-SQLite with web databases.

Is there a reason to use wa-sqlite's MemoryVFS instead of SQLite's default internal memory database?

simolus3:
Good to see you both here :)

> That made me curious, care to share it ?

I'll work on a prototype for that today, but my basic approach would be to trigger checkpoints directly in JavaScript before giving out the write connection. We have the luxury of knowing that a transaction will never span across `readLock` / `writeLock` calls, so we could:

- Use three shared array buffers (metadata, main database, write-ahead log) as state shared across workers.
- In the main tab only, introduce a read/write lock implemented in JS.
- Have reads work as usual, letting workers use the WAL as an overlay. Also acquire a read lock while the `readLock` call is active.
- For writes, acquire a read lock for the duration of `writeLock` too and append to the WAL. But first, check the size of the WAL to determine whether a checkpoint is necessary and acquire a write lock for that. Because that means no queries will run concurrently, we need no further coordination.

This is kind of cheap and the real OPFSWriteAhead implementation has actual concurrency around checkpoints, but I suspect briefly blocking everything to copy a few buffers around won't hurt performance that much. That's something I'd need to verify with benchmarks though.

> If you use a lazy write lock (only release the navigator lock on BroadcastChannel request), you would only pay for lock acquisition when a different connection wants to write

This isn't about performance, we're concerned about [this WebKit issue](https://bugs.webkit.org/show_bug.cgi?id=316904) where locks held by workers are not reliably returned when the worker is force-closed due to e.g. a navigation change (this possibly has other causes too). This isn't an issue for OPFSWriteAhead since that doesn't work on Safari, but for an in-memory VFS we also use as a fallback, it would be nice to not have to use locks.

> Is there a reason to use wa-sqlite's MemoryVFS instead of SQLite's default internal memory database?

That is a good point. I didn't know there was a default VFS in WA-SQLite, I thought we had to pick one even for in-memory databases for things like randomness and the current time. Either way, backing in-memory databases by a VFS should be beneficial if we want to re-use connections across workers with shared buffers.

Chriztiaan (approved):

rhashimoto:
> This isn't about performance, we're concerned about [this WebKit issue](https://bugs.webkit.org/show_bug.cgi?id=316904) where locks held by workers are not reliably returned when the worker is force-closed due to e.g. a navigation change (this possibly has other causes too).

Instead of not using locks, you could create a watchdog to detect these zombie locks and break them (using the ["steal" lock option](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API#options)). That would only be safe for zombie exclusive locks (because you can't break just one shared lock of several), but with WAL I think you don't need shared locks if you can do without the seldom-used exclusive locking mode.

The ideal way to implement such a watchdog would use [Clients.matchAll()](https://developer.mozilla.org/en-US/docs/Web/API/Clients/matchAll) which is only available in a service worker. An alternative would be some sort of heartbeat. Maybe PowerSync already has a heartbeat somewhere?

simolus3:
I've thought about the heartbeat workaround too, my concern is that e.g. debuggers might pause a worker with a database lock for a while and then another tab could steal the lock. That's probably not realistic for apps since not too many people debug their database workers, but it's annoying when _we_ want to debug something and it feels fragile. But I am also running out of reliable workarounds here...

rhashimoto:
> I've thought about the heartbeat workaround too, my concern is that e.g. debuggers might pause a worker with a database lock for a while and then another tab could steal the lock.

You could write your watchdog to accept configuration changes via BroadcastChannel, then use that to override the missing heartbeat threshold for all tabs when you debug. You could send that message manually, or automatically with a browser extension content script on your development Safari (a content script also could be written to ask the user to approve breaking the lock).

I'm a little surprised your bug hasn't seen any real action yet. When I reported a [very similar bug](https://bugs.webkit.org/show_bug.cgi?id=239614) against WebKit a few years ago it got fixed the next day.

simolus3:
Interesting, the bug we're facing is essentially the combination of yours and [this one](https://bugs.webkit.org/show_bug.cgi?id=283528), but that took 8 months to see a PR. We'll see 🤞

PR #1032, opened 2026-07-08
[v1] fix(ts): some types polyshing, shorten filenames in web package dist/worker folder, shorten worker chunks names in dist folder

delagen:
retarget to v1 #1015

### DIFF
--- a/.changeset/tasty-comics-brush.md
+++ b/.changeset/tasty-comics-brush.md
@@ -1,0 +1,5 @@
+---
+'@powersync/common': patch
+---
+
+fix typescript error on ILogLevel types
--- a/packages/common/src/client/sync/stream/AbstractRemote.ts
+++ b/packages/common/src/client/sync/stream/AbstractRemote.ts
@@ -42,7 +42,7 @@ const KEEP_ALIVE_LIFETIME_MS = 90_000;
 /**
  * @internal
  */
-export const DEFAULT_REMOTE_LOGGER = Logger.get('PowerSyncRemote');
+export const DEFAULT_REMOTE_LOGGER: ILogger = Logger.get('PowerSyncRemote');
 
 /**
  * @internal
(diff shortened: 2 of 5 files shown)


packages/web/webpack.workers.config.js
```diff
@@ -18,6 +18,7 @@ export default () => {
     },
     output: {
       filename: 'worker/[name].umd.js',
+      chunkFilename: 'worker/[contenthash].umd.js',
```
simolus3: Why is this change necessary? I'd prefer not having any build changes on v1 to keep it safe.

simolus3 (commented):
Could you also run `pnpm -w changeset add` to add a changeset entry for `@powersync/common`? We don't really need one for `@powersync/web` since we have another change coming up.

packages/web/webpack.workers.config.js
delagen: It's just to compact this filenames
image

simolus3: We only use those for React Native Web, not worth optimizing that I believe. If it's still an issue with the updated build script on v2 then that's something we could look at.

delagen: They used in Web mode, and loaded from Worker automatically when I specified some VFS, including them in website looks ugly

simolus3: Unless you import the `./umd` module, it is your bundler which is responsible for bundling up web workers and their dependencies. The default export of the `@powersync/web` package is code in `./lib` which does not include these bundles directly.

The problem will be fixed in the v2 release, I don't think it's critical enough to adopt the old build logic.

delagen: > The problem will be fixed in the v2 release, I don't think it's critical enough to adopt the old build logic.

Remove this line?

delagen: Reverted this

simolus3 (approved):
Thanks!

PR #1033, opened 2026-07-09
[v1] fix for http clients getting stuck on 401

Chriztiaan:
Originally reported internally.

Fixed an issue where clients using the HTTP connection method could get stuck in an endless auth-error retry loop when their token expired while disconnected (e.g. after network interruptions or device sleep).

Unlike the WebSocket path and other HTTP requests, the HTTP sync stream did not invalidate cached credentials when receiving a 401 response, so every reconnect attempt reused the same expired token instead of fetching a fresh one from the connector. The sync stream now invalidates cached credentials on a 401, so the next retry calls `fetchCredentials()` for a fresh token.

## AI disclosure

This PR was created with the help of Claude Fable. Help constitutes assistance in research, planning, and rough outline of implementation. Beyond having a hand in the implementation, I have also manually tested this work.

### DIFF
--- a/.changeset/funny-buses-compete.md
+++ b/.changeset/funny-buses-compete.md
@@ -1,0 +1,5 @@
+---
+'@powersync/common': patch
+---
+
+Fixed an issue where clients using the HTTP connection method could get stuck in an endless auth-error retry loop when their token expired while disconnected (e.g. after network interruptions or device sleep).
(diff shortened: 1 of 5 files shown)


simolus3 (approved):
The fix looks good to me. It looks like the same issue currently exists on `main`, so we should probably fix it there as well.

simolus3 (approved):

PR #1039, opened 2026-07-14
feat(attachments): streaming attachment transport + saveFileFromUri

khawarizmus:
## Summary

Addresses the streaming request in discussion #968. The `AttachmentQueue` currently round-trips every file body through an `ArrayBuffer` in JS memory (`localStorage.readFile` → `remoteStorage.uploadFile`, and the reverse on download), which causes memory pressure / OOM for large files on lower-end devices.

This models the remote side (transfer **and** delete) as a single adapter so implementations can stream file-URI to the network natively, never materializing the file in the JS heap.

## What's included

- **`AttachmentTransportAdapter`** (`common`) — owns all remote operations: `upload`/`download`/`delete`. Supply this to fully replace `remoteStorage`.
- **`AttachmentQueue` options are now a discriminated union** — provide **exactly one** of `remoteStorage` or `transportAdapter`. Supplying both, or neither, is a compile-time type error. When only `remoteStorage` is given, an internal buffered transport is composed from it automatically, so existing setups are unchanged.
- **`saveFileFromUri`** — registers a file already on disk and queues it for upload without reading it into memory. Only available when the queue is configured with a `StreamingLocalStorageAdapter` (enforced in the type system).
- **`StreamingLocalStorageAdapter`** — a subinterface of `LocalStorageAdapter` adding buffer-free `moveFile`; implemented by the Node, Expo, and RN-FS adapters.
- **Native streaming transports** are created from the storage adapter via `createTransportAdapter(options)` (the transport classes are not exported directly):
  - **`NodeFileSystemAdapter`** (`node`) — streams via `fetch` + Node filesystem streams.
  - **`ExpoFileSystemStorageAdapter`** (`attachments-storage-react-native`) — uses Expo's `File.upload` / `File.downloadFileAsync`. Requires Expo SDK 56+ (`expo-file-system >=56`); the storage adapter itself still works on Expo SDK 54+.
  - **`ReactNativeFileSystemStorageAdapter`** (`attachments-storage-react-native`) — uses `@dr.pogodin/react-native-fs`'s `uploadFiles` / `downloadFile`.

## Backward compatibility

Default behavior is unchanged. When no `transportAdapter` is supplied, a buffered transport is composed internally from `localStorage`/`remoteStorage`, and `RemoteStorageAdapter` is unchanged. Existing `{ localStorage, remoteStorage }` setups still compile and behave identically.

One intended behavior change for the `RemoteStorageAdapter`: `downloadFile` now receives the record with its destination `localUri` populated (previously `null` for fresh downloads).

Generated with Claude, tested and reviewed manually.


packages/attachments-storage-react-native/src/ExpoFileSystemTransportAdapter.ts
```diff
@@ -0,0 +1,140 @@
+import type { AttachmentTransportAdapter, LocatedAttachmentRecord } from '@powersync/common';
+
+/**
+ * Describes the HTTP request used to upload a file's bytes to remote storage.
+ * Typically points at a presigned URL.
+ */
+export interface ExpoUploadRequest {
```
simolus3, line 13: Would it be possible to add an optional dependency on expo and then use expo types here?

packages/common/src/attachments/BufferedAttachmentTransport.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Default {@link AttachmentTransportAdapter}, composing the local and remote
+ * storage adapters.
+ *
+ * The full file body is materialized as an `ArrayBuffer` in JS memory between the
+ * two adapter calls. This is fine for small files but can cause memory pressure on
+ * large ones — environments that support native transfer should provide a custom
+ * transport instead.
+ *
+ * @experimental
+ * @alpha This is currently experimental and may change without a major version bump.
```
simolus3, line 16: Do we even need to export this from `common`? I assume users wouldn't construct this directly, they'd rely on this being the default right?

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -24,13 +26,31 @@ export interface AttachmentQueueOptions {
    */
   db: CommonPowerSyncDatabase;
   /**
-   * Remote storage adapter for upload/download operations
+   * Remote storage adapter for upload/download/delete operations. Used to build the
+   * default {@link BufferedAttachmentTransport}, which delegates all three operations
+   * to it.
+   *
+   * Required unless a {@link AttachmentQueueOptions.transportAdapter} is provided. If
```
simolus3: Can we encode this requirement in types? E.g. we could have `export type AttachmentQueueOptions = { ... remaining fields ... } & ({ remoteStorage: RemoteStorageAdapter } | { transportAdapter: AttachmentTransportAdapter })`

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -83,8 +103,8 @@ export class AttachmentQueue {
   /** Adapter for local file storage operations */
   readonly localStorage: LocalStorageAdapter;
 
-  /** Adapter for remote file storage operations */
-  readonly remoteStorage: RemoteStorageAdapter;
+  /** Adapter for remote file storage operations. Undefined when a custom `transportAdapter` is used. */
+  readonly remoteStorage?: RemoteStorageAdapter;
```
simolus3: `remoteStorage` is not used anywhere anymore. Given that attachments are still in beta and we're about to make a breaking change, let's just remove the field.

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -180,10 +201,16 @@ export class AttachmentQueue {
     this.downloadAttachments = downloadAttachments;
     this.logger = logger ?? db.logger;
     this.attachmentService = new AttachmentService(db, this.logger, tableName, archivedCacheLimit);
+
+    if (!transportAdapter && !remoteStorage) {
```
simolus3: We should also throw when both are set (or better yet make it a type error), because the remote implementation wouldn't be used in that case which sounds unintentional.

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -474,6 +501,83 @@ export class AttachmentQueue {
...
+   *
+   * Prefer this over {@link AttachmentQueue.saveFile} for large, app-originated files
+   * (recordings, videos): it avoids reading the file into an `ArrayBuffer` just to write
+   * it back to disk. Requires the local storage adapter to implement `moveFile`.
+   *
+   * @param options - File registration options
+   * @returns Promise resolving to the created attachment record
+   * @throws Error if the local storage adapter does not support moving files
+   */
+  async saveFileFromUri({
```
simolus3: Can we move common options into a shared interface and introduce a private method to implement `saveFile` and `saveFileFromUri` given that they're nearly the same?

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -474,6 +501,83 @@ export class AttachmentQueue {
...
+     */
+    id?: string;
+    /**
+     * Optional callback to execute additional database operations within the same transaction as the attachment
+     * creation.
+     */
+    updateHook?: (transaction: Transaction, attachment: AttachmentRecord) => Promise<void>;
+  }): Promise<AttachmentRecord> {
+    if (!this.localStorage.moveFile) {
+      throw new Error('The configured local storage adapter does not support moveFile, required by saveFileFromUri.');
```
simolus3: I would ideally also like this to be a static error.

I wonder if there's a way to make support for streaming downloads a subinterface of `LocalStorageAdapter`, and add a type parameter to `AttachmentQueue` to declare the type of local storage adapter used. Then we could use conditional types to declare that `saveFileFromUri` is only available when a stream-aware local adapter is used.

Additionally, it would also help with the option case above (we could require a `RemoteAdapter` if the local storage adapter is non-streaming only). Essentially, `AttachmentTransportAdapter` would then be a subinterface of `LocalStorageAdapter` and we'd add `moveFile` to it only.

simolus3 (commented):
I think the changes and the `BufferedAttachmentTransport` fallback are clean, but we can probably restructure the public interface a bit to get rid of runtime errors.

packages/attachments-storage-react-native/src/ExpoFileSystemTransportAdapter.ts
khawarizmus: I could not import the types from Expo. The dependency is already present.

packages/common/src/attachments/BufferedAttachmentTransport.ts
khawarizmus: That's true. I think there is no usage outside the common package for this. It's dropped.

packages/common/src/attachments/AttachmentQueue.ts
khawarizmus: This has been addressed now.

packages/common/src/attachments/AttachmentQueue.ts
khawarizmus: Removed. The remoteStorage option still works (it builds the default buffered transport), so apps that configure it are unaffected

packages/common/src/attachments/AttachmentQueue.ts
khawarizmus: Encoded this in the type now.

packages/common/src/attachments/AttachmentQueue.ts
khawarizmus: Made the functions more DRY by extracting the common logic.

packages/common/src/attachments/AttachmentQueue.ts
khawarizmus: This was addressed by introducing a new class `StreamingLocalStorageAdapter` which extends the `LocalStorageAdapter` and implements the `moveFile` function.

Now Node/Expo/RN all implement that class.

khawarizmus:
@simolus3 Tests that are outside the scope of the PR are failing. But the PR is ready.

khawarizmus:
@simolus3 CI is passing now.

packages/attachments-storage-react-native/src/ExpoFileSystemTransportAdapter.ts
simolus3: What errors did you get? We might need to tweak some typescript configs but copying the types doesn't sound right.

packages/attachments-storage-react-native/src/ExpoFileSystemTransportAdapter.ts
```diff
@@ -0,0 +1,150 @@
...
+  private fs: ExpoFileSystemLegacy;
+
+  constructor(private options: ExpoFileSystemTransportAdapterOptions) {
+    // `uploadAsync` / `downloadAsync` live in `expo-file-system/legacy` on SDK 54+,
+    // and in the main module on older SDKs. Metro only bundles static `require`
+    // string literals, so each candidate is required explicitly.
+    let resolved: ExpoFileSystemLegacy | undefined;
+    try {
+      const legacy = require('expo-file-system/legacy');
+      if (typeof legacy?.uploadAsync === 'function' && typeof legacy?.downloadAsync === 'function') {
```
simolus3: What are we guarding against here, `require` returning a bogus module? Shouldn't it throw if the import can't be resolved?

packages/attachments-storage-react-native/src/ExpoFileSystemTransportAdapter.ts
```diff
@@ -0,0 +1,150 @@
...
+ * on SDK 54+).
+ *
+ * @experimental
+ * @alpha This is currently experimental and may change without a major version bump.
+ */
+export class ExpoFileSystemTransportAdapter implements AttachmentTransportAdapter {
+  private fs: ExpoFileSystemLegacy;
+
+  constructor(private options: ExpoFileSystemTransportAdapterOptions) {
+    // `uploadAsync` / `downloadAsync` live in `expo-file-system/legacy` on SDK 54+,
```
simolus3: Given that this is a new functionality, I think we can also just make Expo 54+ a requirement for streaming transport.

That would then also let us use modern Expo APIs (`File.downloadAsync`) and avoid the deprecated import.

packages/attachments-storage-react-native/src/ReactNativeFSTransportAdapter.ts
```diff
@@ -0,0 +1,132 @@
...
+ * `uploadFiles` / `downloadFile`.
+ *
+ * The bytes never enter the JS heap, so large files can be transferred without the
+ * memory pressure of the buffer-based transport. Requires
+ * `@dr.pogodin/react-native-fs`.
+ *
+ * @experimental
+ * @alpha This is currently experimental and may change without a major version bump.
+ */
+export class ReactNativeFSTransportAdapter implements AttachmentTransportAdapter {
```
simolus3: Nit: Given that `ReactNativeFileSystemStorageAdapter` doesn't use abbreviations, I think this one shouldn't do that either.

packages/attachments-storage-react-native/src/ReactNativeFSTransportAdapter.ts
```diff
@@ -0,0 +1,132 @@
...
+  private rnfs: typeof import('@dr.pogodin/react-native-fs');
+
+  constructor(private options: ReactNativeFSTransportAdapterOptions) {
+    let rnfs: typeof import('@dr.pogodin/react-native-fs');
```
simolus3: `ReactNativeFileSystemStorageAdapter` has the same logic, can we extract it into a common helper?

Perhaps we also don't need to expose a second class for this at all, and `ReactNativeFileSystemStorageAdapter` could have a `createTransportAdapter(options: ReactNativeFSTransportAdapterOptions): AttachmentTransportAdapter` method to expose adapter (which would then be a class not exported from `index.ts`, and the storage adapter could pass its existing resolvede module to this).

(The same thing applies to the expo setup as well)

packages/common/src/attachments/AttachmentQueue.ts
simolus3: Can we also remove this check then? I don't really see the point in runtime checks for users who get the types wrong.

packages/common/src/attachments/AttachmentQueue.ts
```diff
@@ -474,6 +504,37 @@ export class AttachmentQueue {
...
+   * it back to disk. Requires the local storage adapter to implement `moveFile`.
+   *
+   * Only available when the queue is configured with a {@link StreamingLocalStorageAdapter}
+   * (one that implements `moveFile`).
+   *
+   * @param options - The existing file's `localUri` plus {@link SaveAttachmentOptions}
+   * @returns Promise resolving to the created attachment record
+   */
+  async saveFileFromUri(
+    this: AttachmentQueue<StreamingLocalStorageAdapter>,
```
simolus3, line 529: This is pretty neat, I didn't know this was possible 👍

packages/common/src/attachments/AttachmentTransportAdapter.ts
```diff
@@ -0,0 +1,49 @@
+import { AttachmentRecord } from './Schema.js';
+
+/**
+ * An {@link AttachmentRecord} that is guaranteed to have a `localUri`.
```
simolus3, line 4: Isn't `localUri` always going to be a path? If so, I'm not sure why we use the uri terminology instead of using file paths directly?

If we use URIs, we'd then have to guard against e.g. non-`file://`values. If we just say it's a file path, that should be easier to understand.

packages/node/src/attachments/NodeFileSystemAdapter.ts
```diff
@@ -1,12 +1,12 @@
...
 import * as path from 'path';
-import { AttachmentData, EncodingType, LocalStorageAdapter } from '@powersync/common';
+import { AttachmentData, EncodingType, StreamingLocalStorageAdapter } from '@powersync/common';
 
 /**
  * NodeFileSystemAdapter implements LocalStorageAdapter using Node.js filesystem.
  * Suitable for Node.js environments and Electron applications.
  */
-export class NodeFileSystemAdapter implements LocalStorageAdapter {
+export class NodeFileSystemAdapter implements StreamingLocalStorageAdapter {
```
simolus3, line 14: Should we also export a streaming transport helper for node that doesn't materialize the buffer?

To download, we can use `fs.createWriteStream().toWeb()`, then fetch the URL and call `.pipeTo()` to stream it to the file. The reverse should also work for uploads.

packages/attachments-storage-react-native/src/ExpoFileSystemTransportAdapter.ts
khawarizmus: This was removed.

packages/attachments-storage-react-native/src/ExpoFileSystemTransportAdapter.ts
khawarizmus: Refactored this to accommodate this. We now require v56+ for streaming

packages/attachments-storage-react-native/src/ReactNativeFSTransportAdapter.ts
khawarizmus: I have renamed FS across

packages/attachments-storage-react-native/src/ReactNativeFSTransportAdapter.ts
khawarizmus: Now both React Native file system and Expo use the same approach. We are not exporting another class and letting the storage adapter create the transport adapter

packages/common/src/attachments/AttachmentTransportAdapter.ts
khawarizmus: As discussed previously. This was skipped since it can introduce a breaking change.

packages/node/src/attachments/NodeFileSystemAdapter.ts
khawarizmus: Introduced a new Node streaming transport. It follows the same pattern as React native and Expo.

packages/node/src/attachments/NodeFileSystemTransportAdapter.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (mimeType) {
+      headers['Content-Type'] = mimeType;
+    }
+
+    const response = await fetch(request.url, {
+      method: request.httpMethod ?? 'PUT',
+      headers,
+      body: Readable.toWeb(createReadStream(attachment.localUri)) as ReadableStream,
+      duplex: 'half'
+    } as RequestInit & { duplex: 'half' });
```
simolus3, line 86: Maybe add a comment pointing to https://github.com/node-fetch/node-fetch/issues/1769 as an explanation.

packages/attachments-storage-react-native/README.md
```diff
@@ -59,43 +65,133 @@ const attachmentQueue = new AttachmentQueue({
...
+
+A transport adapter owns **all** remote operations (`upload` / `download` / `delete`) and transfers bytes natively — the file never enters the JS heap. Implement remote delete via the `deleteFile` callback.
+
+The queue takes **exactly one** remote mechanism: either `remoteStorage` or `transportAdapter`. A `transportAdapter` replaces `remoteStorage` entirely, so none is needed alongside it.
+
+Both transports are backend-agnostic: you supply resolver callbacks that map an attachment to a request (typically a presigned URL from your backend).
+
+### With Expo File System (`File.upload` / `File.downloadFileAsync`)
+
+> The Expo streaming transport requires **Expo SDK 56+** (`expo-file-system >=56`). The `ExpoFileSystemStorageAdapter` itself still works on Expo SDK 54+.
```
simolus3, line 97: It's odd to have this as a quote. If you want an alert, add a `> [!NOTE]` line above. Otherwise, just add it without the quote.

simolus3 (dismissed):
Looks good to me 👍

packages/attachments-storage-react-native/src/ExpoFileSystemTransportAdapter.ts
khawarizmus: This was resolved.

simolus3 (approved):

packages/attachments-storage-react-native/package.json
```diff
@@ -58,7 +58,7 @@
     "@rollup/plugin-node-resolve": "15.2.3",
     "@rollup/plugin-typescript": "^12.3.0",
     "@dr.pogodin/react-native-fs": "^2.36.1",
```
kobiebotha, line 60: @khawarizmus should this perhaps be bumped to `^2.39.1` ? See https://github.com/birdofpreyru/react-native-fs/pull/143

Issue #524, opened 2025-03-10
Kysely error


code-504:
I having this error:

```
node_modules\kysely\dist\esm\migration\file-migration-provider.js: node_modules\kysely\dist\esm\migration\file-migration-provider.js:Invalid call at line 32: import(/* webpackIgnore: true */(0, _classPrivateFieldLooseBase2.default)(this, _props)[_props].path.join((0, _classPrivateFieldLooseBase2.default)(this, _props)[_props].migrationFolder, fileName))
```

I apply #[this solution](https://github.com/powersync-ja/powersync-js/issues/262) but it doesn't work. Here is my `metro.config.js`:

```
const { getDefaultConfig } = require("expo/metro-config");
const { mergeConfig } = require("@react-native/metro-config");
const { withNativeWind } = require("nativewind/metro");
const path = require("node:path");

const projectRoot = __dirname;
const workspaceRoot = path.resolve(projectRoot, "../..");

const defaultConfig = getDefaultConfig(projectRoot);

const config = {
  resolver: {
    unstable_enablePackageExports: true,
    resolveRequest: (context, moduleName, platform) => {
      if (platform === "web") {
        if (
          ["react-native-prompt-android", "@powersync/react-native"].includes(
            moduleName,
          )
        ) {
          return {
            type: "empty",
          };
        }
        const mapping = {
          "react-native": "react-native-web",
          "@powersync/web": "@powersync/web/dist/index.umd.js",
          kysely: "kysely/dist/cjs/index.js",
        };
        if (mapping[moduleName]) {
          console.log("remapping", moduleName);
          return context.resolveRequest(context, mapping[moduleName], platform);
        }
      } else {
        if (["@powersync/web"].includes(moduleName)) {
          return {
            type: "empty",
          };
        }
      }

      return context.resolveRequest(context, moduleName, platform);
    },
    transformer: {
      getTransformOptions: async () => ({
        transform: {
          inlineRequires: {
            blockList: {
              [require.resolve("@powersync/react-native")]: true,
            },
          },
        },
      }),
    },
    watchFolders: [workspaceRoot],
    nodeModulesPaths: [
      path.resolve(projectRoot, "node_modules"),
      path.resolve(workspaceRoot, "node_modules"),
    ],
    disableHierarchicalLookup: true,
    unstable_enableSymlinks: true,
  },
};

const mergedConfig = mergeConfig(defaultConfig, config);

[...]

PR #1045, opened 2026-07-22
Update vitest and kysely

simolus3:
This updates vitest to 4.1.10 to avoid [this audit issue](https://github.com/advisories/GHSA-p63j-vcc4-9vmv). It also declares support for versions 0.29.x of kysely, which according to [this comment](https://github.com/powersync-ja/powersync-js/issues/524#issuecomment-4936862513) should fix some React Native issues.

### DIFF
--- a/packages/kysely-driver/tests/sqlite/sqlite-dialect.test.ts
+++ b/packages/kysely-driver/tests/sqlite/sqlite-dialect.test.ts
@@ -28,7 +28,7 @@ describe('PowerSyncDialect', () => {
       dialect: {
         createQueryCompiler: vitest.fn(),
         createDriver: vitest.fn(),
-        createAdapter: vitest.fn(),
+        createAdapter: () => new SqliteAdapter(),
         createIntrospector: vitest.fn()
       }
     });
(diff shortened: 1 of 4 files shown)


packages/kysely-driver/tests/sqlite/sqlite-dialect.test.ts
```diff
@@ -28,7 +28,7 @@ describe('PowerSyncDialect', () => {
       dialect: {
         createQueryCompiler: vitest.fn(),
         createDriver: vitest.fn(),
-        createAdapter: vitest.fn(),
+        createAdapter: () => new SqliteAdapter(),
```
simolus3, line 31: I'm not sure what exactly this was even supposed to test, but we can no longer return `undefined` as kysely queries the `supportsMultipleConnections` property on adapters.

Chriztiaan (approved):

pnpm-workspace.yaml
```diff
@@ -43,18 +43,18 @@ catalog:
   eslint-plugin-prettier: ^5.5.4
   glob: ^11.0.0
   jsdom: ^24.0.0
-  kysely: ^0.28.0
+  kysely: '^0.28.0 || ^0.29.0'
```
Chriztiaan, line 46: Is there a strong reason to keep 0.28 around?

simolus3: It's a dependency we publicly declare, so if someone wants to keep using 0.28.x with the latest `@powersync/kysely-driver`, they could.

PR #1047, opened 2026-07-22
Send sync status updates over broadcast channel when shared workers are disabled

simolus3:
To coordinate the sync process across tabs, we prefer to use shared workers. Shared workers are [broken](https://bugs.webkit.org/show_bug.cgi?id=318873) and disabled by default on Safari though, we also disable them on mobile. In these cases, each tab would attempt to sync on its own, with most tabs blocked on the navigator sync lock, leaving an app stuck waiting for `hasSynced`.

As a minor improvement, this makes the connected tab send its status update through a broadcast channel that other tabs trying to connect will listen to. So download progress and fields like `hasSynced` are then shared across tabs, and it appears like all tabs make sync progress similar to the setup with shared workers. Additionally, stream subscription changes from any tab are also broadcast to update the sync client in the active tab.

This solution is not perfect, being able to rely on shared workers would definitely be better. But it's still an improvement and provides limited multi-tab support even without shared workers.

### DIFF
--- a/.changeset/mighty-dodos-prove.md
+++ b/.changeset/mighty-dodos-prove.md
@@ -1,0 +1,7 @@
+---
+'@powersync/shared-internals': patch
+'@powersync/web': minor
+---
+
+When shared sync workers are disabled, use broadcast channels to share the current sync status and
+subscriptions across tabs.
--- a/.changeset/mighty-dodos-prove.md
+++ b/.changeset/mighty-dodos-prove.md
@@ -1,5 +1,5 @@
 ---
-'@powersync/shared-internals': patch
+'@powersync/shared-internals': minor
 '@powersync/web': minor
 ---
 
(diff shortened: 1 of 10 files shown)


.changeset/mighty-dodos-prove.md
```diff
@@ -0,0 +1,7 @@
+---
+'@powersync/shared-internals': patch
```
Chriztiaan: Since we are on 1.x.x for shared internals and we are changing an exported field to protected, maybe this should be minor? Or am I missing some context?

Chriztiaan (dismissed):

.changeset/mighty-dodos-prove.md
simolus3: I agree that a minor bump is better here.

Technically it doesn't matter since `shared-internals` is not a user-facing package and doesn't require semver (all packages depending on it would depend on the new version).

Chriztiaan (approved):

PR #1048, opened 2026-07-23
Document common package, include interfaces

simolus3:
This includes `@powersync/common` in the generated API reference, since it is a public package users can rely on. I wish we could make the `@powersync/common` re-exports in other packages link to those instead of being duplicated, but that seems to require a [different typedoc setup](https://typedoc.org/documents/Options.Input.html#packages) instead of invoking it once per independent package. Maybe something worth revisiting in the future.

It also includes interfaces for `@powersync/common` as dedicated pages, since we want to link some of those from the main documentation site.

### DIFF
--- a/docs/docusaurus.config.ts
+++ b/docs/docusaurus.config.ts
@@ -4,34 +4,44 @@ import type { PluginOptions } from 'docusaurus-plugin-typedoc';
 import 'dotenv/config';
 import { themes as prismThemes } from 'prism-react-renderer';
 import type { TypeDocOptionMap } from 'typedoc';
-import { DOC_FOLDER, packageMap } from './utils/packageMap';
+import { DOC_FOLDER, packageMap, Packages } from './utils/packageMap';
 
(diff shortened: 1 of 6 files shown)


docs/docusaurus.config.ts
```diff
@@ -4,34 +4,44 @@ import type { PluginOptions } from 'docusaurus-plugin-typedoc';
...
+      excludeInternal: true,
+      indexFormat: 'table',
+      disableSources: true,
+      expandObjects: true,
+      useCodeBlocks: true,
+      typeDeclarationFormat: 'table',
+      membersWithOwnFile: membersWithOwnFile
+      //      textContentMappings: {
+      //        'title.memberPage': '{name}'
+      //      }
```
simolus3: I have [reverted these changes](https://github.com/powersync-ja/powersync-js/commit/7785685a29ee82dc560d77edbe5fab47f0a2ebd2), I don't know why they're still showing here lol

Chriztiaan (approved):

PR #1051, opened 2026-07-28
Fix connect() transiently clobbering the offline-restored sync status

### LINKED ISSUE #1050: v2: connect() transiently clobbers the offline-restored SyncStatus (streams empty, hasSynced undefined) — useQuery waitForStream gates regress on warm boots
## Environment

- `@powersync/web` 2.0.0, `@powersync/common` 2.0.0, `@powersync/react` 2.0.0, `@powersync/shared-internals` 1.0.1
- `@journeyapps/wa-sqlite` 1.7.0, React 19.2, Vite
- OPFS storage, SharedWorker sync (defaults), HTTP connection method (the v2 default)
- Sync streams (`edition: 3` in the service config), one stream with `auto_subscribe: true` (`by_user`); every `useQuery` in the app passes `streams: [{ name: 'by_user', waitForStream: true }]`
- Chromium-based browsers
- **Regression from** `@powersync/web` 1.39.0 / `@powersync/common` 1.57.2 / `@powersync/react` 1.10.0 — same app code, differential builds (details at the bottom)

## Summary

On a **warm boot** (existing OPFS replica, existing session), the offline-restored `SyncStatus` — from `powersync_offline_sync_status()`, resolved during database init — correctly reports the auto-subscribed stream as synced before any query mounts. But ~30 ms into `connect()`, the client emits a `statusChanged` whose `SyncStatus` is **blank**: `connected: false`, `hasSynced: undefined`, empty stream list — `status.forStream({ name: 'by_user', parameters: null })` returns `undefined` for a stream the immediately preceding status had just reported as synced. The correct status is re-emitted one macrotask later, and the connected status follows.

That one blank emission is enough to regress every settled `useQuery` in the app to `{ isLoading: true, data: [] }` for a frame (see *Impact*), which unmounts all data-gated UI and produces a visible triple flash of the whole page on **every** warm page load. v1 (same app code) does not emit the blank status and loads clean.

## Observed `statusChanged` timeline (warm boot)

```text
[db open]            offline restore (powersync_offline_sync_status):
                     connected=false  hasSynced=true  lastSyncedAt=<restored>
                     forStream(by_user) → { subscription: { hasSynced: true, ... } }

[~30ms into connect]  ← the blip
                     connected=false  hasSynced=undefined  lastSyncedAt=undefined
                     stream list empty; forStream(by_user) → undefined

[+1 macrotask]       restored status again:
                     hasSynced=true; forStream(by_user) → { subscription: { hasSynced: true, ... } }

[shortly after]      connected=true, then normal stream progress
```

The blip and the restore land in **separate macrotasks**, so React commits the blank state — it is not batched away.

## Impact (via `@powersync/react`)

`useQuery` with `streams: [{ name, waitForStream: true }]` gates on `useAllSyncStreamsHaveSynced` (`src/hooks/streams.ts`):

```ts
return streams.every((stream, i) => {
  if (!stream.waitForStream) return true;
  return statuses[i]?.subscription?.hasSynced === true;
});
```

With the blank status, `forStream(...)` finds nothing → the gate closes → `useQuery` (`src/hooks/watched/useQuery.ts`) returns the loading state for a query that had long since settled:

```ts
if (!streamsHaveSynced) {
  return { ..._loadingState }; // isLoading: true, data: []
}
```

Two visible consequences, compounding:

1. Every settled read in the app flips to `isLoading && data === []` simultaneously → all data-driven UI unmounts at once (teardown 1).
2. The gate also drives the watched-query `active` flag, so the underlying watched queries are disposed and re-created when the gate reopens — passing through a second genuine `isLoading` pass (teardown 2).

Net effect: content appears, vanishes, reappears, vanishes, settles — three paints, two blanks, on every warm load. Anything gating on `useStatus().hasSynced` blips to `undefined` the same way.

## Where it appears to come from (v2 sources)

- `@powersync/shared-internals` `src/client/BasePowerSyncDatabase.ts`: init resolves the offline status (`SELECT powersync_offline_sync_status()`) and emits it — this part is correct, warm boots start synced. But the listener attached to a newly created sync implementation **replaces the database status wholesale**:

  ```ts
  statusChanged: (snapshot) => {
    this.currentStatus = snapshot;
    this.iterateListeners((cb) => cb.statusChanged?.(snapshot));
  }
  ```

- `@powersync/shared-internals` `src/client/sync/stream/AbstractStreamingSyncImplementation.ts`: every sync implementation created by `connect()` starts from a blank snapshot:

  ```ts
  this.syncStatus = new SyncStatusSnapshot(null, {});
  ```

  so whatever it emits before the Rust core's first `UpdateSyncStatus` instruction (which carries the real, DB-derived subscription state) is derived from `core = null` — no streams, `hasSynced` undefined. That early emission clobbers the offline-restored status at the database level, and the core's first real status restores it a macrotask later.

Notably, the disconnect path already guards against exactly this kind of state loss — `markAsDisconnected()` preserves `current?.streams ?? []` when it emits. The connect-time handoff has no equivalent: the new implementation's status starts empty instead of being seeded from the last known status.

## Expected behavior / suggestion

`connect()` should never observably "forget" persisted stream subscription state that the same client just restored from the database. Possible shapes, in whatever form fits the architecture best:

- Seed the freshly created sync implementation's initial `syncStatus` from the database's `currentStatus` at the `ConnectionManager.createSyncImplementation` handoff (mirroring what `markAsDisconnected` already does for streams), or
- In `BasePowerSyncDatabase`'s listener, merge rather than replace when the incoming snapshot has `core == null` (i.e. the implementation hasn't heard from the sync core yet), or
- Suppress emissions from a new sync implementation until the core has produced its first real status.

## Repro sketch

1. Web SDK with defaults (OPFS, SharedWorker sync), sync streams with one `auto_subscribe: true` stream; sign in and let the first sync complete.
2. Reload the page (warm boot — replica and session exist).
3. Log every `statusChanged`:

   ```ts
   db.registerListener({
     statusChanged: (s) =>
       console.log(performance.now(), {
         connected: s.connected,
         hasSynced: s.hasSynced,
         stream: s.forStream({ name: 'by_user', parameters: null })?.subscription?.hasSynced
       })
   });
   ```
4. Observe the blank emission between the offline-restored status and the reconnect. With `@powersync/react`, watch any `waitForStream: true` query flip `isLoading: true` / `data: []` during it.

## Regression evidence

Reproduced deterministically (not a race in app code): a production build of our app at the commit **before** the v2 migration (web 1.39.0 / common 1.57.2 / react 1.10.0) warm-loads clean 3/3 under an identical instrumented scenario (per-painted-frame sentinel); the first commit **after** the migration flashes 3/3. No SDK release newer than 2.0.0 exists to pick up at the time of writing.

Downstream we've worked around it by holding each query's last settled result in our `useQuery` wrapper and returning it during later `isLoading` flips of the same query — but the underlying status blip seems worth fixing at the source, since it affects any v2 app using `waitForStream: true` (or `useStatus().hasSynced`) on warm boots.

sincraianul:
## What & why

On a warm boot, `BasePowerSyncDatabase` restores the offline sync status (`powersync_offline_sync_status()`) during init, so returning clients correctly start with `hasSynced: true` and their persisted stream subscription state. But two status emissions from a freshly created sync stream implementation could reach the database's `statusChanged` listener before the sync core reported its first status for the connection attempt, transiently clobbering that restored state (full symptom timeline and downstream impact in #1050):

- The CRUD upload loop reports its upload state after the first pass, which can happen before the core's first `UpdateSyncStatus`. That snapshot carries `core == null`, and the listener replaced `currentStatus` wholesale — `hasSynced` flips to `undefined` and `forStream(...)` loses the subscription until the core's first status lands a task later.
- `markAsDisconnected()` on an implementation whose core never reported a status (an aborted or failed connection attempt) synthesized an empty but non-null core status (no streams, no priority entries), with the same effect.

## Changes

- `BasePowerSyncDatabase`: when an incoming snapshot has no core state while the database has one, keep the known core state and adopt the snapshot's `jsState` (the same merge idiom `resolveOfflineSyncStatus` already uses).
- `AbstractStreamingSyncImplementation.markAsDisconnected()`: return early when the core never reported a status for this connection attempt — there is no connection state to reset, and synthesizing an empty status would drop restored state. This extends the stream-preserving behavior `markAsDisconnected` already had for established connections.
- Two regression tests in `packages/node/tests/sync.test.ts` (run for both JSON and BSON): sync → close → reopen → connect, and the disconnect-during-connect variant. They hold the database write lock while connecting, which blocks the core's first status (`powersync_control` needs the write connection) while the upload loop's initial empty-queue pass is read-only — making the originally ~30 ms race deterministic. Both fail on `main` (`hasSynced` becomes `undefined`/`false`) and pass with this change.

## Notes for reviewers

- The database-level merge also covers the web SDK: worker-forwarded statuses (`SharedWebStreamingSyncImplementation`) and broadcast-channel statuses (`TabLocalStreamingSyncImplementation`) enter the tab through `updateSyncStatus(core, dataFlow)` and land in the same listener.
- The guard only affects the pre-first-core-status window: no code path resets a snapshot's `core` back to null after the core has reported once.

Fixes #1050

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/eight-jokes-refuse.md
+++ b/.changeset/eight-jokes-refuse.md
@@ -5,4 +5,4 @@
 Fix `connect()` transiently reporting a sync status without core state (`hasSynced: undefined`, no
 stream subscriptions), clobbering the offline sync status restored when the database was opened.
 Also fix aborted or failed connection attempts resetting the restored status the same way before
-the sync core had reported a status for them.
+the core extension had reported a status for them.
--- a/.changeset/eight-jokes-refuse.md
+++ b/.changeset/eight-jokes-refuse.md
@@ -1,0 +1,8 @@
+---
+'@powersync/shared-internals': patch
+---
+
+Fix `connect()` transiently reporting a sync status without core state (`hasSynced: undefined`, no
+stream subscriptions), clobbering the offline sync status restored when the database was opened.
+Also fix aborted or failed connection attempts resetting the restored status the same way before
+the sync core had reported a status for them.
--- a/packages/shared-internals/src/client/BasePowerSyncDatabase.ts
+++ b/packages/shared-internals/src/client/BasePowerSyncDatabase.ts
@@ -167,8 +167,17 @@ export abstract class BasePowerSyncDatabase<Options extends BasePowerSyncDatabas
           const sync = this.generateSyncStreamImplementation(connector, options);
           const onDispose = sync.registerListener({
             statusChanged: (snapshot) => {
-              this.currentStatus = snapshot;
-              this.iterateListeners((cb) => cb.statusChanged?.(snapshot));
+              // A new sync stream implementation can report JavaScript-side state (like upload
+              // progress) before the sync core has reported its first status. Those snapshots
+              // have no core state - keep the known core state (e.g. the offline sync status
+              // resolved during initialization) instead of clobbering it with an empty one.
+              const updatedStatus =
+                snapshot.core == null && this.currentStatus.core != null
+                  ? new SyncStatusSnapshot(this.currentStatus.core, snapshot.jsState)
+                  : snapshot;
+
+              this.currentStatus = updatedStatus;
+              this.iterateListeners((cb) => cb.statusChanged?.(updatedStatus));
             }
           });
           await sync.waitForReady();
--- a/packages/shared-internals/src/client/BasePowerSyncDatabase.ts
+++ b/packages/shared-internals/src/client/BasePowerSyncDatabase.ts
@@ -167,10 +167,7 @@ export abstract class BasePowerSyncDatabase<Options extends BasePowerSyncDatabas
           const sync = this.generateSyncStreamImplementation(connector, options);
           const onDispose = sync.registerListener({
             statusChanged: (snapshot) => {
-              // A new sync stream implementation can report JavaScript-side state (like upload
-              // progress) before the sync core has reported its first status. Those snapshots
-              // have no core state - keep the known core state (e.g. the offline sync status
-              // resolved during initialization) instead of clobbering it with an empty one.
+              // For a JavaScriptSyncState update before the sync client was able to resolve the full status from the core extension, use the known offline sync state resolved during initialization.
               const updatedStatus =
                 snapshot.core == null && this.currentStatus.core != null
                   ? new SyncStatusSnapshot(this.currentStatus.core, snapshot.jsState)
--- a/packages/shared-internals/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
+++ b/packages/shared-internals/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
@@ -336,9 +336,7 @@ The next upload iteration will be delayed.`
   private markAsDisconnected() {
     const current = this.syncStatus.core;
     if (current == null) {
-      // The sync core never reported a status for this connection attempt, so there is no
-      // connection state to reset. Synthesizing an empty core status here would clobber sync
-      // state previously restored from the database.
+      // There's no connection attempt for which state could be cleared.
       return;
     }
 
--- a/packages/shared-internals/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
+++ b/packages/shared-internals/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
@@ -335,12 +335,19 @@ The next upload iteration will be delayed.`
 
   private markAsDisconnected() {
     const current = this.syncStatus.core;
+    if (current == null) {
+      // The sync core never reported a status for this connection attempt, so there is no
+      // connection state to reset. Synthesizing an empty core status here would clobber sync
+      // state previously restored from the database.
+      return;
+    }
+
     this.updateSyncStatus({
       connected: false,
       connecting: false,
-      priority_status: current?.priority_status ?? [],
+      priority_status: current.priority_status,
       downloading: null,
-      streams: current?.streams ?? []
+      streams: current.streams
     });
   }
 
--- a/packages/node/tests/sync.test.ts
+++ b/packages/node/tests/sync.test.ts
@@ -6,13 +6,16 @@ import {
   ProgressWithOperations,
   Schema,
   SyncOptions,
+  SyncStatus,
   SyncStreamConnectionMethod
 } from '@powersync/common';
 import {
   bucket,
+  checkpoint,
   MockSyncService,
   createMockSyncServiceTest,
   mockSyncServiceTest,
+  stream,
   TestConnector,
   waitForSyncStatus
 } from './utils.js';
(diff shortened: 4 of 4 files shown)


packages/shared-internals/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -335,12 +335,19 @@ The next upload iteration will be delayed.`
 
   private markAsDisconnected() {
     const current = this.syncStatus.core;
+    if (current == null) {
+      // The sync core never reported a status for this connection attempt, so there is no
+      // connection state to reset. Synthesizing an empty core status here would clobber sync
+      // state previously restored from the database.
+      return;
```
simolus3: nit: make less verbose

```suggestion
      // There's no connection attempt for which state could be cleared.
      return;
```

packages/shared-internals/src/client/BasePowerSyncDatabase.ts
```diff
@@ -167,8 +167,17 @@ export abstract class BasePowerSyncDatabase<Options extends BasePowerSyncDatabas
           const sync = this.generateSyncStreamImplementation(connector, options);
           const onDispose = sync.registerListener({
             statusChanged: (snapshot) => {
-              this.currentStatus = snapshot;
-              this.iterateListeners((cb) => cb.statusChanged?.(snapshot));
+              // A new sync stream implementation can report JavaScript-side state (like upload
+              // progress) before the sync core has reported its first status. Those snapshots
+              // have no core state - keep the known core state (e.g. the offline sync status
+              // resolved during initialization) instead of clobbering it with an empty one.
```
simolus3: nit: We typically use "core extension" instead of "sync core".

```suggestion
              // For a JavaScriptSyncState update before the sync client was able to resolve the full status from the core extension, use the known offline sync state resolved during initialization.
```

simolus3 (commented):
Thanks for the contribution! The change looks good to me, I just have two nits to reduce verbosity and to make comments more accurate.

packages/shared-internals/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
sincraianul: Applied the suggested comment verbatim in 60e55f5c.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/shared-internals/src/client/BasePowerSyncDatabase.ts
sincraianul: Applied the suggested wording ("core extension") verbatim in 60e55f5c.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.changeset/eight-jokes-refuse.md
```diff
@@ -0,0 +1,8 @@
+---
+'@powersync/shared-internals': patch
+---
+
+Fix `connect()` transiently reporting a sync status without core state (`hasSynced: undefined`, no
+stream subscriptions), clobbering the offline sync status restored when the database was opened.
+Also fix aborted or failed connection attempts resetting the restored status the same way before
+the sync core had reported a status for them.
```
Chriztiaan: > We typically use "core extension" instead of "sync core".

Chriztiaan: There's 3 more `sync core` references left in your diff :)

sincraianul: Updated the changeset to say "core extension" in 1e18a021.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

simolus3 (approved):

PR #1057, opened 2026-07-30
Fix watched queries leaking listeners when closed during initialization

### LINKED ISSUE #1056: Every React useQuery mount permanently leaks a tablesUpdated adapter listener (close-during-init race in watched queries)
### Description

If a `WatchedQuery` is closed while its initialization is still awaiting, it permanently leaks listeners. `@powersync/react`'s `useQuery` triggers this race on **every mount**, so apps leak one `tablesUpdated` DB-adapter listener per mounted query. On our React Native app this degrades performance minute-over-minute: each table change fans out to an ever-growing set of dead listeners (each scheduling query re-execution machinery) that can never be removed.

Confirmed on `main` (2.0.0 / `@powersync/shared-internals` 1.0.1) and on the `v1` branch (`@powersync/common` 1.57.2 — same code in `packages/common/src/client/AbstractPowerSyncDatabase.ts:1298`). We originally isolated it on 1.57.0.

### The race (file refs against current `main`)

`packages/react/src/hooks/watched/useWatchedQuery.ts:37` creates the watched query inside a `useState` initializer during render, and the mount effect (lines 40-49) immediately `close()`s that instance and creates a replacement. One `create → close-one-frame-later` cycle per mount, while the watcher's async init is still in flight. Three gaps then compound:

1. **`AbstractQueryProcessor.init`** (`packages/shared-internals/src/client/watched/AbstractQueryProcessor.ts:158-186`): registers the `closing` listener, `await db.waitForReady()`, registers `schemaChanged`, and only then assigns `this.disposeListeners`. A `close()` during the await runs `this.disposeListeners?.()` while it is still `null` → the `closing` listener leaks, and `schemaChanged` is registered *after* the query is closed and leaks too.
2. **`OnChangeQueryProcessor.linkQuery`** (`packages/shared-internals/src/client/watched/OnChangeQueryProcessor.ts:42-46`) and **`DifferentialQueryProcessor.linkQuery`** (`DifferentialQueryProcessor.ts:146-160`): after `await db.resolveTables(...)` there is no `this.closed || abortSignal.aborted` check, and the dispose function returned by `db.onChangeWithCallback(...)` is discarded — cleanup relies entirely on the signal, which by now is already aborted.
3. **`BasePowerSyncDatabase.onChangeWithCallback`** (`packages/shared-internals/src/client/BasePowerSyncDatabase.ts:770-812`): registers `this.database.registerListener({ tablesUpdated: ... })` unconditionally (line 799) and wires cleanup only via `resolvedOptions.signal?.addEventListener('abort', ...)` (line 806). An **already-aborted** signal never dispatches `abort` → the adapter listener is orphaned forever.

(`watchWithCallback`, `BasePowerSyncDatabase.ts:699-716`, has the same already-aborted-signal shape for the deprecated watch API.)

### Repro (vitest, distilled)

```ts
// db.database is a real BaseObserver-backed DBAdapter fake exposing listeners.size
const controller = new AbortController();
controller.abort();

db.onChangeWithCallback({ onChange: () => {} }, { signal: controller.signal, tables: ['todos'] });
expect(adapter.listeners.size).toBe(0); // FAILS on main: 1, and nothing can ever remove it

// End-to-end (the useQuery mount shape): park resolveTables on the adapter's read lock,
// close the watch mid-await, then release:
const query = db
  .customQuery({ compile: () => ({ sql: 'SELECT 1', parameters: [] }), execute: async () => [] })
  .watch({});
await tick();
await query.close();
releaseReadLock();
await settle();
expect(adapter.listeners.size).toBe(0); // FAILS on main: 1 leaked tablesUpdated listener
```

### Expected

Closing a watched query at any point of its lifecycle — including while init is awaiting `waitForReady()` or `resolveTables()` — removes every listener it registered. Passing an already-aborted signal to `onChangeWithCallback` registers nothing.

### Fix

A PR accompanies this issue: an early no-op-dispose return in `onChangeWithCallback` for already-aborted signals, closed/aborted bail-outs after the awaited `resolveTables` in both `linkQuery` implementations, and disposal of the `closing` listener when `close()` lands during `init`'s `waitForReady()` — plus regression tests that fail on current `main`. A backport to `v1` would be appreciated, as 1.57.x has the identical defect.

Thorsson:
## Overview

Closing a `WatchedQuery` while its initialization is still in flight permanently leaks listeners. `@powersync/react`'s `useQuery` hits this race on every mount, so React/React Native apps leak one `tablesUpdated` adapter listener per mounted query and degrade over time.

Three compounding gaps, innermost first:

1. `BasePowerSyncDatabase.onChangeWithCallback` (`packages/shared-internals/src/client/BasePowerSyncDatabase.ts:770`) registers `this.database.registerListener({ tablesUpdated })` unconditionally and cleans up only via `resolvedOptions.signal?.addEventListener('abort', ...)`. An already-aborted signal never emits `abort`, so the adapter listener can never be removed.
2. `OnChangeQueryProcessor.linkQuery` / `DifferentialQueryProcessor.linkQuery` `await db.resolveTables(...)` and then call `db.onChangeWithCallback(...)` without re-checking `this.closed || abortSignal.aborted`, discarding the returned dispose function — relying entirely on the broken already-aborted-signal path above.
3. `AbstractQueryProcessor.init` registers the `closing` listener, awaits `db.waitForReady()`, registers `schemaChanged`, and only then assigns `this.disposeListeners`. A `close()` that lands during the await finds `disposeListeners === null`, so both database listeners leak (`schemaChanged` is even registered after the query is already closed).

The trigger in React: `useWatchedQuery` creates the watched query in a `useState` initializer during render and `close()`s it from the mount effect one frame later, while init is still parked on `waitForReady()`/`resolveTables()`.

## Changes

- `onChangeWithCallback` returns a no-op dispose immediately when the signal is already aborted, before allocating the executor or registering the adapter listener.
- Both `linkQuery` implementations bail after the awaited `resolveTables` when the query was closed or aborted.
- `AbstractQueryProcessor.init` disposes the `closing` listener and returns when `close()` happened during `waitForReady()`.
- Regression tests in `packages/shared-internals/tests/client/watched/listenerLeaks.test.ts` assert real `BaseObserver` listener counts across the abort/close races; the three leak tests fail on current `main`.

Fixes #1056.

### DIFF
--- a/packages/shared-internals/tests/client/watched/listenerLeaks.test.ts
+++ b/packages/shared-internals/tests/client/watched/listenerLeaks.test.ts
@@ -1,0 +1,195 @@
+import { DBAdapter, LockContext, RawQueryResult, Schema } from '@powersync/common';
+import { describe, expect, it } from 'vitest';
+import { BasePowerSyncDatabase } from '../../../src/client/BasePowerSyncDatabase.js';
+
+class MockLockContext extends LockContext {
+  async executeRaw(): Promise<RawQueryResult> {
+    return { columnNames: [], rawRows: [] };
+  }
+}
+
+class MockDBAdapter extends DBAdapter {
+  /**
+   * Read locks await this promise before executing.
+   * Tests use this to park a query (e.g. `resolveTables`) mid-await.
+   */
+  readGate: Promise<void> = Promise.resolve();
+
+  get name() {
+    return 'mock-db';
+  }
+
+  /**
+   * The number of registered adapter listeners. Exposed for leak assertions.
+   */
+  get registeredListenerCount() {
+    return this.listeners.size;
+  }
+
+  async close() {}
+
+  async refreshSchema() {}
+
+  async readLock<T>(fn: (tx: LockContext) => Promise<T>): Promise<T> {
+    await this.readGate;
+    return fn(new MockLockContext());
+  }
+
+  async writeLock<T>(fn: (tx: LockContext) => Promise<T>): Promise<T> {
+    return fn(new MockLockContext());
+  }
+}
+
+class TestPowerSyncDatabase extends BasePowerSyncDatabase {
+  /**
+   * When set, `waitForReady` awaits this promise.
+   * Tests use this to park watched-query initialization mid-await.
+   */
+  readyGate: Promise<void> | null = null;
+
+  /**
+   * The number of registered database listeners. Exposed for leak assertions.
+   */
+  get registeredListenerCount() {
+    return this.listeners.size;
+  }
+
+  get mockAdapter() {
+    return this.database as MockDBAdapter;
+  }
+
+  protected openDBAdapter(): DBAdapter {
+    return new MockDBAdapter();
+  }
+
+  protected generateSyncStreamImplementation(): never {
+    throw new Error('Sync is not required for these tests');
+  }
+
+  protected generateBucketStorageAdapter(): never {
+    return null as never;
+  }
+
+  protected async _initialize() {}
+
+  // The real initialization runs queries against the SQLite database, which is not required here.
+  protected async initialize() {
+    this.ready = true;
+  }
+
+  async waitForReady(): Promise<void> {
+    await this.readyGate;
+  }
+}
+
+const createTestDatabase = () => new TestPowerSyncDatabase({ schema: new Schema([]) });
+
+const watchQuery = (db: TestPowerSyncDatabase) =>
+  db
+    .customQuery<{ id: string }>({
+      compile: () => ({ sql: 'SELECT * FROM todos', parameters: [] }),
+      execute: async () => []
+    })
+    .watch({});
+
+const deferred = () => {
+  let resolve!: () => void;
+  const promise = new Promise<void>((r) => {
+    resolve = r;
+  });
+  return { promise, resolve };
+};
+
+/** Waits for pending micro and macro tasks to complete. */
+const settle = async () => {
+  for (let i = 0; i < 10; i++) {
+    await new Promise((r) => setTimeout(r));
+  }
+};
+
+describe('watched query listener cleanup', () => {
+  it('onChangeWithCallback should not register an adapter listener for an already-aborted signal', () => {
+    const db = createTestDatabase();
+    const controller = new AbortController();
+    controller.abort();
+
+    const dispose = db.onChangeWithCallback({ onChange: () => {} }, { signal: controller.signal, tables: ['todos'] });
+
+    expect(db.mockAdapter.registeredListenerCount).toEqual(0);
+    // The returned dispose function should still be safe to call
+    dispose();
+    expect(db.mockAdapter.registeredListenerCount).toEqual(0);
+  });
+
+  it('onChangeWithCallback should dispose the adapter listener when the signal aborts', () => {
+    const db = createTestDatabase();
+    const controller = new AbortController();
+
+    db.onChangeWithCallback({ onChange: () => {} }, { signal: controller.signal, tables: ['todos'] });
+
+    expect(db.mockAdapter.registeredListenerCount).toEqual(1);
+    controller.abort();
+    expect(db.mockAdapter.registeredListenerCount).toEqual(0);
+  });
+
+  it('should register and dispose listeners across a watched query lifecycle', async () => {
+    const db = createTestDatabase();
+    const baseListenerCount = db.registeredListenerCount;
+
+    const query = watchQuery(db);
+    await settle();
+
+    expect(db.mockAdapter.registeredListenerCount).toEqual(1);
+    // The closing and schemaChanged listeners
+    expect(db.registeredListenerCount).toEqual(baseListenerCount + 2);
+
+    await query.close();
+    await settle();
+
+    expect(db.mockAdapter.registeredListenerCount).toEqual(0);
+    expect(db.registeredListenerCount).toEqual(baseListenerCount);
+  });
+
+  it('should not leak listeners when a watched query is closed while resolving tables', async () => {
+    const db = createTestDatabase();
+    const baseListenerCount = db.registeredListenerCount;
+
+    // Park the query's `resolveTables` call on its read lock
+    const readGate = deferred();
+    db.mockAdapter.readGate = readGate.promise;
+
+    // Mimics React's useQuery, which creates a watched query during render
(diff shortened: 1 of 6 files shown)


packages/shared-internals/tests/client/watched/listenerLeaks.test.ts
```diff
@@ -0,0 +1,195 @@
+import { DBAdapter, LockContext, RawQueryResult, Schema } from '@powersync/common';
+import { describe, expect, it } from 'vitest';
+import { BasePowerSyncDatabase } from '../../../src/client/BasePowerSyncDatabase.js';
+
+class MockLockContext extends LockContext {
```
simolus3, line 5: Instead of testing against a mock database, can we write regression tests against a real instance? E.g. in `packages/react/tests/useQuery.test.tsx` which runs them against the web SDK?

simolus3 (commented):
Thank you for the contribution!

> `OnChangeQueryProcessor.linkQuery` / `DifferentialQueryProcessor.linkQuery` `await db.resolveTables(...)` and then call `db.onChangeWithCallback(...)` without re-checking `this.closed || abortSignal.aborted`, discarding the returned dispose function — relying entirely on the broken already-aborted-signal path above.

But now that that's fixed, do we still need these changes? Relying on the abort signal to be both accurate and actually honored by `onChangeWithCallback` feels like the most elegant solution to me, so unless there's another bug here the `linkQuery` changes feel superfluous?

rachitsrivastava07:
We hit this in a production build on our beta deployment, used by real people, and can confirm the diagnosis from a heap snapshot, plus a real-SDK browser reproduction. It is severe for long-lived tabs, so we'd love to see this land.

**What we saw.** Our versions are `@powersync/web` 2.4.1, `@powersync/react` 2.0.1, `@powersync/common` 2.3.0 and `@powersync/shared-internals` 1.3.0, with React 18.3.1. After a long session, one tab's JS heap was 1,018 MB. The database's `listeners` Set held 44,244 `closing` and 44,245 `schemaChanged` listeners. They belonged to 74,536 query processors, each with `_closed: true`, `state.isLoading: true`, and `disposeListeners` set. The adapter held 8,323 `tablesUpdated` listeners, 7,821 of them owned by closed processors, while only 326 processors were still open. Everything dominated by that Set came to 545 MB, 54 % of the heap.

**Why each leak costs so much in Chromium.** `close()` calls `this.abortController.abort()`, which creates a `DOMException`. V8 keeps the frames of an exception's unformatted stack alive until `.stack` is read. One of those frames is `useWatchedQuery`'s mount effect, and its closure holds the `setWatchedQuery` dispatch, which is bound to the component's fiber. So every leaked processor pins its whole unmounted component tree, not just itself. The stack captured in the heap was:
`flushSyncCallbacks → performSyncWorkOnRoot → commitRootImpl → flushPassiveEffects → commitHookEffectListMount → useWatchedQuery mount effect → close()`.
Separately from this fix, passing a shared reason to `abort()` would stop each `close()` from capturing a stack. We haven't checked what that would mean for anything that inspects `signal.reason`.

**The two windows, both reproduced in real browsers.** We tested Chromium on `OPFSWriteAheadVFS`, and Firefox and WebKit on `OPFSCoopSyncVFS`, with production builds. The app mounts 20 `useQuery` components and replaces them on each click, transition or timer update.
1. A close while `init()` awaits `waitForReady()` leaks about 1 `closing` and 1 `schemaChanged` per mounted `useQuery`, on every mount and in all three lanes. Over 410 remounts that came to 8,242 of each in Chromium and WebKit, and 8,220 in Firefox. The Chromium heap grew from 2.4 MB to 30.3 MB.
2. A close while `resolveTables` is pending leaks `tablesUpdated` on the adapter. We triggered it by holding the write lock so reads queue on a single-connection VFS. Adapter listeners went from 44 to 444 in Chromium and WebKit, and to 564 in Firefox.

**On the review question.** We agree the `linkQuery` changes are unnecessary once `onChangeWithCallback` honours an already-aborted signal. We ship only these two hunks, as a pnpm patch on `@powersync/shared-internals@1.3.0`:

```js
// AbstractQueryProcessor.init()
await db.waitForReady();
if (this._closed) {
    disposeCloseListener();
    return;
}
```

```js
// BasePowerSyncDatabase.onChangeWithCallback()
const resolvedOptions = options ?? {};
if (resolvedOptions.signal?.aborted) {
    return () => { };
}
```

With that patch, listener counts stayed at the live count in all three browsers, across every lane and both windows. Live updates behaved the same as before: same tab, across tabs through the shared worker, offline, after swapping to a new database, and on `close()`. `db.close()` also got faster, because it awaits each `closing` listener one at a time.

**Standalone repro (zip attached at the end of this comment).** It uses the published packages only, at the versions above. There's no StrictMode, no sync connection and no backend: 20 components each call the stock `useQuery`. Run `npm ci && npm start` and watch the listener panel while you click "Next page ×10" then "Hide all". Alternatively, `node run.mjs` builds it both stock and with the two guards, then counts live processors via CDP `Runtime.queryObjects` after forced GC, with nothing mounted:

| Build | Mounts | Processors alive | `closing` | `schemaChanged` |
|---|---|---|---|---|
| stock | 220 | 440 | 220 | 221 |
| stock | 2,020 | 4,040 | 2,020 | 2,021 |
| with the two guards | 220 | 40 | 0 | 1 |
| with the two guards | 2,020 | 40 | 0 | 1 |

Stock leaks exactly two processors and one listener pair per mount. With the guards the count stays flat.

**On testing against a real instance.** A regression test in `packages/react/tests/useQuery.test.tsx` could:
- render a component that calls `useQuery` inside a synchronous update, such as `act()` around a click or `flushSync`;
- unmount it and mount a new one N times;
- assert that `db.listeners.size` and the adapter's `tablesUpdated` count return to their baseline.

The first window is deterministic under a sync update, because React flushes passive effects inside the same microtask as the commit. For the second, delay `resolveTables`, for example by holding the write lock on a single-connection VFS, then unmount before it resolves. We're happy to contribute that test if it helps.

A related `@powersync/react` issue, which this PR doesn't touch: when React discards a render, the processor built in `useWatchedQuery`'s `useState` initializer is never closed. StrictMode in dev does this on every mount. Filed as #1119.

[powersync-leak-repro.zip](https://github.com/user-attachments/files/32620270/powersync-leak-repro.zip)

PR #1059, opened 2026-08-04
Port recent web fixes to v1 branch

simolus3:
This applies two recent PRs (https://github.com/powersync-ja/powersync-js/pull/1053 fixing minor concurrency issues and https://github.com/powersync-ja/powersync-js/pull/1054 allowing longer file names) to the `v1` branch.

After this is merged, I'll manually apply `changesets version` and open a follow-up PR to release this.


.github/workflows/audit.yaml
```diff
@@ -1,25 +0,0 @@
-name: Audit
```
simolus3, line 1: Getting this to pass requires upgrading vitest, which is not happening on the `v1` branch.

Chriztiaan (approved):

PR #1065, opened 2026-08-11
Open read-only connections in parallel

simolus3:
For the OPFS write-ahead filesystem implementation, we allow spawning additional workers hosting read-only connections to a database file. These additional workers are currently opened one after another. In some cases, the setup step for workers can be expensive: They need to scan a write-ahead log file to compute their database page overlay, and we've seen cases after large transactions where the WAL is very large.

This still opens the primary (read-write) worker before any other, but refactors the opening process to open read workers in parallel. I have verified these changes with a small demo opening a database with a large WAL and a varying amount of read workers. There's still a opening time increase when adding additional workers, but the factor is well below 1.

### DIFF
--- a/.changeset/dirty-trees-add.md
+++ b/.changeset/dirty-trees-add.md
@@ -1,0 +1,5 @@
+---
+'@powersync/web': patch
+---
+
+Allow opening read workers for the OPFS WriteAhead file system in parallel, improving startup performance.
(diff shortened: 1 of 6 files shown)


Chriztiaan (approved):
Looks reasonable, I ran the failing tests locally and they passed.

PR #1073, opened 2026-08-18
Explicit checkpoint requests

simolus3:
As a follow-up to #1072 which implements [request checkpoint protocol changes](https://github.com/orgs/powersync-ja/discussions/317), this adds the [explicit sync API](https://github.com/orgs/powersync-ja/discussions/324).

Added methods closely match the ones in the Swift SDK: `PowerSyncDatabase.requestCheckpoint()` requests a checkpoint (throwing if we're not connected). On checkpoints, `hasSynced` reports whether a checkpoint has synced and `waitForSync` can be used to wait for it. There's no `waitForSync(timeout:)` equivalent, passing an `AbortSignal.timeout()` instead is too obvious.

A notable difference is that the Swift SDK adds typed errors for different checkpoint failure modes, where this currently uses a plain `Error` class. This isn't great, but matches what we do in most other parts of the SDK too. We can revisit better/typed error reports in a future update.

This is adopted in two demos, using an opt-in environment flag in both cases:

1. In the React Supabase todolist demo, there's a refresh button next to the connectivity indicator when connecting. It spins while we're waiting for a checkpoint, a snackbar displays results.
2. In the React Native Supabase todolist demo, a swipe-to-refresh pattern is used to request checkpoints.

AI use: Changes in demos are mostly generated with Claude Code, I also used it to review these changes and made minor improvements in response.

### DIFF
--- a/demos/react-supabase-todolist/src/app/views/layout.tsx
+++ b/demos/react-supabase-todolist/src/app/views/layout.tsx
@@ -1,15 +1,19 @@
 import { LOGIN_ROUTE, SQL_CONSOLE_ROUTE, TODO_LISTS_ROUTE } from '@/app/router';
 import { useNavigationPanel } from '@/components/navigation/NavigationPanelContext';
-import { useSupabase } from '@/components/providers/SystemProvider';
+import { syncOptions, useCheckpointRequests, useSupabase } from '@/components/providers/SystemProvider';
 import ChecklistRtlIcon from '@mui/icons-material/ChecklistRtl';
 import ExitToAppIcon from '@mui/icons-material/ExitToApp';
 import MenuIcon from '@mui/icons-material/Menu';
 import NorthIcon from '@mui/icons-material/North';
+import RefreshIconBase from '@mui/icons-material/Refresh';
 import SignalWifiOffIcon from '@mui/icons-material/SignalWifiOff';
 import SouthIcon from '@mui/icons-material/South';
 import TerminalIcon from '@mui/icons-material/Terminal';
 import WifiIcon from '@mui/icons-material/Wifi';
+import { css, keyframes } from '@emotion/react';
 import {
+  Alert,
+  AlertColor,
   AppBar,
   Box,
   Divider,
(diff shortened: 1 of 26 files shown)


packages/common/src/client/sync/CheckpointRequest.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Use this value to wait until the local database has applied server-side changes up to the requested checkpoint. This
+ * is useful for explicit refresh flows where the caller wants confirmation that the local view has caught up to the
+ * service.
+ *
+ * Checkpoint requests are backed by request ids tracked in the local database, so they are reusable across disconnect
+ * and reconnect cycles. A wait interrupted by a disconnect throws an error, but the same request can be awaited again
+ * once a new connection is established.
+ *
+ * Requests do to surive {@link CommonPowerSyncDatabase#disconnectAndClear}, instances created before a clear should be
```
Chriztiaan: typo?

packages/common/src/client/CommonPowerSyncDatabase.ts
```diff
@@ -233,6 +234,19 @@ export interface CommonPowerSyncDatabase extends BaseObserverInterface<PowerSync
    */
   syncStream(name: string, params?: Record<string, any>): SyncStream;
 
+  /**
+   * Requests a checkpoint from thte PowerSync service.
```
Chriztiaan: typo

demos/react-supabase-todolist/src/app/views/layout.tsx
```diff
@@ -40,6 +47,28 @@ export default function ViewsLayout({ children }: { children: React.ReactNode })
...
+  const snackbarStatus = useMemo<{ severity: AlertColor; message: string } | undefined>(() => {
+    if (explicitRefresh == 'success') {
+      return { severity: 'success', message: 'Synced successfully!' };
+    } else if (explicitRefresh == 'error') {
+      return { severity: 'error', message: 'Sync checkpoint failed' };
+    }
+  }, [explicitRefresh]);
+  const abortSignal = useAbortSignal();
+
+  async function explicitSync() {
```
Chriztiaan, line 60: This is awesome, straight to the docs.

Chriztiaan (approved):

PR #1076, opened 2026-08-18
React Native: Add Swift Package Manager support

simolus3:
React Native version 0.87.0 added experimental support for [Swift Package Manager](https://reactnative.dev/blog/2026/08/11/react-native-0.87) to build iOS apps.

This adds `ios/Package.swift` with an empty Swift target depending on the core extension to ensure those builds will bundle it. The main challenge here was figuring out how to name the package. By default, autolinking in React Native uses the unqualified part of a package name and shortens scopes around hyphens. So `@react-native/foo` would be named `RNFoo`, for example. Unfortunately, the expected default name for `@powersync/react-native` is simply `ReactNative`, which conflicts with React Native itself.

To fix the package name picked up by autolinking, we can write a `react-native.config.js` file. These options appear to be undocumented and actively contradict the design doc for SwiftPM integration in React Native. Additionally, some components of the React Native build process read that file via `await import()` while others `require()` it, which requires additional workarounds. So the whole thing is a huge mess, but it works and with React Native, that's as good as it gets.

I have tested these changes by upgrading `tools/powersynctests` to React Native 0.87.0 and running iOS and Android tests locally. To ensure the `react-native.config.js` change doesn't break older apps, I've also tested `react-native-supabase-todolist` which keeps using React Native 0.86.0 and CocoaPods.

AI use: Scanned build files with Claude Code, used that to fix some minor issues.


packages/react-native/ios/swiftpm/empty.swift
```diff
@@ -0,0 +1,2 @@
+// Deliberately left empty: We need a SwiftPM target with sources, but the only responsibility of
+// the build is to link the core extension.
```
rkistner, line 2: Just for my understanding: This package contains `PowerSyncOpSqlite.mm` and `PowerSyncOpSqlite.h` - where do they fit in with the Swift package?

And if the only purpose of this Swift package is to depend on `powersync-sqlite-core-swift`, I assume there isn't a way to have that dependency directly instead of via this intermediate SwiftPM package? I.e. have a structure of `app -> powersync-sqlite-core-swift` instead of `app -> PowerSyncReactNative -> powersync-sqlite-core-swift`

simolus3: > Just for my understanding: This package contains `PowerSyncOpSqlite.mm` and `PowerSyncOpSqlite.h` - where do they fit in with the Swift package?

They're not included in that build (but empty anyways). We might be able to remove them from the CocoaPods build too, but I didn't want to touch that build setup here and I'm not sure if CocoaPods has a different validation logic for missing source files.

> I assume there isn't a way to have that dependency directly instead of via this intermediate SwiftPM package?

According to [this](https://github.com/chrfalch/discussions-and-proposals/blob/chrfalch/add-rfc-draft-for-spm/proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md#third-party-libraries), we should even be able to depend on a precompiled XCFramework directly. That doesn't appear to be implemented yet, though. Removing the Swift Package makes React Native complain about us only supporting CocoaPods, for example.

Chriztiaan (approved):

PR #1078, opened 2026-08-20
Fix stream subscription changes made during connect()

joshuabrink:
## What happens

A `subscribe()` or `unsubscribe()` called, while `connect()` is bringing up the sync implementation, is never sent to the PowerSync Service. It returns normally and the client reports the new stream set, but the service keeps the set from when `connect()` started. The sets stay out of step until the set changes again or the connection restarts.

## Why

`packages/shared-internals/src/client/ConnectionManager.ts`

```ts
// connectInternal(), :228
const { sync, onDispose } = await this.options.createSyncImplementation(connector, {
  subscriptions: this.activeStreams,        // (1) the set is read here
  serializedSchema: schema
});
this.syncStreamImplementation = sync;       // (3) only now can (2) reach the implementation

// :358
private subscriptionsMayHaveChanged() {
  this.syncStreamImplementation?.updateSubscriptions(this.activeStreams);   // (2) silent no-op
}
```

A change landing between (1) and (3) is lost: Too late for the set read at (1), and at (2) `syncStreamImplementation` is still null, so `?.` skips the call without raising anything. 

## Fix

Save the identity of the set it read at (1), and then after (3) check for a difference, if true then update the stream set.

## AI Disclosure

Claude Opus was used to build the [reproduction](https://github.com/joshuabrink/powersync-js/tree/repro/connect-window-subscription-drop/demos/repro-connect-window-subscription) and tests. Issue and fix manually verified.

### DIFF
--- a/.changeset/brave-otters-listen.md
+++ b/.changeset/brave-otters-listen.md
@@ -2,7 +2,6 @@
 '@powersync/shared-internals': patch
 ---
 
-Fix sync stream subscription changes being lost when they happen while `connect()` is still creating the sync
-implementation. A `subscribe()` or `unsubscribe()` in that window was silently dropped and never retried, leaving a
-stream syncing after nothing was subscribed to it, or a stream that the client reports as subscribed but which is never
-requested (so no data arrives and `waitForFirstSync()` never resolves).
+Fix stream subscription changes made while `connect()` is still bringing up the sync implementation being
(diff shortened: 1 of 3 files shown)


simolus3 (approved):
In our other SDKs, we solve this issue by providing both the current snapshot and a channel emitting update notifications. That feels like a cleaner solution, but would require many more changes so I'm happy with this fix.

PR #1084, opened 2026-08-24
fix(react): forward throttleMs from useQuery to the watched query

### LINKED ISSUE #1082: useQuery accepts `throttleMs` and silently discards it
`useQuery` accepts `throttleMs` in its options type but never passes it down, so you always get the 30ms default. Nothing warns.

The options object it builds for `useWatchedQuery` is just:

```js
{
  reportFetching: options.reportFetching,
  rowComparator: options.rowComparator
}
```

`useWatchedQuery` does read `hookOptions.throttleMs` (in `differentialWatch`, `watch`, and `updateSettings`), so it's only that literal that's missing it.

```js
// types fine, still throttles at 30ms
const { data } = useQuery('SELECT * FROM lists', [], { throttleMs: 200 });
```

Hit this on `@powersync/react@2.0.0`, and `packages/react/src/hooks/watched/useQuery.ts` on main looks the same today. Also reproduced on 1.10.0, so it isn't new.

While I was in there I noticed the hook's options type comes from the callback API rather than the watched-query one:

```ts
HookWatchOptions extends Omit<SQLOnChangeOptions, 'signal'>
```

so `tables` and `triggerImmediate` type-check too, and are also dropped. Those two have no equivalent in `WatchedQueryOptions` so they could never have applied. `triggerOnTables` — the one that would actually limit trigger tables — isn't on the hook type at all. `throttleMs` seems to be the only one of the four that's both advertised and implementable. The type lineage might be worth a separate look, though changing it would break anyone passing the dead keys.

Two small things I ran into:

Passing `throttleMs` isn't entirely inert. `checkQueryChanged` does `JSON.stringify(options)` on the raw user options, so changing it between renders flips `queryChanged` and fires an `updateSettings({ throttleMs: undefined })`.

`useSuspenseQuery` is unaffected — `QueryStore.getQuery` forwards `throttleMs` properly.

The fix looks like one line:

```diff
     options: {
       reportFetching: options.reportFetching,
+      throttleMs: options.throttleMs,
       rowComparator: options.rowComparator
     },
```

Workaround for anyone who finds this first, since both exports are public:

```js
const lists = db.query({ sql, parameters }).differentialWatch({ rowComparator, throttleMs: 200 });
const { data } = useWatchedQuerySubscription(lists);
```

icole:
Fixes #1082.

`useQuery` accepts `throttleMs` in its options type, but the object it builds for `useWatchedQuery` only carries `reportFetching` and `rowComparator`, so the value never arrives and you always get the 30ms default.

`useWatchedQuery` already reads `hookOptions.throttleMs` in three places, so it's just the one missing property.

The other thing I mentioned in the issue is left alone here: `tables` and `triggerImmediate` also type-check on the hook and are also dropped, and `triggerOnTables` isn't reachable through it at all. That's the options type inheriting from the callback API rather than the watched-query one, and untangling it is a breaking change, so it seemed like your call rather than something to fold in.

### DIFF
--- a/.changeset/olive-moons-tickle.md
+++ b/.changeset/olive-moons-tickle.md
@@ -1,0 +1,5 @@
+---
+'@powersync/react': patch
+---
+
+Forward `throttleMs` from `useQuery` to the underlying watched query. It was accepted by the options type but never passed on, so the default 30ms throttle always applied.
--- a/packages/react/src/hooks/watched/useQuery.ts
+++ b/packages/react/src/hooks/watched/useQuery.ts
@@ -80,6 +80,7 @@ export function useQuery<RowType = any>(
     queryChanged,
     options: {
       reportFetching: options.reportFetching,
+      throttleMs: options.throttleMs,
       // Maintains backwards compatibility with previous versions
       // Differentiation is opt-in by default
       // We emit new data for each table change by default.
--- a/packages/react/tests/useQuery.test.tsx
+++ b/packages/react/tests/useQuery.test.tsx
@@ -933,6 +933,80 @@ describe('useQuery', () => {
         expect(finalState.isFetching).toEqual(false);
         expect(finalState.isLoading).toEqual(false);
       });
+
+      it('should forward throttleMs to the watched query', async () => {
+        const db = openPowerSync();
+        const throttleMs = 1234;
+
+        // Spy on the watched query construction to assert the option reaches it.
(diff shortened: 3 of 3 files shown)


simolus3 (approved):

simolus3 (approved):
Thanks! Yeah, there are lots of broken types in `packages/react` and we'll probably need a breaking release to sweep through them at some point.

I agree with this nice and immediate fix.

PR #1087, opened 2026-08-28
fix(attachments-storage-react-native): await File.move before reading size in moveFile

icole:
`ExpoFileSystemStorageAdapter.moveFile` fires off `File.prototype.move()` without awaiting it, then reads `.size` on the very next line. `move` is an async native call dispatched off the JS thread, while `.size` is a synchronous property read, so the read happens before the relocation has completed. `size` resolves to `null` for a file it can't read yet, and `moveFile`'s own `?? 0` turns that into a `0` returned for a file that goes on to move correctly.

It's reachable through the public API: `AttachmentQueue.saveFileFromUri` calls straight into this for any `StreamingLocalStorageAdapter`. The two sibling adapters (`ReactNativeFileSystemStorageAdapter`, `NodeFileSystemAdapter`) already await their move/rename before stat'ing, so this brings the Expo adapter in line with them.

The fix is one line: add the missing `await`. A changeset is included.

### DIFF
--- a/.changeset/fix-move-file-race.md
+++ b/.changeset/fix-move-file-race.md
@@ -1,0 +1,5 @@
+---
+'@powersync/attachments-storage-react-native': patch
+---
+
+Fix `ExpoFileSystemStorageAdapter.moveFile` returning a stale size.
--- a/packages/attachments-storage-react-native/src/ExpoFileSystemStorageAdapter.ts
+++ b/packages/attachments-storage-react-native/src/ExpoFileSystemStorageAdapter.ts
@@ -101,7 +101,7 @@ To use the Expo File System attachment adapter please install expo-file-system (
       if (target.exists) {
         target.delete();
       }
-      new this.File(sourceUri).move(target);
+      await new this.File(sourceUri).move(target);
     }
     return new this.File(targetUri).size ?? 0;
   }


.changeset/fix-move-file-race.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+Fix `ExpoFileSystemStorageAdapter.moveFile` returning a stale size. The
+promise from `File.prototype.move` was not awaited before reading `.size`
+on the destination file, so for files large enough that the underlying OS
+relocation takes measurable time, the returned size could read back as `0`
+(coerced from `size`'s `null`-when-unreadable result) while the file itself
+went on to relocate successfully. `AttachmentQueue.saveFileFromUri` — the
+public entry point that reaches this method for any `StreamingLocalStorageAdapter`
+— could therefore record a stale `0` size for a fully-transferred file.
```
simolus3: The rest of this is not relevant for users and likely wrong (`moveFile` is atomic as long as only one file system is involved, so it doesn't depend on file size, it's just a plain race condition depending on how native calls are scheduled to threads).

```suggestion
Fix `ExpoFileSystemStorageAdapter.moveFile` returning a stale size.
```

packages/attachments-storage-react-native/src/ExpoFileSystemStorageAdapter.ts
```diff
@@ -19,13 +19,23 @@ export class ExpoFileSystemStorageAdapter implements StreamingLocalStorageAdapte
   private Directory: typeof Directory;
   private storageDir: Directory;
 
-  constructor(storageDirectory?: string) {
+  /**
```
simolus3: Instead of adding extra complexity here for tests, couldn't we use vitest to temporarily mock the `require()` function?

packages/attachments-storage-react-native/tests/ExpoFileSystemStorageAdapter.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * A minimal in-memory fake of expo-file-system's `File`/`Directory` API - just enough
+ * surface for `ExpoFileSystemStorageAdapter` to run against under plain Node, without
+ * Expo's native bindings. `moveImpl` is reassignable per test so a relocation's
+ * completion can be held open under test control, which is what makes the race in
+ * `moveFile` observable deterministically rather than by chance timing.
+ */
+function createFakeFileSystem() {
+  const fileStore = new Map<string, Uint8Array>();
```
simolus3: This is overkill for a test that just calls `moveFile`, return a stub file constructor that only implements `size` and `move` using vitest mock utilities instead of an in-memory VFS.

(IMO, given how trivial this change is we don't necessarily need tests at all)

simolus3 (approved):

icole:
@simolus3 Thanks for the feedback here!

simolus3:
Of course, and thank you for contributing this fix 👍

PR #1088, opened 2026-08-31
Remove optimizeDeps.exclude from vite.config.ts

KazimirPodolski:
I'm on Vite v8.8.2, and apparently the issues with web workers were fixed, because it just works for me without the `optimizeDeps: {exclude: ['@powersync/web']}`. Perhaps the requirement can now be removed from the example? (And the docs, but I didn't find any.)

### DIFF
--- a/demos/example-vite/vite.config.ts
+++ b/demos/example-vite/vite.config.ts
@@ -12,11 +12,6 @@ export default defineConfig({
     emptyOutDir: true
   },
   envDir: '..', // Use this dir for env vars, not 'src'.
-  optimizeDeps: {
-    // Don't optimize these packages as they contain web workers and WASM files.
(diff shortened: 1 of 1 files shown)


simolus3 (commented):
Searching for `optimizeDeps` reveals that we have this pattern in a lot more demos and the diagnostic app too. We need to keep it in vitest config files as an optimized dependency causes a test reload that then fails vitest, but if it works here then it should work in other demos too.

Could you apply this to the other excludes as well? I'm happy to help smoke test changed demos then. We would likely also have to update vite in those demos then.

KazimirPodolski:
If it's not trivial, I would just leave it as it is now. My priorities are currently elsewere, sorry.

PR #1092, opened 2026-09-06
fix(common): stop expireCache looping with no archived attachments

giaBaoJS:
### The bug

`AttachmentQueue.expireCache()` never resolves when the number of archived attachments is at or below `archivedCacheLimit`, which is the normal case. It keeps re-running the same `SELECT` against the attachments table forever.

### Reproduction

```ts
const queue = new AttachmentQueue({ db, localStorage, remoteStorage, watchAttachments });
// No archived attachments (or fewer than archivedCacheLimit, which defaults to 100).
await queue.expireCache(); // never resolves; the same query runs in a tight loop
```

### Root cause

`AttachmentContext.deleteArchivedAttachments` returns a "cleanup finished" flag: the last line returns `archivedAttachments.length < limit`, so a partial page means there is nothing left to page through. The early return for an empty page returned `false` instead, which is the opposite: an empty page is the smallest possible partial page. `AttachmentQueue.expireCache` loops `while (!isDone)` on that flag, so the empty case spins forever.

The other caller, `AttachmentQueue.syncStorage`, ignores the return value, which is why this has not shown up there.

For comparison, the same operation in the older `@powersync/attachments` package (`AbstractAttachmentQueue.expireCache`) returns immediately when the query comes back empty.

### What the tests assert

- `AttachmentContext > deleteArchivedAttachments > reports completion when there is nothing to delete`: `deleteArchivedAttachments()` returns `true` for an empty page, and neither the callback nor the delete statement runs. Without the fix this fails with `expected false to be true`.
- `AttachmentContext > deleteArchivedAttachments > reports completion after deleting a partial page`: pins the existing behaviour that `true` means "done" for a page smaller than the internal limit.
- `AttachmentQueue > expireCache > stops once there is nothing archived left to delete`: `expireCache()` resolves after a single query. The fake database throws after a few queries so the non-terminating case fails as an assertion rather than hanging the suite.

`pnpm --filter @powersync/common test`: 27 tests passing before, 30 after.


packages/common/tests/attachments/AttachmentContext.test.ts
```diff
@@ -0,0 +1,51 @@
+import { describe, expect, it, vi } from 'vitest';
```
simolus3: Unless it makes them much more complicated, could you move these tests to e.g. the `node` package and run them against a real database using `databaseTest`?

.changeset/fix-expire-cache-loop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/common': patch
+---
+
+Fix `AttachmentQueue.expireCache()` never returning when there is nothing to delete. `AttachmentContext.deleteArchivedAttachments` reported that it was not finished for an empty page, so the pagination loop in `expireCache` kept re-running the same query indefinitely.
```
simolus3: No need to mention implementation details in changelogs


```suggestion
Fix `AttachmentQueue.expireCache()` never returning when there is nothing to delete.
```

simolus3 (commented):
Thanks for your contribution!

.changeset/fix-expire-cache-loop.md
giaBaoJS: Applied your suggestion, the changeset is now just that one line.

packages/common/tests/attachments/AttachmentContext.test.ts
giaBaoJS: Moved. They now live in `packages/node/tests/attachments-expire-cache.test.ts` and run against a real database.

One small deviation: I used `customDatabaseTest({ schema })` rather than `databaseTest` directly, because the default `AppSchema` in `tests/utils.ts` has no `AttachmentTable` for the queue to write to. Same fixture otherwise.

There are four tests now: two go through `queue.withAttachmentContext((ctx) => ctx.deleteArchivedAttachments(...))`, and two call `expireCache()`. The `expireCache` ones resolve a description of how the call ended instead of awaiting it directly, so reverting the fix fails with `expected 'still running after 2000ms' to be 'returned'` rather than a bare test timeout.

I also removed the mocked `AttachmentQueue` test in `common`, since the node tests cover that path now.

simolus3 (approved):

PR #1094, opened 2026-09-09
Universal PowerSync Diagnostics (POC)

khawarizmus:
This PR implements a live-attach diagnostics tool built once and reused across different hosts. It shows the live state of the PowerSync client side.

## How it works

The agent runs next to your database. It reads the database and pushes snapshots. The UI rebuilds its state from these snapshots. On the web, the agent runs in your app page. Vite DevTools loads it after you trust the tab. On node, the agent runs in the same process as the database. A React Native app connects to the standalone window over the network (RN has not been tested yet).

The MCP endpoint exposes the same functions as the tools, allowing AI agents to directly query the client app and retrieve diagnostic information.

## New packages

- `@powersync/diagnostics-core`. The protocol. It is one interface, `SdkIntegration`, plus the data shapes. The UI talks only to this interface. Each SDK implements it. The package also has the JavaScript agent and the iframe bridge.
- `@powersync/diagnostics-ui`. The UI. Rendered from an `SdkIntegration`. It runs in an iframe or in a standalone window.
- `@powersync/diagnostics`. The host integration, built on [devframe](https://devfra.me/). It gives a dock in Vite DevTools, a tab in Nuxt DevTools, a dev server for Node apps, a standalone window via CLI (`powersync-devtools`), and MCP tools for AI agents.

## SDK changes

- `@powersync/common`. a `diagnostics` option on `connect()`. It turns on the SQLite core diagnostics stream. This gives pre-bucket download stats and inferred column types.
- `@powersync/shared-internals`: the sync client keeps the connector for the whole connection. It forwards the core diagnostics events.
- `@powersync/web`. a new `@powersync/web/devtools` subpath. It lists the open databases in the page. The agent finds them without any app code.
- `@powersync/nuxt`. **breaking**. `NuxtPowerSyncDatabase` is removed. Use `PowerSyncDatabase`  from `@powersync/web` and pass `connect(connector, { diagnostics: true })`. The module shows the PowerSync tab on Nuxt DevTools 3 and the dock on Nuxt DevTools 4.

## Tested by hand

- Nuxt 4.5 with Nuxt DevTools 3 and with Nuxt DevTools 4 (Vite devtools). With and without `compatibilityVersion: 5`.
- React with Vite 8.3 (`devtools: true`) and with Vite 7 (`@vitejs/devtools` plugin).
- The standalone window (using React with Vite): no app tab, one tab, several tabs, closing tabs, reload.
- MCP from curl and from Claude, on the Vite dev server and on the Node server.
- A Node app with its own DevTools window.
- The CLI window with a pre-shared token, from a Node process.
- Production builds of the React app and the Nuxt demo not shipping any diagnostics code in the production build.

Claude code was used to steer the discovery and development. The code was reviewed, tested and modified manually.


simolus3 (changes requested):
To share more context here (I've also shared that offline with more context @khawarizmus). I think the idea is awesome, my initial architectural complaints are:

1. It's not clear whether we need a browser extension, so remove that from this PR which is already very big (you can open a follow-up PR for that separately).
2. I think a JSON-based protocol over "any two way channel" is the wrong abstraction here. It e.g. forces us to re-invent an RPC protocol when some implementations we'll use (say the Dart VM service protocol) are already RPC based.
3. None of this should require any changes to an existing SDK package (aside from forwarding core diagnostics events, but even that should likely be a separate PR to validate that the basic integration can work without SDK changes). We can likely inject an agent by writing a vite devtools package instead. Or initially, we could also expose it as a seperate package a user would depend on to expose the channel (e.g. a `enablePowerSyncDiagnostics(port: MessagePort, db: BasePowerSyncDatabase)`).
4. Related to 2 and 3, the interface the diagnostics tool expects an SDK to provide should be an actual TypeScript interface with async methods and event listeners. It is the responsibility of each integration (vite, `postMessage` iframe, Dart) to figure out how to implement that interface by doing the serialization itself.

packages/diagnostics-core/src/bridge.ts
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * The UI always runs in an iframe; the integration lives on the other side (the app page, or a host
+ * that reaches the app another way). comlink turns the interface's method calls into messages and
+ * correlates request/response itself, so the tool carries no wire format of its own. Each pairing
+ * uses one dedicated `MessagePort`, so several UIs can attach to one integration without sharing a
+ * channel.
+ */
+
+/** The shape exposed over comlink: `observeEvents` takes a proxied callback instead of a function. */
+type Remote = Omit<SdkIntegration, 'observeEvents'> & {
```
simolus3: Let's just make `observeEvents` return a promise on `SdkIntegration`, or a `Promise<void> | void`.

packages/diagnostics-core/src/index.ts
```diff
@@ -0,0 +1,13 @@
...
+export * from './integration.js';
+
+// The JavaScript integration (runs in the app page) and the structural database it reads.
+export * from './live-database.js';
+export * from './agent.js';
+export * from './state.js';
+
+// Moving an integration across an iframe boundary, and deriving UI state from its events.
+export * from './bridge.js';
+export * from './store.js';
```
simolus3: Relying on structural typing to avoid pulling in the JS SDK here is clever, but also feels somewhat fragile.

Given that these files are still only useful when providing diagnostics for JavaScript SDKs, maybe they shouldn't be part of the default export. I think putting them into a separate package is overkill, but maybe they should be exported from a separate entrypoint (say `@powersync/diagnostics-core-js-web`).

(and then if they're separate exports, I wouldn't mind this package having an optional dependency on `@powersync/common` and defining the JS agent on `BasePowerSyncDatabase`).

packages/diagnostics-core/PROTOCOL.md
```diff
@@ -0,0 +1,195 @@
...
+│   UI  ──calls──▶  SdkIntegration  ◀──implemented by──  SDK side  │
+└──────────────────────────────────────────────────────────────────┘
+```
+
+- The **UI** calls `SdkIntegration` methods and renders what comes back.
+- Each **SDK** provides one implementation. It reaches the live client with whatever that environment already has.
+
+The tool never carries a wire format of its own. Request/response correlation, connection and reconnection are the implementation's concern.
+
+## 2. The interface
```
simolus3: Having the overview here is helpful, but it could also be part of the readme. Sections 2-5 of this document should be tsdoc comments instead.

packages/shared-internals/src/client/sync/stream/core-instruction.ts
```diff
@@ -16,7 +16,29 @@ export type NonInterruptingInstruction =
   | { UpdateSyncStatus: UpdateSyncStatus }
   | { FetchCredentials: FetchCredentials }
   | { FlushFileSystem: any }
-  | { DidCompleteSync: any };
+  | { DidCompleteSync: any }
+  | { HandleDiagnostics: DiagnosticsEvent };
+
+/**
+ * Emitted by the core extension when diagnostics are enabled on the sync stream (see
+ * {@link https://github.com/powersync-ja/powersync-sqlite-core diagnostics}). Reports detailed
```
simolus3: Should probably link to https://github.com/powersync-ja/powersync-sqlite-core/blob/main/crates/core/src/sync/diagnostics.rs directly?

simolus3 (commented):
I didn't take a look at the UI yet, but the architecture of how devtools are served and integrated looks pretty good at this point. I'll also try to scaffold a Dart integration over the vm service protocol.

packages/diagnostics-ui/playground/mockDatabase.ts
```diff
@@ -0,0 +1,391 @@
...
+        data: JSON.stringify({ description: `Row ${i}`, completed: i % 2, priority: i % 4 })
+      })) as T[];
+    }
+    if (/ps_buckets/i.test(sql)) {
+      return [
+        { name: 'user_tasks[]', ops: 142, size: 48213, last_op: '1042' },
+        { name: 'global[]', ops: 18, size: 3120, last_op: '88' }
+      ] as T[];
+    }
+    if (/count\(\*\)[\s\S]*ps_crud/i.test(sql)) {
```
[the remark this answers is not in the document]
khawarizmus: This is resolved. Although it's not relevant since this code was in the playground

packages/diagnostics-core/src/bridge.ts
khawarizmus: This was addressed

packages/diagnostics-core/PROTOCOL.md
khawarizmus: This was addressed. The relevant parts are now in the README, and the rest has been ported to tsdocs comments

packages/shared-internals/src/client/sync/stream/core-instruction.ts
khawarizmus: This is done

packages/diagnostics-core/src/index.ts
khawarizmus: The JS-specific code now lives in `@powersync/diagnostics-core/js`. I kept the structural type instead of adding an optional `@powersync/common` dependency. The Vite client can now pass WebPowerSyncDatabase directly, and only the Playground UI mock needs to implement that small interface.

packages/diagnostics-ui/src/composables/theme.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Shared, persisted light/dark theme for the diagnostics UI.
+ *
+ * The panel applies `.dark` to its own root from this state, so it owns its theme rather than
+ * inheriting a host-forced class. Defaults to the host/OS preference and can be overridden per
+ * viewer (persisted in localStorage). A host that follows its own theme (Nuxt DevTools, the Chrome
+ * panel) can push it in via `setTheme(...)`.
+ */
+const STORAGE_KEY = 'powersync-diagnostics-theme';
```
simolus3, line 11: I feel like host-controlled dark mode should also disable the builtin theme persistence, could we have perhaps something like `provideDiagnostics` that lets an embedder provide the current theme + changes and skips everything else?

For Flutter DevTools, we inherit the theming preference from the outer devtools frame and shouldn't show our own toggle.

packages/diagnostics-core/src/shapes.ts
```diff
@@ -0,0 +1,237 @@
...
+  params?: unknown[];
+}
+
+/** A serialized SQL result set. */
+export interface QueryResult {
+  /** Column names, in order. */
+  columns: string[];
+  /** One object per row, keyed by column name. */
+  rows: Record<string, unknown>[];
+  rowCount: number;
```
simolus3, line 29: Isn't this the same as `rows.length`?

packages/diagnostics-core/src/shapes.ts
```diff
@@ -0,0 +1,237 @@
...
+  sql: string;
+  params?: unknown[];
+}
+
+/** A serialized SQL result set. */
+export interface QueryResult {
+  /** Column names, in order. */
+  columns: string[];
+  /** One object per row, keyed by column name. */
+  rows: Record<string, unknown>[];
```
simolus3, line 28: Given that we have column names on this object, I think an `unknown[][]` representation might be more efficient (representing each row as an array, `columns` can still be used to turn this into an object).

packages/diagnostics-core/src/integration.ts
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * When each event is emitted:
+ *
+ * | `type`        | `payload`              | When                                                        |
+ * | ------------- | ---------------------- | ----------------------------------------------------------- |
+ * | `status`      | {@link SyncState}      | The sync status changes.                                    |
+ * | `streams`     | {@link StreamState}[]  | The sync status changes.                                    |
+ * | `buckets`     | {@link BucketState}[]  | An internal table changes, or a core event updates a total. |
+ * | `uploadQueue` | {@link UploadQueueState} | The sync status or `ps_crud` changes.                     |
+ * | `logs`        | {@link LogRecord}[]    | The SDK logs. Not replayed to late subscribers.             |
```
simolus3, line 31: Maybe it should be called something like `new_logs` then.

packages/diagnostics-core/src/integration.ts
```diff
@@ -0,0 +1,117 @@
...
+ * | `core`        | {@link CoreDiagnosticsEvent} | The SQLite core emits a diagnostics event.            |
+ */
+export type DiagnosticsEvent =
```
simolus3, line 34: I also wonder if we really need separate calls for `currentSyncStatus` and `getUploadQueueStats`. For `observeSources` for example, the contract states that an integration would provide its current value to new listeners asap. We could do the same for these two.

packages/diagnostics-core/src/shapes.ts
```diff
@@ -0,0 +1,237 @@
...
+  downloading: boolean;
+  uploading: boolean;
+  hasSynced: boolean | null;
+  /** Epoch milliseconds of the last completed sync, or null. */
+  lastSyncedAt: number | null;
+  downloadProgress: ProgressState | null;
+  priorities: PriorityState[];
+  downloadError: string | null;
+  uploadError: string | null;
+  message: string;
```
simolus3, line 63: The message is derived from other fields and should not be serialized imo, the diagnostics ui can reconstruct it from other fields if necessary.

packages/diagnostics-core/src/shapes.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// --- sync status ---
+
+/** Download progress expressed in operation counts. */
+export interface ProgressState {
+  downloadedOperations: number;
+  totalOperations: number;
+  /** `0` to `1`. */
+  downloadedFraction: number;
```
simolus3, line 39: This is also something that can be reconstructed by the diagnostics, it shouldn't be transmitted.

packages/diagnostics-core/src/shapes.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// --- connection info ---
+
+/** Connection metadata for the attached client. */
+export interface ProtocolInfo {
+  /** The PowerSync service endpoint. */
+  endpoint: string | null;
+  /** Derive from the token subject when the SDK does not expose it. */
+  userId: string | null;
```
simolus3, line 124: Instead of forcing SDKs to do this, couldn't we have integrations send the raw JWT and decode the user id in the diagnostics UI?

packages/diagnostics-core/src/shapes.ts
```diff
@@ -0,0 +1,237 @@
...
+ * | ------------------- | -------------------------- | ---------------------------------------------------------------------- |
+ * | `reconnect`         |                            | Disconnect, then connect again with the last connector.                |
+ * | `disconnect`        |                            | Disconnect the client.                                                 |
+ * | `clearData`         |                            | Clear the local database, then connect again.                          |
+ * | `requestCheckpoint` |                            | Confirm the client is caught up. Needs checkpoint requests enabled.    |
+ * | `subscribeStream`   | {@link StreamActionArgs}   | Subscribe to a sync stream. `ttl` defaults to `0`.                     |
+ * | `unsubscribeStream` | `{ name, params? }`        | Release a subscription created with `subscribeStream`.                 |
+ */
+export type ActionName =
+  | 'reconnect'
```
simolus3, line 217: Is reconnecting that useful? We might add "half-connected" states in the future where we're only connected for downloads or uploads, which make a single action for this ambiguous. So I think we should remove this for now.

packages/diagnostics-core/src/live-database.ts
```diff
@@ -0,0 +1,106 @@
+/**
+ * The slice of a live PowerSync JavaScript database the in-page agent reads.
+ *
+ * Declared structurally so the tool never imports an SDK package: any `PowerSyncDatabase` satisfies
+ * this interface by shape. Only what the agent actually calls is listed. If the SDK renames one of
+ * these members, the agent fails to type-check against it here — deliberately, at the seam.
```
simolus3, line 6: With the Devframe setup, do we really still need `agent.ts`, `state.ts` and these interface definitions in `diagnostics-core`? Couldn't we move them to `diagnostics`, and depend on types from the JS SDK there?

This pretends to be SDK-agnostic when it's not: `SdkIntegration` is the SDK-agnostic interface.

packages/diagnostics-core/src/store.ts
```diff
@@ -0,0 +1,110 @@
+import { atom, type ReadableAtom, type WritableAtom } from 'nanostores';
```
simolus3, line 1: These stores are only used to derive UI state, right? Shouldn't they be in `diagnostics-ui` then?

packages/diagnostics-core/README.md
```diff
@@ -0,0 +1,94 @@
+# PowerSync Diagnostics — core
```
simolus3, line 1: Please also add a note that this is an internal package used by PowerSync SDKs, and not something that users would ever have to import or depend on directly.

(same for diagnostics-ui)

.changeset/angry-yaks-push.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/diagnostics-ui': minor
+---
+
+Initial release: the shared PowerSync diagnostics UI, rendered from an `SdkIntegration`.
```
simolus3, line 5: The initial release will not be published with changeset, remove this or replace it with a "Switch to trusted publishing" patch message for the subsequent release.

Also applies to `.changeset/shiny-items-reflect.md` and `.changeset/silent-dots-reflect.md`, you can probably merge them into a single item too.

.changeset/silly-apricots-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/nuxt': minor
+---
+
+**Breaking:** `NuxtPowerSyncDatabase` is removed; use `PowerSyncDatabase` with `connect(connector, { diagnostics: true })`. Diagnostics now come from `@powersync/diagnostics`: a PowerSync tab on Nuxt DevTools 3, the devframe dock with MCP tools on Nuxt DevTools 4.
```
simolus3, line 5: It looks like I have accidentally bumped that package to `1.0.0` a while ago 🤦 This needs to be `major` for semver compatibility now.

packages/diagnostics-ui/src/components/StatusBar.vue
```diff
@@ -0,0 +1,112 @@
...
+  warning: 'text-warning',
+  error: 'text-destructive',
+  muted: 'text-muted-foreground'
+};
+
+const errorText = computed(() => status.value?.downloadError ?? status.value?.uploadError ?? null);
+const pending = computed(() => uploadQueue.value?.count ?? 0);
+const bucketCount = computed(() => buckets.value.length);
+const totalOps = computed(() => buckets.value.reduce((sum, b) => sum + (b.downloadedOperations || 0), 0));
+const lastSynced = computed(() => (status.value?.lastSyncedAt ? formatRelative(status.value.lastSyncedAt) : null));
```
simolus3, line 60: Relative times are static, which looks weird for seconds (e.g. a "last synced 10s ago" doesn't increment automatically).

packages/diagnostics-ui/src/components/DiagnosticsPanel.vue
```diff
@@ -0,0 +1,227 @@
...
+  plugins: [powersyncDevtools()]
+});
+
+// Enable the core diagnostics stream for per-bucket progress.
+db.connect(connector, { diagnostics: true });`
+  },
+  {
+    label: 'Dart',
+    lang: 'dart',
+    code: `// Diagnostics are on in debug builds and off in release builds. Open the
```
simolus3, line 78: The snippet for JavaScript tells users to install the vite tool, which is reasonable.

Dart uses a `diagnostics: true` parameter that doesn't exist, diagnostics are enabled by default in debug builds as the comment points out. Maybe the screen should just say something like "If any PowerSync databases are opened in your app, their state can be inspected here"?

We could also put a link to the relevant setup guide for each SDK.

packages/diagnostics-ui/src/components/DiagnosticsPanel.vue
```diff
@@ -0,0 +1,227 @@
...
+  { value: 'streams', label: 'Streams' },
+  { value: 'config', label: 'Config' },
+  { value: 'logs', label: 'Logs' }
+];
+
+// Keep the selected tab across refreshes (per viewer).
+const TAB_KEY = 'powersync-diagnostics-tab';
+function initialTab(): string {
+  try {
+    const stored = localStorage.getItem(TAB_KEY);
```
simolus3, line 100: I think `sessionStorage` is a more appropriate location for this.

packages/diagnostics-ui/src/components/DiagnosticsPanel.vue
```diff
@@ -0,0 +1,227 @@
...
+            <IconSync :class="['size-4', syncing && 'animate-spin']" />
+          </button>
+          <button
+            type="button"
+            class="rounded p-1 text-muted-foreground transition-colors hover:bg-destructive/15 hover:text-destructive disabled:opacity-50"
+            title="Clear &amp; re-sync (wipe local data and download again)"
+            aria-label="Clear and re-sync"
+            :disabled="clearing"
+            @click="clearAndResync"
+          >
```
simolus3: This is a destructive action, kind of crazy to put it in every panel bar without any confirmation.

The sync status tab has a dedicated red button for this, I don't think we need another.

packages/shared-internals/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -193,6 +195,20 @@ export abstract class AbstractStreamingSyncImplementation
...
     this.crudUpdateListener = undefined;
+    this.diagnosticsChannel?.close();
+    this.diagnosticsChannel = undefined;
+  }
+
+  /**
+   * Broadcasts a core diagnostics event on a same-origin channel for diagnostics tooling to consume.
+   * Works from a shared worker to the page. No-op where BroadcastChannel is unavailable.
+   */
+  emitDiagnostics(event: DiagnosticsEvent): void {
```
simolus3, line 206: I believe the web SDK has subclasses of this anyway, so instead of checking for `BroadcastChannel` in shared code, could we make this a no-op here and let interested SDKs override the method?

tools/powersynctests/package.json
```diff
@@ -18,9 +18,9 @@
     "chai-as-promised": "^8.0.2",
     "chance": "^1.1.13",
     "mocha": "^10.8.2",
-    "react": "^19.2.4",
+    "react": "19.2.4",
```
simolus3, line 21: Why did we adopt exact versions here?

packages/diagnostics/README.md
```diff
@@ -0,0 +1,154 @@
...
+  // Vite >= 8.3. On Vite 7, spread `...(await DevTools())` from `@vitejs/devtools` into `plugins` instead.
+  devtools: true,
+  plugins: [powersyncDevtools()]
+});
+```
+
+Enable the core diagnostics stream when you connect. This gives the Buckets tab per-bucket totals:
+
+```ts
+await db.connect(connector, { diagnostics: true });
```
simolus3, line 42: I don't mind this, but since we already have a vite plugin I wonder if we could apply some transformation magic to make this the default for debug builds when the devtools plugin is applied?

That's also kind of hacky / surprising though, so maybe we shouldn't do it. Does the UI show a hint when the user forgets to enable this?

packages/diagnostics/README.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+Start the dev server and open your app. The first time, Vite DevTools asks you to confirm the browser with a one-time code printed in your terminal. Then the dock at the bottom of the page shows a **PowerSync** entry.
+
+How it works: the plugin mounts the definition into Vite DevTools. The dock runs a small script inside your app page that finds the open PowerSync database and serves it to the dev server. The diagnostics UI, in the dock's iframe, and any MCP client talk to the dev server, which forwards to the page. So an MCP call reaches your database only while a trusted browser tab has the app open, and the standalone DevTools window at `/__devtools/` shows "No PowerSync database attached" until one is. Every open tab attaches its own database; the UI shows the first one and offers a picker when there are several.
```
simolus3, line 47: I think this readme should only mention things relevant for users, IMO this paragraph has too many details.

packages/diagnostics/README.md
```diff
@@ -0,0 +1,154 @@
...
+const db = new PowerSyncDatabase({ schema, database: { dbFilename: 'app.db' } });
+await db.connect(connector, { diagnostics: true });
+
+if (process.env.NODE_ENV !== 'production') {
+  const devtools = await enablePowerSyncDiagnostics(db, { port: 9999 });
+  console.log(`PowerSync DevTools: ${devtools.url}`);
+}
+```
+
+Your process serves the DevTools window at that port. The browser confirms itself with a one-time code printed in your terminal; pass `auth: false` to trust every local browser. MCP tools are at `<url>/__mcp`.
```
simolus3: > The browser confirms itself with a one-time code printed in your terminal

That sounds annoying. Dart implements this by:

1. Picking a random port.
2. Making the authentication code part of the URL, so if you click on it you don't need to authenticate.

I think that's good enough, we just need to avoid a predictable port that other websites could probe via `fetch`.

packages/diagnostics/README.md
```diff
@@ -0,0 +1,154 @@
...
+import { connectAgent } from '@powersync/diagnostics/agent';
+
+const stop = await connectAgent(db, {
+  baseURL: 'http://localhost:9999/',
+  authToken: process.env.POWERSYNC_DEVTOOLS_TOKEN, // the window's --token, or start it with --no-auth
+  sdk: '@powersync/react-native'
+});
+```
+
+This is the path for React Native and for any host without a DevTools dock. Verified from a node process; the phone is the next test. For a device on the same network, start the window with `--host 0.0.0.0 --token <secret>`, point `baseURL` at your computer's address, and pass the same secret as `authToken`. The client fills in a `location` global when the runtime has none (Hermes, node) and skips the core event channel when `BroadcastChannel` is missing, so bucket totals stay empty there.
```
simolus3: This paragraph doesn't belong in a readme.

(note: We should probably have either Kobie or Benita proof-read this as well, since it's public documentation. Or maybe we shorten this readme and just link to the docs page once we have one)

packages/diagnostics/README.md
```diff
@@ -0,0 +1,154 @@
+# PowerSync DevTools
```
simolus3, line 1: Outside of the happy path (vite), I think we should explain the delivery modes more clearly (whether it's the app itself launching a server or whether the devtools ui is served by another tool that the app connects to).

I'd prefer the app launching a server and printing a URL to connect to. That requires some indirection for Android emulators and debug-specific permissions on both iOS and Android, but launching a standalone devtools UI, pasting the URL in code and only then launching the app is very unwieldy.

For this PR, maybe we should focus on things that work (vite + node api to launch a dev server). Skip the cli binary, authentication and "app connects to existing devtools instance" mode.

packages/diagnostics/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+```ts
+import { PowerSyncDatabase } from '@powersync/node';
+import { enablePowerSyncDiagnostics } from '@powersync/diagnostics/node';
+
+const db = new PowerSyncDatabase({ schema, database: { dbFilename: 'app.db' } });
+await db.connect(connector, { diagnostics: true });
+
+if (process.env.NODE_ENV !== 'production') {
+  const devtools = await enablePowerSyncDiagnostics(db, { port: 9999 });
```
simolus3, line 67: Could we add this to the existing node demo?

packages/diagnostics-core/src/shapes.ts
khawarizmus: Yes, and I was also wondering if we can should a mechanism to override the jwt to do impersonation of a specefic user. That would remove the need for the diagnostics client. But at this stage, I think it's best for it to be in another PR.

packages/diagnostics-ui/src/components/DiagnosticsPanel.vue
khawarizmus: I had to add it to every panel because it’s the only way to check that sync is working and debug things like Sync Streams, progress, etc.

What I’ve done now is remove the duplicate button from the SyncDiagnostics tab and keep the button in the panel. The button now has a tooltip saying “Hold to wipe”, so the user has to hold it for a moment to trigger the action (while they hold it, there is a circular progress bar in the button showing when the action will trigger).

packages/diagnostics/README.md
khawarizmus: I have dropped the CLI completely from this PR. We can explore mobile later in a different PR/Proposal

PR #1103, opened 2026-09-15
[Fix] Capacitor database locked after webview reload on Android

### LINKED ISSUE #1100: Sqlite Database Locks After App Restart
My database locks whenever the app restarts, which it does after every file update. This has led to me having to rename the database file after every update that occurs when the app is connected to powersync. If I dont connect to the power sync service then it is fine. I would assume this has to do with powersync performing some scheduled query during restart.

Has anyone else faced / is facing this problem who has solved it? I'm not sure what code would help find a solution so let me know what will help.

bean1352:
Fixes #1100.

On Android, if the webview reloads while a write transaction is open (live reload does this), every write after the reload fails with `database is locked (code 5)` until the app is killed.

The native connection keeps that transaction on the Capacitor plugin thread, which survives the reload. Android does not close a connection that is still in a transaction, so `closeConnection` in `init()` returns normally while the file stays locked. The plugin itself has the same bug, see capacitor-community/sqlite#659.

This change rolls back the old transaction before closing the connections, like the web adapter does when it hands out a connection lease.

## AI Disclosure

I reproduced this bug and implemented the fix using Claude Code (Fable 5.1)


simolus3 (dismissed):
Just to confirm, did you test this on Android? I didn't, but the changes make sense to me.

bean1352:
> Just to confirm, did you test this on Android? I didn't, but the changes make sense to me.

Yes, I was able to reproduce the error using the following function:
```
function reloadMidTransaction() {
  log('starting a write transaction and reloading the page INSIDE it (no COMMIT)…', 'warn');
  void db.writeTransaction(async (tx) => {
    await tx.execute('INSERT INTO repro_writes (note, ts) VALUES (?, ?)', ['reload-mid-tx', new Date().toISOString()]);
    location.reload();
    await new Promise(() => {}); // never resolve: the transaction must still be open when this JS context dies
  });
}
```

Chriztiaan (approved):

PR #1104, opened 2026-09-16
[Fix] kysely-driver: release the write lock when COMMIT or ROLLBACK fails

### LINKED ISSUE #1102: kysely-driver: write lock leaks when commit/rollback throws (no finally around releaseTransaction)
### Summary

`PowerSyncConnection` (`packages/kysely-driver/src/sqlite/sqlite-connection.ts`, v2.0.0) holds the PowerSync write lock open between `beginTransaction()` and `releaseTransaction()`. `releaseTransaction()` is only reached on the success path of `commitTransaction()` / `rollbackTransaction()`, so if `COMMIT` or `ROLLBACK` throws, the lock is never released and the `writeTransaction` callback awaits `transactionCompleted` forever.

### The code

```ts
async beginTransaction(): Promise<void> {
  return new Promise<void>((resolve, reject) => {
    this.#db
      .writeTransaction(async (tx) => {
        this.#tx = tx;
        const transactionCompleted = new Promise<void>((resolve) => {
          this.#completeTransaction = resolve;
        });
        resolve();
        await transactionCompleted;   // parked until releaseTransaction() runs
      })
      .catch(reject);
  });
}

async commitTransaction(): Promise<void> {
  if (!this.#tx) throw new Error('Transaction is not defined');
  await this.#tx.commit();            // throws here ...
  this.releaseTransaction();          // ... and this never runs
}

async rollbackTransaction(): Promise<void> {
  if (!this.#tx) throw new Error('Transaction is not defined');
  await this.#tx.rollback();          // same
  this.releaseTransaction();
}
```

There is no `finally`. A `ROLLBACK` that throws is the realistic trigger: Kysely calls `rollbackTransaction()` when the user callback rejects, so an error inside a transaction is exactly the path where a second failure leaves the connection parked.

### Impact

The write lock is process-wide, so one stuck connection blocks every subsequent write until the page or process is restarted — and it fails silently: no rejection surfaces from the parked `writeTransaction`, the next write simply never resolves.

`TransactionImplementation.runWith` in `@powersync/common` (`src/db/DBAdapter.ts`) already handles this correctly with a `finally`, so the two transaction paths behave differently under error.

### Suggested fix

Release in a `finally`, and make `releaseTransaction()` idempotent rather than throwing when `#completeTransaction` is already null:

```ts
async commitTransaction(): Promise<void> {
  if (!this.#tx) throw new Error('Transaction is not defined');
  try {
    await this.#tx.commit();
  } finally {
    this.releaseTransaction();
  }
}
```

(Same shape for `rollbackTransaction`.) Releasing the lock after a failed `COMMIT` is the right call regardless of the outcome: the SQLite transaction is over either way, and holding the lock cannot recover it.

`releaseConnection()` is also a no-op today; releasing any still-open transaction there would add a second safety net for the case where Kysely tears the connection down without a commit or rollback.

### Note

We hit a `database is locked` incident whose actual root cause was a second OS process, not this — this one showed up while auditing every path that can hold the write lock. Reporting it as a latent defect, not as a live bug we have reproduced in production.

bean1352:
Fixes #1102.

The driver opens a PowerSync write transaction when Kysely begins one and keeps it open until Kysely commits or rolls back. We only released it when the COMMIT or ROLLBACK succeeded.

If a transaction callback throws after the SQLite transaction has already ended, Kysely's ROLLBACK throws too, so the release never runs. The write lock stays held and every write after that waits forever without an error. An `INSERT OR ROLLBACK` that hits a duplicate key does it, so does a trigger using `RAISE(ROLLBACK, ...)`.

What changed:

- `commitTransaction` and `rollbackTransaction` release in a `finally`
- `releaseTransaction` no longer throws when it runs twice
- `releaseConnection` rolls back and releases a transaction Kysely left open

## AI Disclosure

I reproduced this bug and implemented the fix using Claude Code (Fable 5.1)


Chriztiaan (dismissed):

packages/kysely-driver/tests/sqlite/transaction-release.test.ts
```diff
@@ -0,0 +1,69 @@
+import { AbstractPowerSyncDatabase } from '@powersync/common';
```
simolus3: Nit: Use `CommonPowerSyncDatabase` in new code, `AbstractPowerSyncDatabase` is a deprecated type-alias to it.

simolus3 (dismissed):

simolus3 (approved):

PR #1108, opened 2026-09-16
[Fix] Sync disconnect hangs and stale error status

Manrich121:
## Problem

`disconnect()` waits for the sync loops to finish. A checkpoint retry waiting on an already-aborted signal or a pending shared-worker connector call could prevent it from completing, leaving subsequent `connect()` calls pending.

Intentional disconnects could also leave a stale `downloadError` of `Disconnect has been requested` on the sync status.

## Changes

- **Settle the checkpoint retry loop on disconnect.** Handle signals that have already aborted and exit when checkpoint readiness is cancelled. This can happen when disconnect interrupts the retry delay and affects non-legacy checkpoint modes.
- **Cancel pending connector waits on disconnect.** Forward the sync abort signal to `uploadCrud` and `postCheckpointRequest`, allowing the shared worker to stop waiting when disconnected. Connector calls have no timeout; `fetchCredentials` remains unbounded while its tab stays open.
- **Avoid reporting intentional disconnects as download errors.** Check `signal.aborted` to distinguish requested disconnects from failures that should still be reported and retried.
- **Clear cached transient status when disposing the sync implementation.** Prevent new tabs from inheriting stale errors while preserving core status, including `hasSynced` and `lastSyncedAt`.

## Testing

TypeScript build passes. All 13 focused tests pass across checkpoint readiness, checkpoint retry cancellation, connector signal forwarding, and disconnect error reporting.

## 🤖 AI Disclosure

*I used Claude Code Opus 5 to investigate the issue, implement the initial changes, and draft the original PR description. I used OpenAI Codex to review the changes, simplify comments, remove connector timeouts, and update this description.*


packages/shared-internals/src/client/BasePowerSyncDatabase.ts
```diff
@@ -267,6 +267,7 @@ export abstract class BasePowerSyncDatabase<Options extends BasePowerSyncDatabas
     return {
       ...options,
       adapter: this.bucketStorageAdapter,
+      // PowerSyncBackendConnector accepts no abort signal; these callbacks await it directly.
```
simolus3, line 271: Maybe turn this into a todo comment, I think we should eventually forward the abort signal to controllers as well.

packages/shared-internals/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -191,7 +198,9 @@ export abstract class AbstractStreamingSyncImplementation
   abstract obtainLock<T>(lockOptions: LockOptions<T>): Promise<T>;
 
   private async requestNextCheckpointFromService(abort: AbortSignal): Promise<string> {
-    await this.checkpoints.waitForCheckpointRequestsReady(abort);
+    if (!(await this.checkpoints.waitForCheckpointRequestsReady(abort))) {
+      abort.throwIfAborted();
```
simolus3: `waitForCheckpointRequestsReady` can only return false when aborted, so to make the intention clearer this should perhaps just throw directly

```suggestion
      throw abort.reason;
```

packages/shared-internals/src/client/sync/stream/AbstractStreamingSyncImplementation.ts
```diff
@@ -455,7 +468,11 @@ The next upload iteration will be delayed.`
          * The WebRemote should only abort pending fetch requests or close active Readable streams.
          */
 
-        if (ex instanceof AbortOperation) {
+        // Check the disconnect signal: connector timeouts also throw AbortOperation
+        // but must still be reported and retried.
```
simolus3: I don't like `SharedSyncImplementation` setting a timeout for connector invocations. It forces us to handle web-specific things in here, and might break for harmless cases like a developer setting a breakpoint in their `fetchCredentials` implementation.

Even without a timeout, wouldn't this fix the issue because the abort is now properly propagated? IMO, doing that consistently (including on `fetchCredentials`) would be better than setting static timeouts.

Manrich121: Good point, it has been reverted/removed

.changeset/checkpoint-retry-disconnect-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+'@powersync/shared-internals': patch
+---
+
+Fix `disconnect()` never completing when it interrupts the checkpoint request retry delay. Waking the delay let the retry loop wait on a signal that had already been aborted, so it could no longer be notified and never returned. Because `disconnect()` waits for the sync loops to finish, both it and every `connect()` queued behind it stayed pending for the rest of the session. Only affects connections using a `checkpointMode` other than `legacy`.
```
simolus3: nit: IMO this is overly detailed for a changelog entry:

```suggestion
Fix `disconnect()` never completing in some cases when `checkpointMode` is set to requests.
```

.changeset/stale-download-error-after-disconnect.md
```diff
@@ -0,0 +1,10 @@
+---
+'@powersync/shared-internals': patch
+'@powersync/web': patch
+---
+
+Fix a stale `downloadError` of "Disconnect has been requested" being reported after a normal disconnect.
+
+Stopping the sync client on request was recorded as a download failure. Because `downloadError` is only cleared by a completed sync, it then stayed on the status indefinitely whenever no sync completed afterwards. Aborts unrelated to a requested disconnect are still reported and retried.
+
+On the web, the shared worker also kept the last status of a sync implementation after disposing it and handed that status to every tab that connected afterwards, so new tabs could report an error they never encountered. The worker now drops that in-flight state with the implementation, while keeping `hasSynced` and `lastSyncedAt`.
```
simolus3: Same here

```suggestion
```

packages/shared-internals/tests/client/sync/stream/CheckpointDisconnect.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    const firstRead = Promise.withResolvers<void>();
+    const readCheckpointRequestId = vi.fn(async () => {
+      firstRead.resolve();
+      return '1';
+    });
+    const sync = new TestSync({
+      subscriptions: [],
+      logger: { log: vi.fn() },
+      adapter: { readCheckpointRequestId }
+    } as unknown as AbstractStreamingSyncImplementationOptions);
```
simolus3: This `as unknown as` trick makes our code harder to maintain as we'll have to deal with test failures tsc can't catch when we change implementation details.

Can't we run this test against a real database, mocking only `readCheckpointRequestId`? Putting this to the "checkpoint requests" group in `node/tests/sync.test.ts` works for me:

```TypeScript
    databaseTest(
      'disconnect settles when the download iteration ends during the checkpoint retry delay',
      async ({ database }) => {
        await database.init();

        class TestSync extends AbstractStreamingSyncImplementation {
          obtainLock<T>({ callback }: any): Promise<T> {
            return callback();
          }

          async startCheckpointRetry() {
            const controller = new AbortController();
            this.abortController = controller;
            await this['checkpoints'].markCheckpointsReady(Promise.resolve());
            this.streamingSyncPromise = Promise.all([
              Promise.resolve(),
              Promise.resolve(),
              this['repostUnacknowledgedCheckpointRequests'](
                controller.signal,
                resolveSyncOptions({ checkpointMode: 'requests' }, SyncStreamConnectionMethod.HTTP)
              )
            ]);
          }

          endDownloadIteration() {
            this['checkpoints'].downloadIterationEnded();
          }
        }

        const firstRead = Promise.withResolvers<void>();
        const logger: PowerSyncLogger = { log: vi.fn() };
        const adapter = new SqliteBucketStorage(database.database, logger);

        const readCheckpointRequestId = vi.spyOn(adapter, 'readCheckpointRequestId');
        readCheckpointRequestId.mockImplementation(async (_) => {
          firstRead.resolve();
          return '1';
        });

        const sync = new TestSync({
          subscriptions: [],
          serializedSchema: new Schema([]).toJSON(),
          logger,
          adapter,
          uploadCrud: vi.fn(),
          postCheckpointRequest: vi.fn(),
          remote: new NodeRemote(
            {
              fetchCredentials: vi.fn()
            },
            logger,
            {
              customFetch: () => {
                throw new Error(`Unexpected HTTP request`);
              }
            }
          )
        });

        await sync.startCheckpointRetry();
        await firstRead.promise;
        // The download loop resets checkpoint readiness as it exits. Disconnect wakes the
        // retry delay, which used to wait again on the already-aborted signal forever.
        sync.endDownloadIteration();
        await sync.disconnect();

        expect(sync.isConnected).toBe(false);
      },
      500
    );
```

packages/shared-internals/tests/client/sync/stream/ConnectorAbortSignal.test.ts
```diff
@@ -0,0 +1,58 @@
...
+function createSync(options: Partial<AbstractStreamingSyncImplementationOptions>) {
+  return new TestSync({
+    subscriptions: [],
+    logger: { log: vi.fn() },
+    ...options
+  } as unknown as AbstractStreamingSyncImplementationOptions);
+}
+
+// The shared worker needs this signal to stop waiting on unresponsive tabs.
+describe('connector calls receive the sync abort signal', () => {
```
simolus3: In the web package, we have the ability to run sync tests with a real shared worker using a mocked http client. That feels like a more reliable test than these mocks, we could write a test connecting with a connector returning promises that never settle. Then, calling `disconnect()` after requesting a checkpoint must still complete.

packages/shared-internals/tests/client/sync/stream/DownloadErrorOnDisconnect.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    return this.iterate();
+  }
+
+  runStreamingSync(signal: AbortSignal) {
+    return this['streamingSync'](signal, syncOptions);
+  }
+}
+
+describe('downloadError on disconnect', () => {
+  test('a requested disconnect is not recorded as a download error', async () => {
```
simolus3: Couldn't we run this test against a real database too, connecting in `node/tests/sync.test.ts` and asserting that the download error is undefined after disconnecting?

packages/shared-internals/tests/client/sync/stream/DownloadErrorOnDisconnect.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    await started.promise;
+    controller.abort();
+    aborted.resolve();
+    await running;
+
+    expect(sync.syncStatus.downloadError).toBeUndefined();
+  });
+
+  test('a genuine sync failure is still recorded', async () => {
```
simolus3: We already have tests asserting a download error is reported for errors, no need to repeat that.

packages/web/src/worker/sync/SharedSyncImplementation.ts
```diff
@@ -147,7 +147,13 @@ export class SharedSyncImplementation extends BaseObserver<SharedSyncImplementat
 
         return {
           sync,
-          onDispose
+          onDispose: () => {
+            removeStatusListener();
+
+            // Clear transient state so new tabs don't inherit stale errors.
+            // Preserve core status, including hasSynced and lastSyncedAt.
```
simolus3, line 154: I'm not sure if this is a real problem or if `onDispose` is only called in disconnected states to begin with, but we might want to set `connected`, `connecting` and `downloading` to false / null on the core sync status here?

packages/web/src/worker/sync/SharedSyncImplementation.ts
```diff
@@ -387,25 +393,29 @@ export class SharedSyncImplementation extends BaseObserver<SharedSyncImplementat
             }
           },
           fetchCredentials: () => {
```
simolus3: Should this receive an abort signal as well? I didn't take a look in detail, but it seems like adding an optional signal parameter on `RemoteConnector.fetchCredentials` works (it would be forwarded by `AbstractRemote.{getCredentials, prefetchCredentials, fetchCredentials, buildRequest}`, which in turn is called by `fetchAndDecodeJson`, `socketStreamRaw` and `fetchStreamRaw`, methods that all have an abort signal provided by the caller).

packages/shared-internals/tests/client/sync/stream/CheckpointDisconnect.test.ts
Manrich121: Thanks for this, I moved it

packages/shared-internals/tests/client/sync/stream/DownloadErrorOnDisconnect.test.ts
Manrich121: Moved it

packages/web/src/worker/sync/SharedSyncImplementation.ts
Manrich121: Added this

simolus3 (dismissed):
Could you rebase ontop of / merge main? That contains a fix for the test action job not starting, apart from that this LGTM.

simolus3 (approved):
We'll need another +1 now. For context, I've added [these commits](https://github.com/powersync-ja/powersync-js/pull/1108/changes/b5626241a25b1f345e9efbe9925690724f4ba533..0f578a5f8bc0319fb4f59ee00d9c6188110b028e) to fix web reconnect tests.

I used Claude code to find the issue and implement the fix.

Chriztiaan (approved):

PR #1109, opened 2026-09-17
[WIP] Move the dashboard's diagnostics logic to diagnostics-core

bean1352:
The dashboard built a lot of diagnostics logic on top of the protocol. This moves it into `@powersync/diagnostics-core` so other hosts get it too.

## What moved

Reading a Sync Config's parameters, rolling bucket stats up per stream, table and bucket stats over SQL, recovering a session from service logs, and inferring a schema from core events. It all works on plain data, so the main entry point still imports no SDK.

## New `./web` entry point

Opens a headless client and serves it as an `SdkIntegration`. `@powersync/web` is an optional peer, so hosts that only inspect an app's own client never load it.

Two bits answer review feedback on the PoC:

- Defaults to OPFS. IndexedDB gets slow past a gigabyte.
- Reset deletes the database files instead of clearing rows, and disconnects first so the client doesn't download it all again.

## Protocol

`unsubscribeStream` gains `mode: 'all'`, plus a new `unsubscribeAllStreams`. Releasing a subscription only starts a TTL, and diagnostics needs the stream to actually stop.

## AI disclosure

I implemented this change using Claude (Fable 5.1), and reviewed and tested it myself.


.changeset/quiet-lamps-observe.md
```diff
@@ -0,0 +1,12 @@
+---
+'@powersync/diagnostics-core': minor
+'@powersync/diagnostics': patch
+---
+
+Add the logic a diagnostics host needs on top of the protocol, and a headless web test client.
+
+The main entrypoint gains SDK-free helpers: `readSyncConfigParameters` (what a Sync Config expects of clients), `parseStreamBucketName` and `collectStreamStats` (buckets to streams), `readTableStats` / `readBucketStats` / `readStoredSubscriptions` (over `runQuery`), `collectImpersonationTarget` and `recoverSubscriptions` (a session recovered from service logs), `ObservedSchema` (a schema inferred from core `SchemaChange` events), and small token, log and SQL helpers. `./js` gains `BroadcastCoreEvents`.
```
simolus3, line 8: Is `readSyncConfigParameters` used anywhere outside the dashboard? I assume it requires the caller to invoke a service API to resolve config metadata first?

simolus3 (commented):
To show that this generalizes beyond what the dashboard needs, maybe good demo could be to update the diagnostics app to be based on `diagnostics-core` with a headless sdk (since that's basically what it already does).

I also feel like functionality that requires non-SDK service endpoints to work (like parsing log messages or analyzing a sync config for parameters) should likely not be here, unless we expect SDK diagnostics tooling to eventually make these requests too.

PR #1113, opened 2026-09-19
Fix electron demo

rkistner:
1. Switch from `better-sqlite3` to `node:sqlite`: Not a bug, but no more rebuild required.
2. Upgrade to latest electron.
3. Fix/workaround for `SyncStatus` not having a public `toJSON` anymore. Maybe we should add a public `toJSON` again?
4. Fix `getAll` incorrectly using `get`.
5. Fix further build issues.

This now runs the builds on CI via `test:build`, which should catch _some_ of the above issues in the future.

This still needs a webpack workaround for better-sqlite3. There might be a better solution for this, such as splitting the workers in the NodeJS SDK. For now, this works without modifying the SDK.

## AI Usage

Implemented changes using Codex gpt-6-astra. Manually tested.


demos/example-electron-node/src/main/index.ts
```diff
@@ -59,14 +61,25 @@ const createWindow = (): void => {
...
 // initialization and is ready to create browser windows.
 // Some APIs can only be used after this event occurs.
-app.whenReady().then(() => {
-  database.connect(new BackendConnector(), { connectionMethod: SyncStreamConnectionMethod.HTTP });
+app.whenReady().then(async () => {
+  await database.init();
 
   const forwardSyncStatus = (port: MessagePortMain) => {
-    port.postMessage(database.currentStatus.toJSON());
+    const postStatus = (status: typeof database.currentStatus) =>
```
simolus3: The `SyncStatus` interface itself is still exported (I hope), can we use

```suggestion
    const postStatus = (status: SyncStatus) =>
```

simolus3 (commented):
> Fix/workaround for `SyncStatus` not having a public `toJSON` anymore. Maybe we should add a public `toJSON` again?

The intention behind hiding this was that it's mostly an implementation detail needed for the web SDK. Given the use case of "I need to send a sync status across a custom channel", I think we could potentially add it again.

We only expose the interface of `SyncStatus` too though, no way to actually instantiate it from a JSON snapshot. So that would have to be added as well.

simolus3 (dismissed):

simolus3 (approved):

PR #1116, opened 2026-09-21
Run react tests on node

simolus3:
Tests for `packages/react/` run in a real browser using playwright. This is not really necessary: We want to test react hooks, which don't depend on browser features. Using `@powersync/node` instead of `@powersync/web` makes tests easier to debug locally, is faster and hopefully makes them more reliably in CI (in particular, it actually uses a fresh location for each test instead of re-using databases).

This also declares support for the latest `better-sqlite3` versions.

AI use: Mostly generated by Claude, manually reviewed by me.


packages/react/tests/useQuery.test.tsx
```diff
@@ -850,7 +852,7 @@ describe('useQuery', () => {
       });
 
       it('should handle dependent query parameter changes with correct state transitions', async () => {
```
simolus3, line 854: This test had an implicit assumption of queries not running in parallel, which is true for the default web database using IndexedDB but not with Node.

Chriztiaan (approved):

